PR #5, opened 2019-11-14
DP-16475 Sanitize user roles and name

mrossi113:
**Description:**
Removing user names and roles from the database.

**Jira:**
https://jira.mass.gov/browse/DP-16475

**To Test:**

**Screenshots/GIFs:**

---


weitzman (changes requested):
Instead of creating new Drush commands, I think it makes sense to extend sql:sanitize, like the rest of that file does. See how the new --sanitize-entities and --sanitize-keyvalue options are implemented.

drush/Commands/SanitizationCommands.php
```diff
@@ -195,4 +196,42 @@ private function getEntityStorage(string $type) {
...
+   * Sanitize the database table for the username.
+   *
+   * @hook post-command sql-sanitize
+   */
+  public function userName($result, CommandData $commandData) {
+    if (!$commandData->input()->getOption('sanitize-name')) {
+      return;
+    }
+    // User data table updated.
+    $name = Database::getConnection()->truncate('users_data table')->execute();
```
weitzman: User names are stored in `users_field_data`, not users_data. Also, we dont want to truncate that whole table, or else all our content will have no owner. I think we want a method very similar to https://github.com/drush-ops/drush/blob/40092d5dcf6f850fdf15d817c33602fe92c67c7c/src/Drupal/Commands/sql/SanitizeUserTableCommands.php#L26-L84. That sanitizes email addresses and passwords.

weitzman (changes requested):

drush/Commands/SanitizationCommands.php
mrossi113: I've updated the username sanitization option.

weitzman (commented):
userName looks good. You now need to declare the `sanitize-name` option by adding it to options() and messages() methods in same file

mrossi113:
Added the `sanitize-name` option to the `options()` and `messages()` methods.

PR #7, opened 2019-11-15
DP-16322 Removing YAML content module

mrossi113:
**Description:**
Removing some of the modules that are not needed anymore.

**Jira:**
https://jira.mass.gov/browse/DP-16322

**To Test:**
- [ ] Review the changes to extension.
- [ ] Removed the yaml_content and mass_yaml_content modules completely.

**Screenshots/GIFs:**

---


weitzman:
The code here looks good. IMO this is premature. I would merge this after we have green tests with the super sanitized database. I'm happy to proceed if others think thats best.

weitzman:
Upon further contemplation, YAML content never worked out for us. If we end up needing this, lets use YAML files and our own entity creation.

PR #9, opened 2019-11-19
Super sanitized DB building at Circle

weitzman:
Enhance mysql_rebuild workflow to also build a super sanitized DB and push it to docker/openmass (analogous to our current sanitized DB for development).

1. [At least 4 follow-ups exist](https://jira.mass.gov/browse/DP-16469?focusedCommentId=306969&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-306969) 
1. Anyone have a thought how to architect a test/monitor that looks at you super sanitized and makes sure that usernames, unpublished content etc. are not present?

**Jira:**
https://jira.mass.gov/browse/DP-16469

**To Test:**
- [ ] See 2 new CircleCI "populate" jobs in the test results below. Make sure they are green. That means we are building regular and "super" DB images.
- [ ] Try to use regular sanitized DB with our local dev environment (i.e. we made no changes so it should work as usual). Make sure we can bootstrap via `drush status`. Make sure email addresses are sanitized but usernames and user__roles table is not.
- [ ] Use the super sanitized DB by moving aside `docker-compose.override.yml` and doing `ahoy down && ahoy up`. Make sure we can bootstrap via `drush status`. Make sure everything is sanitized. 

---


.circleci/config.yml
```diff
@@ -893,20 +907,26 @@ workflows:
               only: /.*/
 
 
-  # Daily rebuild of our sanitized mysql data image, runs at 12:30 PM EST.
+  # Daily rebuild of our sanitized and super-sanitized mysql data images.
```
mrossi113, line 907: Do you want to have two different times for this instead? Because of the API will be available at 12:30 pm each day.

mrossi113 (commented):
One comment about the time.

.circleci/config.yml
weitzman: At first blush, I dont really care about what time populate our DBs. Are you saying that backups happen before 12:30 so we should populate then? Please clarify.

mrossi113: The Acquia API becomes available to us at 12:30 pm ET that why the sanitized DB runs in the afternoon. If we change this to 8:48 am ET it will grab a two day old API compared to a day old database.

.circleci/config.yml
```diff
@@ -893,20 +907,26 @@ workflows:
...
   mysql_rebuild:
     jobs:
       - build
       - populate:
           requires: [build]
+          tag: comass/mysql-sanitized:latest
+      - populate:
+          requires: [build]
+          isSuper: true
+          tag: massgov/openmass:latest
```
rbayliss: Can we rename to `image` rather than tag?  Not super important - it's just a little confusing, cause it's not just a tag.

.circleci/config.yml
```diff
@@ -577,10 +584,17 @@ jobs:
       - run:
           command: docker-compose exec drupal scripts/ma-refresh-local --db-prep-only
           no_output_timeout: 60m
-      - run: docker-compose stop --timeout 60 mysql
       - run: docker login -u $DOCKER_USER -p $DOCKER_API_KEY
-      - run: docker commit $(docker-compose ps -q mysql) comass/mysql-sanitized:latest
-      - run: docker push comass/mysql-sanitized:latest
+      - when:
```
rbayliss: Can we tuck this away behind a flag to `ma-refresh-local`? As it stands today, the complexity of actually fetching and sanitizing the database is contained there, and I think we should keep it that way so as not to spread it around more than is necessary.

rbayliss (changes requested):
This looks pretty good - one thing I'm concerned about is MySQL's binlog/redo logs, which retain transactional data about the historical changes to the db. I've poked around the image that's been pushed, and I can't find anything insecure, but I'm concerned that upstream configuration changes could bite us in the future and cause those logs to contain data that should have been wiped out.

The only way I can think of to 100% future-proof this is to import data into one mysql image, sanitize it, then dump it out as SQL and reimport into a clean image.  But I think that can happen later on as a followup, since we have no private data left in there now.

One tip - I was able to remove `/var/lib/mysql-image/ib_logfile*` (InnoDB redo logs), then start the server with no ill effects. That frees up ~100mb of in-container space (no idea what effect that has on the final image). Might be worth adding just it's easy to do.    

Other than that, there are a few smaller asks below.

.circleci/config.yml
weitzman: Seems to me like the databases can be 0-24 hours old since we run once per day. Not seeing how it could be older than that. I dont care much where it sits in the range of 0-24.

.circleci/config.yml
weitzman: @yriahi  - could you look into enhancing ma-refresh-local so it supports a super sanitize option? Thats just an alternate drush sql:sanitize command with more options than the current one. I originally looked into this but was unable to find a clean way to add it.

yriahi: Added the options for `ma-refresh-local` here #16

README.md
```diff
@@ -34,11 +34,8 @@ See the [Table of Contents](/docs/README.md) for additional documentation relate
     ```bash
     sudo wget -q https://github.com/devinci-code/ahoy/releases/download/2.0.0/ahoy-bin-darwin-amd64 -O /usr/local/bin/ahoy && sudo chown $USER /usr/local/bin/ahoy && chmod +x /usr/local/bin/ahoy
     ```
-
-1. Run `ahoy start` to setup a fresh new site (n.b. takes about 1 hour to complete). This does the following: 
-    - Starts the Docker containers, which includes a database.
-    - Runs `composer install` to fetch all dependencies.
-    - Installs a fresh site based on latest config contained in this repository.
+1. Run `ahoy up` to start the Docker containers (n.b. takes about 1 hour to pull down the latest database).
```
mrossi113:
```suggestion
1. Run `ahoy up` to start the Docker containers (n.b. takes a few minutes to have the database container start up). 
```

mrossi113 (changes requested):
👍 LGTM! Just the comment from @rbayliss and one small document change.

PR #14, opened 2019-11-25
Dp 16625 promotional page conditional kpi

joeg8612:
**Description:**
Add a KPI choice field to the promotional page and adds conditional visibility logic to only show KPI fields if they are chosen for use.
Changes click-through and percent KPs to be specified on a 0 to 100 scale with a percent suffix.
Changes related help text

**Jira:**
https://jira.mass.gov/browse/DP-16625

**To Test:**
- [x] On the promotional page, a new field exists "Key Performance Indicators". This is a text select list with 5 choices - each of the possible KPIs.  When you check off a KPI, the field appears on the page for entry.  
- [x] If you don't select any KPIs, you get an error.  
- [x] A percent sign suffix appears with the CTR and Percent KPIs. These allow values between 0 and 100.  The help text is changed to indicate that percentages between 0 and 100 should be used.
- [x] The KPI header on the KPI tab is changed to move some of the help text down to the KPI choice field.
- [x] Click the check marks for  2 KPIs. On one field add a number and for the other KPI don't add a number. Attempt to save. It should display an error with the field name.
- [x] Uncheck the empty KPI field and save the page; It should save without any errors.
- [x] Check 4 KPIs and fill out values for them all. You should get an error that only three are allowed.
- [x] Remove one of the KPI checkboxes but leave the value in the underlying field.  When you save the page, the value will be removed from that field.

---


yriahi (approved):
This looks good to me.

docroot/modules/custom/mass_validation/mass_validation.module
```diff
@@ -652,6 +652,37 @@ function _mass_validation_form_node_campaign_landing_form_validate(array &$form,
...
+    foreach ($promotional_kpi_choice as $choice) {
+      $field_name = $kpi_fields_mapping[$choice['value']];
+      $selected_kpi[] = $field_name;
+      $value = $form_state->getValue($field_name);
+      // If the KPI value is empty the error will be shown.
+      if (empty($value[0]['value'])) {
+        $form_state->setErrorByName($field_name, t('A value must be added to the this field.'));
+      }
+    }
+    # Take the unchecked checkboxes and reset the value to NULL.
```
mrossi113: @yriahi  change the `#` to a `//`. I used the wrong comment syntax here.

docroot/modules/custom/mass_validation/mass_validation.module
```diff
@@ -652,6 +652,37 @@ function _mass_validation_form_node_campaign_landing_form_validate(array &$form,
...
+    $unchecked = array_diff($kpi_fields_mapping, $selected_kpi);
+    foreach ($unchecked as $unchecked_field_name) {
+     $form_state->setValue([$unchecked_field_name, 0, 'value'], NULL);
```
mrossi113: @yriahi this line needs to be indented one more space to pass the code sniffer.

mrossi113: Specifically, line 683

yriahi: The indentation is now fixed.

docroot/modules/custom/mass_validation/mass_validation.module
yriahi: This was changed to use the correct comment syntax `//` instead of `#`.

PR #20, opened 2019-11-27
DRY the deploy jobs and remove ma:release

weitzman:
Also removes ma:release command as it no longer works.

**Jira:**
https://jira.mass.gov/browse/DP-16585

**To Test:**
- [ ] Code review before merge
- [ ] After merge, lets do a release and hotfix, without taking those to Prod. That will exercise some workflows.

---

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -505,7 +505,7 @@ workflows:
               only: /^release\/.*$/
       - backstop:
           requires: [deploy]
-          target: stage
+          target: test
           filters:
             branches:
               only: /^release\/.*$/
@@ -550,7 +550,7 @@ workflows:
               only: /^hotfix\/.*$/
     - backstop:
           requires: [deploy]
-          target: stage
+          target: test
           filters:
             branches:
               only: /^hotfix\/.*$/
@@ -596,7 +596,7 @@ workflows:
               only: /^mayflower\/.*$/
       - backstop:
           requires: [deploy]
-          target: stage
+          target: test
           filters:
             branches:
               only: /^mayflower\/.*$/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -134,7 +134,7 @@ jobs:
     <<: *container_config
     steps:
       - attach_workspace: {at: /var/www}
-      - run: yarn danger ci
+      - run: yarn danger ci --failOnErrors
 
   # Automate the tagging process in GitHub
   tagging_github:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -247,101 +247,32 @@ jobs:
       - run:
           command: "php scripts/ma-mayflower-develop.php"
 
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -134,7 +134,7 @@ jobs:
     <<: *container_config
     steps:
       - attach_workspace: {at: /var/www}
-      - run: yarn danger ci
+      - run: yarn danger ci --failOnErrors
```
weitzman, line 137: Now that Danger is its own job we want that job to fail if danger fails.

.circleci/config.yml
```diff
@@ -499,6 +430,8 @@ workflows:
           requires:
             - build
             - push_acquia
+          refreshDb: "1"
```
mrossi113: Nothing big, but does the order in parameter matter?
I'm seeing a different order in multiple places.

.circleci/config.yml
```diff
@@ -499,6 +430,8 @@ workflows:
...
+          target: cd
```
mrossi113, line 434: Nothing big, but does the order in parameter matter?
I'm seeing a different order in multiple places.

.circleci/config.yml
```diff
@@ -555,21 +488,23 @@ workflows:
...
+          requires: [deploy]
           target: test
           samples: 700
           filters:
             branches:
               only: /^release\/.*$/
       - backstop:
-          requires: [deploy_release_stage]
+          requires: [deploy]
           target: stage
```
mrossi113: I just noticed the target for backstop should be `test` missed in my first check.

.circleci/config.yml
```diff
@@ -599,21 +534,22 @@ workflows:
...
+          requires: [deploy]
           target: test
           samples: 200
           filters:
             branches:
               only: /^hotfix\/.*$/
     - backstop:
-          requires: [deploy_hotfix_stage]
+          requires: [deploy]
           target: stage
```
mrossi113: Same here the backstop should have `test` not stage as target.

.circleci/config.yml
```diff
@@ -643,21 +579,23 @@ workflows:
...
+          requires: [deploy]
           target: test
           samples: 500
           filters:
             branches:
               only: /^mayflower\/.*$/
       - backstop:
-          requires: [deploy_release_stage]
+          requires: [deploy]
           target: stage
```
mrossi113: Same here change target to `test`

mrossi113 (changes requested):
@weitzman sorry for the delay, but I had a question and noticed backstop is using the wrong target for stage environment we use `test`. Similar to the nightcrawler command and deploy.

.circleci/config.yml
weitzman: Order doesnt matter, fortunately.

mrossi113 (approved):
👍 LGTM!

PR #21, opened 2019-11-27
Update dependencies - cache_metrics, devel, drush

weitzman:
**Description:**
Update so we get more accurate cache metrics. Other projects updated for better hygiene

**Jira:**
https://jira.mass.gov/browse/DP-16598

**To Test:**
- [ ] Easiest is to merge this and pay attention to any failures during next deployment. Drush 10 is just a removal of deprecated code that we should not be using. If we are, a fata error will make that clear.

---


weitzman:
FYI I ran `dcdrush ma:deploy feature5 deps -vvv` while using this branch in order to verify that deployment still works. It does.

drush/Commands/DeployCommands.php
```diff
@@ -203,7 +203,7 @@ public function deploy($target, $git_ref, array $options = ['cache-rebuild' => T
       // Turn on Maint mode.
       $args = array('system.maintenance_mode', 1);
       $state_options = array('input-format' => 'integer');
-      $process = Drush::drush($targetRecord, 'state-set', $args, $state_options);
```
rbayliss, line 150: Does Drush 10 remove the aliases for the old-style command names?

rbayliss (commented):
@weitzman, just one question here.

drush/Commands/DeployCommands.php
weitzman: No it does not. Muscle memory was prioritized :)

PR #52, opened 2019-12-19
Upgrade to Acquia Cloud API 2

weitzman:
- Depends on https://github.com/massgov/DS-Infrastructure/pull/189. Lets merge that first.

**Jira:**
https://jira.mass.gov/browse/DP-16283

**To Test:**
- [x] Get new drush aliases file from DS-Infrastructure. See https://github.com/massgov/DS-Infrastructure/pull/189
- [x] Populate 2 new .env variables. See env.example change
- [x] Push this branch to feature5 using ma:deploy. Assure it succeeds.

---

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -366,9 +366,6 @@
             },
             "drupal/quick_node_clone": {
                 "2979426 - Ensures per-content type permissions are respected": "https://www.drupal.org/files/issues/2019-08-16/quick_node_clone-permissions_not_respected-2979426-13.patch"
-            },
-            "typhonius/acquia-php-sdk-v2": {
-                "Adds notifications method (https://github.com/typhonius/acquia-php-sdk-v2/pull/36)": "https://patch-diff.githubusercontent.com/raw/typhonius/acquia-php-sdk-v2/pull/36.patch"
             }
         },
         "composer-exit-on-patch-failure": true
--- a/composer.json
+++ b/composer.json
@@ -101,7 +101,6 @@
     ],
     "require": {
         "php": ">=7.1",
-        "acquia/acquia-sdk-php": "0.10.8",
         "blakestein/datalayer": "master",
         "christian-riesen/base32": "^1.3.1",
         "components/highlightjs": "9.7.*",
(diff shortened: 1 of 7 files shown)


composer.json
```diff
@@ -227,6 +226,7 @@
         "symfony/validator": "~3.4.0",
         "symfony/yaml": "~3.4.5",
         "twig/twig": "^1.38.2",
+        "typhonius/acquia-php-sdk-v2": "1.1.x-dev",
```
rbayliss: Can we constrain this to stable releases, or is there a specific reason we need to be on dev? If there isn't a particular reason, how about if we switch to `~1.1`?

rbayliss (changes requested):
This looks awesome!  My only request is to use a stable version of acquia-php-sdk-v2.

rbayliss:
I did check that the aliases were right, that this deploys cleanly, and that `ma:latest-backup-url` still works.  Everything checks out, but I still think we might want to wait until after the holidays to merge so we don't put ourselves in a position of having nobody around to support a problematic release.  But I'm good to go with these changes as soon as folks are back.

composer.json
weitzman: There is no release yet with https://github.com/typhonius/acquia-php-sdk-v2/pull/36 in it.

typhonius: I'm planning on releasing a new tag for both 1.1 and 2.0 branches over the holiday period if that would help!

typhonius: 1.1.1 and 2.0.1 have been released if you'd like to pin to those. Both have notifications built in so should be of use in this PR.

I'll mostly be working in the 2.x branch now as it's been rewritten to fulfil the growing needs of the library. That being said, because it has been rewritten, there were major API changes so would not apply to this PR without alterations.

rbayliss: @typhonius, thanks for the update!  Very helpful.

yriahi: I have now required `~1.1`. Thank you!

rbayliss:
Ah - actually, I just realized that we need to update `docker-compose.yml` and `.circleci/mysql-rebuild/docker-compose.yml` to use the new environment variables. Let's keep both the `AC_API_*` and `AC_API2_*` variables in there and clean up the `AC_API_*` ones in a followup.

yriahi:
@rbayliss I made the change to preserve both environment variables - Acquia Cloud v1 and v2.

PR #71, opened 2020-01-07
Allow forked PRs, with full testing workflow.

weitzman:
See a forked PR with full test run at https://github.com/massgov/openmass/pull/85. This was made by following the Testing steps below.

I don't love the trigger business we are adding here. I can't find a simpler way given the constraints of CircleCI.

**Jira:**
https://jira.mass.gov/browse/DP-16934

**To Test:**
- [x] Fork openmass and submit a PR. See that the PR gets `build` and `validate` steps and is green. The PR gets trigger-workflows step too but no work is done there.
- [ ] To get the forked PR fully tested, run something like 
```scripts/git-push-fork-to-upstream-branch origin massgov-outsider:forks-rule```
. Replace `massgov-outsider:forks-rule` with the base branch of your fork. You may copy it from the top of PR page. After running this script, notice that the fork PR now gets full testing, and can be merged when green.

---

### DIFF
--- a/.circleci/circle_trigger.sh
+++ b/.circleci/circle_trigger.sh
@@ -14,10 +14,11 @@ fi
 DATA="{ \"branch\": \"$CIRCLE_BRANCH\", \"parameters\": { $PARAMETERS } }"
 URL="${CIRCLE_API}/v2/project/${REPOSITORY_TYPE}/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}/pipeline"
 echo "Triggering pipeline with data:"
-echo -e "  $URL"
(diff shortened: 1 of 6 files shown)


weitzman:
The implementation here has run into a wall where we cant exit early enough from the `test` job because it tries to download the DB image during setup. See https://github.com/massgov/openmass/pull/73 for that failure.

So, I plan to bite the bullet and implement a re-trigger scheme where a fork will trigger pipeline run that itself triggers a pipeline run with a param that makes it run only a new 'fork' workflow. We can also bring back ma:release with this approach.

yriahi (approved):
This looks good to me!

PR #78, opened 2020-01-08
Use typhonius 2.x to download DB during ma:deploy and ma-release-local

weitzman:
**Jira:**
https://jira.mass.gov/browse/DP-17005

**To Test:**
- [ ] Use ma:deploy and make sure it succeeds. Same for ma-refresh-local

**Screenshots/GIFs:**

---

### DIFF
--- a/changelogs/DP-17005.yml
+++ b/changelogs/DP-17005.yml
@@ -1,0 +1,33 @@
+# Write your changelog entry here.  Every PR must have a changelog.
+#
+# You can use the following types:
+#   Added: For new features.
+#   Changed: For changes to existing functionality.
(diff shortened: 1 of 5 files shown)


weitzman:
Awaiting a resolution of https://github.com/typhonius/acquia-php-sdk-v2/pull/40

weitzman:
We may want to revisit this when updating to v2 of typhonius SDK.

weitzman:
Abandoned for now, as the DatabaseBackup class doesn't currently work for our large DB. Some guzzle params need to be set, per typhonius and the library does not yet do that.

PR #80, opened 2020-01-09
DP-16945: media redirect resolution

rbayliss:
**Description:**
This PR resolves an issue with the media download redirects where any media download URL that resulted in multiple redirect hops was "masking" the wrong URL.  The "correct" behavior for the URL masking is to only mask the URL if:
* It results from a request to `/media/XXX/download` or `/doc/XXX/download` AND
* Redirects to a URL in `/files/*`
In all other cases, we should simply serve up the response the server gave us (unmasked).

**Jira:**
https://jira.mass.gov/browse/DP-16945

**To Test:**
- [ ] Visit https://wwwcf.digital.mass.gov/media/1000001/download in your browser with the network tab open.
- [ ] Verify that this results in a redirect from https://wwwcf.digital.mass.gov/media/1000001/download to https://wwwcf.digital.mass.gov/doc/physician-certification-and-attestation-form-for-aca-section-1202-rates-for-physicians-who/download, then serves the file up directly from there. 

**Screenshots/GIFs:**

---


a-username (approved):
LGTM

PR #93, opened 2020-01-17
DP-11540: Remove the option to create an unlimited temporary unpublis…

lynnschumann:
DP-11540: Remove the option to create an unlimited temporary unpublished access token from the node form.

**Description:**
This removes the option to create an "unlimited" temporary unpublished access token from the node form.

**Jira:**
https://jira.mass.gov/browse/DP-11540

**To Test:**
- [ ] In "All Content", locate a node item that is not Published. In the Edit screen, in the right sidebar, expand "Temporary unpublished access."
- [ ] Check the options under "LIfetime". The "Unlimited" option should not appear.
- [ ] Test creating access tokens for the item and confirm that the tokens work as expected.

---

### DIFF
--- a/changelogs/DP-11540.yml
+++ b/changelogs/DP-11540.yml
@@ -1,0 +1,33 @@
+  # Write your changelog entry here.  Every PR must have a changelog.
+  #
+  # You can use the following types:
+  #   Added: For new features.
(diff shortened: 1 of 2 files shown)


ygannett (approved):
LGTM.

PR #107, opened 2020-01-28
DP-17135 Add ability to embed tableau visualizations to information details pages

a-username:
**Description:**
Adds a Tableau visualization paragraph type and field formatter to the Information Details content type to allow embedding visualizations by URL. 

**Jira:**
https://jira.mass.gov/browse/DP-17135

**To Test:**
- [ ] Edit an [Information details](http://mass.local/info-details/local-state-and-federal-laws-related-to-tobacco) page
- [ ] Go to the "Content" tab and expand the "Section content" area of an "Information details section"
- [ ] Verify the "Add tableau visualization" button is available and select it
- [ ] In the text box enter the embed URL of a visualization
- [ ] Save the node and scroll down to the section you've edited to verify the visualization loads.

**Screenshots/GIFs:**
2020-01-28_877x930

---


docroot/themes/custom/mass_theme/templates/content/node--info-details.html.twig
```diff
@@ -29,6 +29,7 @@
 {% set level = 1 %}
 
 {% block preContent %}
+  {{ attach_library('mass_theme/tableau') }}
```
ygannett: Can we make the library to be added only when Tableau content is available?

a-username: Good catch! I've added some logic to `mass_theme.theme` to check for the existence of an entity with the type `tableau_embed` to set the bool value of a variable which is used in the template to attach the library. You can test that out by loading an Info Details page with the "Tableau Visualization" field filled out and verifying that it displays then going to another page without the field filled out and opening the network tab in the inspector to make sure the tableau JS isn't added.

ygannett: Thank you so much for making this change.

ygannett (approved):
LGTM.

PR #110, opened 2020-01-28
Upgrade to Drupal 8.8

weitzman:
`composer update drupal/core drupal/redirect drupal/pathauto drupal/coder --with-dependencies`. FYI, Token and ctools modules were updated as dependencies.

**Jira:**
https://jira.mass.gov/browse/DP-17267

**To Test:**
- [ ] Check testing spreadsheet (see Jira issue).
- [ ] Redirects (redirect module was upgraded)
- [ ] Pathauto and aliases in general (D8.8 changed a lot)
- [ ] Failing patches were rerolled properly (see Jira subtasks)
- [ ] Devel and webprofiler modules are not enabled at an Acquia env

**To Do:**
- [x] Bring back admin content View patch DP-17273

---


danstafford:
Looks good @weitzman ! The only thing that catches my eye is that `path.alias_manager` and some other services have been deprecated. Should we adjust this in the custom modules?
https://www.drupal.org/node/3092086

weitzman:
Thanks. IMO we can pass this PR and open a followup for Drupal 9.x issues. Maybe start a new epic for it.

PR #115, opened 2020-01-30
[WIP] Use SSH to transfer database during ma:deploy

weitzman:
Avoids use of broken Acquia networking thats evident during http  transfer of backup file. This will fix our population of new database images, and fix the --refresh-db that we do during release workflow and ad hoc developer pushes.

**Jira:**
DP-17286

**To Test:**
- [ ] `dcexec scripts/ma-refresh-local -dpo --super`. This is what our DB populate workflow does at Circle.
- [ ] Run code below. This assures that developers and release workflow can ma:deploy successfully:
```
dcdrush ma:deploy cd ssh-deploy -vvv
dcdrush ma:deploy cd ssh-deploy -vvv  --refresh-db
```
The testing step above highlights an inconvenience that we have to deploy this code once in order for the --refresh-db to be functional. This would make our next release awkward. I think we should move the code deploy earlier in ma:deploy so that the refresh-db happens after code is deployed. The current order has been inconvenient in the past as well.

---

### DIFF
--- a/changelogs/DP-17286.yml
+++ b/changelogs/DP-17286.yml
@@ -1,0 +1,33 @@
+# Write your changelog entry here.  Every PR must have a changelog.
+#
+# You can use the following types:
+#   Added: For new features.
+#   Changed: For changes to existing functionality.
+#   Deprecated: For soon-to-be removed features.
+#   Removed: For removed features.
+#   Fixed: For any bug fixes.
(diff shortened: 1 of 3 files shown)


weitzman:
Lets not merge this unless the Acquia API fix is ineffective.

weitzman:
This approach was abandoned in favor of keeping with API.

PR #120, opened 2020-02-03
DP-17286: Fix ma-refresh-local for Acquia V2 API

rbayliss:
**Description:**
This PR updates ma-refresh-local so that it downloads the database using PHP (over an authenticated connection).  Hopefully, this will be the last step toward finally resolving our broken backup situation. 

**Jira:**
https://jira.mass.gov/browse/DP-17286

**To Test:**
- [ ] Run `docker-compose exec drupal scripts/ma-refresh-local --db-prep-only`
- [ ] You should end up with a fresh copy of the production database imported into MySQL, and no errors should be thrown. 

**Screenshots/GIFs:**

---

### DIFF
--- a/changelogs/DP-17286.yml
+++ b/changelogs/DP-17286.yml
@@ -1,0 +1,33 @@
+  # Write your changelog entry here.  Every PR must have a changelog.
+  #
+  # You can use the following types:
+  #   Added: For new features.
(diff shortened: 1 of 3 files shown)


weitzman:
We solve this in #121

PR #125, opened 2020-02-04
Bring back drush ma:release command

weitzman:
An old favorite Drush command is back. 

**Jira:**
https://jira.mass.gov/browse/DP-17243

**To Test:**
- [ ] Deploy this branch, then review its corresponding workflow in Circle to assure it succeeds: `dcdrush ma:release feature5 ma-release --ci-branch=ma-release --skip-maint --refresh-db  -v`. Or just see a green run from when I did it https://circleci.com/gh/massgov/openmass/3523

---

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,6 +93,22 @@ parameters:
     type: boolean
     default: false
 
+  # These params for currently used in drush ma:release
+  ma-release:
+    type: boolean
+    default: false
+  refresh-db:
+    type: string
+    default: ""
+  skip-maint:
+    type: string
+    default: ""
+  target:
+    type: string
+    default: ""
+  git-ref:
+    type: string
+    default: ""
 jobs:
   trigger_workflows:
     docker:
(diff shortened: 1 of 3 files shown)


rbayliss:
This looks great, and seems to function as advertised.  I'm somewhat concerned about the amount of workflows we're accumulating in Circle, though.  It seems tough to tell what's going to trigger what.  As an example, what happens if I use this command with `ci-branch` = `master`?  Would it run the `build_github_tag` workflow, since that's set up to trigger on `master`, and has no schedule condition?  I'd expect so, since that workflow doesn't have a `params.webhook` dependency.

I think our ability to test and debug this is pretty limited until this goes in though, since it'll need to be available on various branches for us to test the impact.  So I'd propose two options:

1. Audit all the workflows now and add the `params.webhook` dependency on anything that could possibly accidentally trigger given the right combination of branches. 
2. Merge this, then create a new `mayflower-dev-testing` branch and run `ma:release --ci-branch=mayflower-dev-testing`.  If I'm right about the issue, it'd also trigger the `deploy_mayflower_cd` workflow when we run it.  If that happens, we'd just need to go back and do the first option.

What do you think, @weitzman?

weitzman:
Yes, it would run the build_github_tag workflow. And thats always how ma:release has always worked. Before the recent allow_forks PR,  all workflows non-scheduled triggerred for each commit, and it was fine. My hope is that we are back to that point now.

Your suggestion to add dependencies on webhook sounds good, but I'm not sure how urgent it is, and I'm kinda worried about breaking a release process which is just getting its sea legs back.

weitzman:
I implemented 1. I'm fine with merging this with or without 1.

rbayliss:
This looks fantastic!  Thanks!

PR #132, opened 2020-02-06
DP-17413 home page related content display issue 

ygannett:
**Description:**
Render the related content paragraph in home page despite of any content types as "card" allowed to add to the paragraph.

- Count any added content types in the related content paragraph as allowed in the paragraph in the condition in `_get_layout()` in `mass_theme.theme` to have `$variables['layout'] = 'actionCards'` at L. 5038.

**Jira:**
https://jira.mass.gov/browse/DP-17413

**To Test:**
- [ ] Edit/add the **Related Content** under the **1up Stacked Row** in the **Rows** tab in Home page(3666).
- [ ] Add any combination of the following content types which allowed in the Related Content paragraph:
      - guide 
      - how to
      - location
      - location details
      - org page
      - right rail
      - service
      - service details
      - stacked layout
      - topic
- [ ] Check no matter what combination of content types added, the area is rendered as cards, no other format.
Mass_gov

**Screenshots/GIFs:**

---

### DIFF
--- a/changelogs/DP-17413.yml
+++ b/changelogs/DP-17413.yml
@@ -1,0 +1,33 @@
+  # Write your changelog entry here.  Every PR must have a changelog.
+  #
+  # You can use the following types:
+  #   Added: For new features.
+  #   Changed: For changes to existing functionality.
+  #   Deprecated: For soon-to-be removed features.
+  #   Removed: For removed features.
+  #   Fixed: For any bug fixes.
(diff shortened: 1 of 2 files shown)


a-username (approved):
Everything works on my end. Looks like this makes more sense than testing individual cases since every content type is added to the array with a 0 value anyway and only given a non-zero value when there are related items.

PR #133, opened 2020-02-07
DP-17054 Disallow query string in robots.txt

danstafford:
**Description:**
I added a line to the robots.txt file to disallow crawling of urls with "?auHash=".

**Jira:**
https://jira.mass.gov/browse/DP-17054

**To Test:**
I don't have any specific testing steps but here is a reference to confirm the matching pattern makes sense. 
https://developers.google.com/search/reference/robots_txt#url-matching-based-on-path-values

---


ygannett (approved):
Looks good.

PR #141, opened 2020-02-13
update pathologic config to include bare domain

lmirabile:
**Description:**
update the confirmation of the pathologic module to include the bare mass.gov domain, so that if someone enters such a path, it will be corrected to point to the relative root of the current environment (e.g. https://www.mass.gov on production).

**Jira:**
https://jira.mass.gov/browse/DP-16214

**To Test:**

**Screenshots/GIFs:**

---


danstafford (commented):
Looks good to me @lmirabile , I reproduced the link href test you outlined in the Jira ticket and got the same results.

PR #144, opened 2020-02-18
Fix nightcrawler

rbayliss:
**Description:**
Fixes Nightcrawler to work with Drupal 8.8.  This PR includes three adjustments:

* Updates the query Nightcrawler uses to determine the list of pages to crawl to match the way 8.8 stores paths. 
* Updates Nightcrawler to 1.2.0 to fix an issue where Nightcrawler wouldn't fail with a proper exit code if the setup step triggered an error. 
* Removes a workaround we put in place to force Nightcrawler to fail when `setup` steps fail.  This issue was causing the failure to swallow any error message that would have helped us debug the issue.  

**Jira:**
https://jira.mass.gov/browse/DP-17527

**To Test:**

**Screenshots/GIFs:**

---


changelogs/DP-17527.yml
```diff
@@ -0,0 +1,41 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
```
clairesunstudio: Should the comments be removed for auto release?

rbayliss: Good question.  We've had some issues recently, so I'll strip them out.

PR #146, opened 2020-02-19
DP-17546: Make Backstop more reliable

rbayliss:
**Description:**
This PR implements several changes to make Backstop testing more reliable:
1. Block Google Maps API requests outright.  This prevents us from using up map credits.
2. Replace Google Maps static images with placeholders. This makes the maps "look ok" to backstop in any environment, including ones that don't have the maps API setup.
3. Block Youtube requests outright. This fixes an issue where pages containing Youtube videos were timing out. 
3. Block the wait time AJAX request outright. Now the wait time widgets will always say "Unavailable", making them consistent for testing. 
4. Replace hero images with placeholders across the board. We used to only do this for the homepage (where they're random), but I've noticed some spurious diffs from images being partially loaded that are resolved by using a very lightweight placeholder.
5. Fixes an issue where the page could not be rendered in less than 2 seconds because we implement Google Optimize's asynchronous hiding snippet to hide the page until Optimize experiments are loaded.  Removing this allows us to snapshot the page a lot faster. 
6. Remove the extra 2s delay on page load.  This just isn't necessary, and slows down testing. 
7. Updates Backstop to the latest stable version for additional stability and performance.

**Jira:**
https://jira.mass.gov/browse/DP-17546

**To Test:**
- [ ] Run `ahoy backstop reference --target=prod`
- [ ] Run `ahoy backstop test --target=cd`
- [ ] Observe that any diffs that are shown are the result of content changes, not Backstop related issues. 

**Screenshots/GIFs:**

---

### DIFF
--- a/backstop/scripts/before.js
+++ b/backstop/scripts/before.js
@@ -7,10 +7,49 @@ module.exports = async (page, scenario, vp) => {
     page.on('request', interceptRequest);
 }
 
-// Block scripts that can cause problems or do analytics tracking.
-const banned = /(www\.googletagmanager\.com|script\.crazyegg\.com|google-analytics\.com|js-agent\.newrelic\.com|translate\.google\.com)/
+function escapeRegexp(string) {
+  return string.replace(/[.*+\-?^${}()|[\]\\]/g, '\\$&');
+}
+
+const banned = [
+  'www.googletagmanager.com',
+  'script.crazyegg.com',
+  'www.google-analytics.com',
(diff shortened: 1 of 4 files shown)


lmirabile (approved):
Everything ran successfully. There were only 10 errors: 3 on the home page (content changes); 6 changes in the vertical position of the "feedback" tab in the footer (an existing, separate thing), and one that was a false positive - there was no reason for the failure - no change could be detected. In other words, awesome!

PR #148, opened 2020-02-19
Dp 17411 scheduler moderation media

a-username:
**Description:**
Integrates the `scheduler_media` module with the `scheduler_content_moderation_integration` module to fix issues with scheduling documents.

**Jira:**
https://jira.mass.gov/browse/DP-17411

**To Test:**
- [x] Pull down the `DP-17411-scheduler-moderation-media` branch and run `drush updb` to ensure the new base fields are installed on media entities.
  - This can be verified (though it's not required) by running `show create table media_field_data;` inside the database and making sure the `publish_state` and `unpublish_state` columns exist.
- [x] Go to the media [listing page](http://mass.local/admin/content/media) and edit a published document
- [x] In the "Scheduling options" sub-section in the right sidebar enter a date/time in the future to unpublish the document (usually `[current-date +2minutes]` works best for testing) and save the document.
  - The media listing page only shows published documents, so searching for your document once it's unpublished won't yield results. For ease of testing you should keep the document open in another tab once you're done editing.
- [x] When the time for unpublishing the document has passed, run `drush cron`. You should see a message that reads `scheduler.NOTICE: Document: scheduled unpublishing of [your document here]` if another notice below that stating `file.ERROR: File '[file path associated with your document]' could not be copied because it does not exist` also appears this is to be expected since files are not available locally (however if testing on an environment where files are available this message should not appear).
- [x] Go back to the document and reload the page. Verify it has been unpublished by scrolling to the bottom and checking the current moderation state.

**Screenshots/GIFs:**

---

### DIFF
--- a/changelogs/DP-17411.yml
+++ b/changelogs/DP-17411.yml
@@ -1,0 +1,41 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
+# #############################################################################
+# You can use one of the following types:
+#  - Added: For new features.
+#  - Changed: For changes to existing functionality.
+#  - Deprecated: For soon-to-be removed features.
+#  - Removed: For removed features.
+#  - Fixed: For any bug fixes.
+#  - Security: In case of vulnerabilities.
+#
+# Format
+# #############################################################################
+# The format is crucial. Please follow the examples below. For reference, the requirements are:
+#  - All 3 parts are required and you must include "Type", "description" and "issue".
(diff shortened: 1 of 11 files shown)


danstafford:
Looks great @a-username ! I checked out the code and ran through the QA steps in my local as well and I was able to schedule a publish and unpublish successfully. The only thing I am seeing is a circleCI failure regarding config I think.

weitzman:
LGTM. 

For that config failure, the usual fix is to just commit the config change thats reported in the log.

PR #150, opened 2020-02-20
DP-17534 Cloudflare Terraform module version lock +  syntax upgrade

yriahi:
**Description:**
This PR locks Terraform's provider module (Cloudflare) to `~> 1.0`. `terraform init` was pulling the latest version of the Cloudflare Terraform provider module, in which some attributes were deprecated. This PR also upgrades Terraform code to version `0.12` syntax (rewrites pre-0.12 module source code for v0.12).

**Jira:**
https://jira.mass.gov/browse/DP-17534

**To Test:**
- [ ] First! a green build on this PR.
- [ ] Run the following commands:
```
$ git checkout DP-17534-cloudflare-terraform-syntax
$ # authenticate to aws via cli
$ cd cloudflare/terraform
$ terraform init
$ terraform plan
```

**Screenshots/GIFs:**

---


rbayliss:
Hey Youssef - I apologize that I didn't catch this before, but it looks like you did this work on the old way we were using Terraform (one massive config containing all page rules and zones).  The new way is split into "global" (only contains some account level settings and the 6 page rules we use), and "environment" (only contains the worker code and firewall rules for a specific "environment" or domain pair.  And both the "global" and "environnment" configuration are already Terraform 0.12 syntax, and locked to a specific version of the Cloudflare provider.  

I think we should repurpose this PR to be about deleting the old terraform files (the `.tf` files directly in `cloudflare/terraform`) so we don't hit this confusion again.

yriahi:
@rbayliss Thanks for the followup. The unused Terraform files have been removed and the change log note has been updated.

PR #119, opened 2020-02-03
DP-17273 add content moderation indices patch

a-username:
**Description:**
Update the content moderation views indeces patch to move the "extra" indeces to the `ModerationStateJoinViewsHandlerTrait`

**Jira:**
https://jira.mass.gov/browse/DP-DP-17273

**To Test:**
- [ ] Since this branch is based off of the `d88` branch you should be able to pull it down and just run `composer install`
- [ ] Verify the new patch ("patches/content-moderation-add-views-filter-indices.patch") applies cleanly

**Screenshots/GIFs:**
Use something like [licecap](http://www.cockos.com/licecap/) to capture gifs to demonstrate behaviors.

---

[Peer Review Checklist](https://github.com/massgov/opemass/blob/develop/docs/peer_review_checklist.md)

PR #152, opened 2020-02-20
DP-17568: Resolve editor-facing view slowness by reapplying the conte…

rbayliss:
…nt moderation index usage patch

**Description:**
This PR just reapplies the patch we used previously.  The patch itself was already reviewed in https://github.com/massgov/openmass/pull/119.

**Jira:**
https://jira.mass.gov/browse/DP-17568

**To Test:**
- [ ] I think just making sure that the patch is actually applied after this PR is enough here. I've verified it fixes the slowness issue locally, and it's worked before on prod. 

**Screenshots/GIFs:**

---

### DIFF
--- a/changelogs/DP-17568-2.yml
+++ b/changelogs/DP-17568-2.yml
@@ -1,0 +1,41 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
+# #############################################################################
+# You can use one of the following types:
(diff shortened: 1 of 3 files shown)


danstafford:
Hey @rbayliss I ended up just pulling this local and running composer to verify the patch installs (circleCI is very slow). I also visited http://mass.local/admin/home and verified the page loads quickly. I am merging this in.

PR #162, opened 2020-02-26
Add Drush commands for creating redirects from legacy files

weitzman:
My thinking is that after these commands are run, we run `drush @prod entity:delete node --bundle=legacy_redirects`. Then we already have a ticket to get rid of that content type in code.

Given that we plan to delete all D2D afterwards, I dont think that any detailed jounaling is needd here but am open to adding it if needed.

Note that the .htaccess change is needed because otherwise requests ending in .pdf will immediately 404 instead of redirect.

**Jira:**
https://jira.mass.gov/browse/DP-12397

**To Test:**
- [ ] Review log output after running both new Drush commands. Try to visit one of the newly created redirects. Verify that it goes to the right target.

---


docroot/modules/custom/mass_utility/src/Commands/MassUtilityCommands.php
```diff
@@ -266,7 +267,7 @@ public function report404Causes($fullpath_ga_404_csv_file) {
     $ga_404_csv_file = new \SplFileObject($fullpath_ga_404_csv_file);
     $ga_404_csv_file->setFlags(\SplFileObject::READ_CSV | \SplFileObject::READ_AHEAD | \SplFileObject::DROP_NEW_LINE | \SplFileObject::SKIP_EMPTY);
     foreach ($ga_404_csv_file as $row) {
-      list($event_404_action, $event_404_url, $event_404_count) = $row;
+      [$event_404_action, $event_404_url, $event_404_count] = $row;
```
rbayliss, line 271: Well aren't you fancy! 🎩

docroot/modules/custom/mass_utility/src/Commands/MassUtilityCommands.php
```diff
@@ -398,4 +399,95 @@ public function queueRevisionsCleanup(array $options = [
...
+   */
+  public function copyD2DRedirects($options = ['limit' => self::REQ, 'min-hits' => 50]) {
+    $query = $this->database->select('node__field_legacy_redirects_legacyurl', 'source')
+      ->fields('source', ['entity_id', 'delta']);
+    $query->addField('source', 'field_legacy_redirects_legacyurl_value', 'source');
+    $query->addField('target', 'field_legacy_redirects_ref_conte_target_id', 'target');
+    if ($options['limit']) {
+      $query->range(0, $options['limit']);
+    }
+    $query->innerJoin('node__field_legacy_redirects_ref_conte', 'target', 'source.entity_id = target.entity_id AND source.delta = target.delta');
```
rbayliss: Hmm, I think the delta condition here is unwanted.  My understanding is that Legacy redirects can have many sources, but only one target per node.  Meaning that every source redirects to target delta = 0.

docroot/modules/custom/mass_utility/src/Commands/MassUtilityCommands.php
```diff
@@ -398,4 +399,95 @@ public function queueRevisionsCleanup(array $options = [
...
+   * @option limit
+   * @option min-hits
+   */
+  public function copyD2DRedirects($options = ['limit' => self::REQ, 'min-hits' => 50]) {
+    $query = $this->database->select('node__field_legacy_redirects_legacyurl', 'source')
+      ->fields('source', ['entity_id', 'delta']);
+    $query->addField('source', 'field_legacy_redirects_legacyurl_value', 'source');
+    $query->addField('target', 'field_legacy_redirects_ref_conte_target_id', 'target');
+    if ($options['limit']) {
+      $query->range(0, $options['limit']);
```
rbayliss, line 419: Um, there's no pagination happening here AFAICT.  I think we either want to execute this query unlimited and just rely on PHP's efficiency in buffering records (which should be ok, based on past experience), or actually do pagination.

docroot/modules/custom/mass_utility/src/Commands/MassUtilityCommands.php
```diff
@@ -398,4 +399,95 @@ public function queueRevisionsCleanup(array $options = [
...
+    }
+    $query->innerJoin('node__field_legacy_redirects_ref_conte', 'target', 'source.entity_id = target.entity_id AND source.delta = target.delta');
+    $records = $query->execute()->fetchAll();
```
rbayliss, line 424: Before we execute... can we limit this by `status` = 1 and `field_legacy_redirect_env_value` = 0?

docroot/modules/custom/mass_utility/src/Commands/MassUtilityCommands.php
```diff
@@ -398,4 +399,95 @@ public function queueRevisionsCleanup(array $options = [
...
+    foreach ($records as $record) {
+      $record->source = parse_url($record->source, PHP_URL_PATH);
+      $record->target = '/node/' . $record->target;
+      if ($this->isActive($record->source, $options)) {
+        $this->saveRedirect($record);
```
rbayliss, line 429: I'm getting some duplicate entries on save.  I think we should validate that no redirect already exists for the source path.

rbayliss: Along those same lines... should we check to make sure that the path is not already aliased before creating?  Some of the previously legacy prefixes now go to Drupal, and I'm not sure we've always had the validation that a legacy redirect actually points to a legacy prefix, so it's theoretically possible that we could have some overlap here.

docroot/modules/custom/mass_utility/src/Commands/MassUtilityCommands.php
```diff
@@ -398,4 +399,95 @@ public function queueRevisionsCleanup(array $options = [
...
+      ->fields('source', ['entity_id', 'delta']);
+    $query->addField('source', 'field_legacy_redirects_legacyurl_value', 'source');
+    $query->addField('target', 'field_legacy_redirects_ref_conte_target_id', 'target');
+    if ($options['limit']) {
+      $query->range(0, $options['limit']);
+    }
+    $query->innerJoin('node__field_legacy_redirects_ref_conte', 'target', 'source.entity_id = target.entity_id AND source.delta = target.delta');
+    $records = $query->execute()->fetchAll();
+    foreach ($records as $record) {
+      $record->source = parse_url($record->source, PHP_URL_PATH);
```
rbayliss: I wonder if we need to deal with query strings?  Not sure

rbayliss: Also, we need to drop trailing slashes from the source, if there are any.

docroot/.htaccess
```diff
@@ -179,10 +179,10 @@ AddType application/octet-stream .rte
...
 
   # Markdown and pdf files, which are used for documentation files with many names.
-  RewriteCond %{REQUEST_URI} !^/files/
-  RewriteCond %{REQUEST_URI} !^/sites/.+/files/
-  RewriteCond %{REQUEST_URI} !^/hq2/
-  RewriteRule (.*)\.(md|MD|pdf|markdown) - [L,R=404]
+#  RewriteCond %{REQUEST_URI} !^/files/
+#  RewriteCond %{REQUEST_URI} !^/sites/.+/files/
+#  RewriteCond %{REQUEST_URI} !^/hq2/
+#  RewriteRule (.*)\.(md|MD|pdf|markdown) - [L,R=404]
```
rbayliss: Can we leave the other file extensions blocked?

rbayliss (changes requested):
Getting there. Some requests below.

docroot/modules/custom/mass_utility/src/Commands/MassUtilityCommands.php
weitzman: I think that this query wants to load of pairs of source and target from the D2D content type. So we do want to make sure we match deltas. If you aren't convinced, lets discuss. This area is a bit hazy for me too.

docroot/modules/custom/mass_utility/src/Commands/MassUtilityCommands.php
weitzman: Agreed. The --limit is there just for development purposes. I've added a note to that effect in the Help for the command.

docroot/modules/custom/mass_utility/src/Commands/MassUtilityCommands.php
weitzman: I'm handling the exceptions that happens with a duplicate save. Do you think we should do more, like try not to save it at all?

docroot/modules/custom/mass_utility/src/Commands/MassUtilityCommands.php
weitzman: I didn't see any querystrings in the URLs I looked at.

Drop trailing slashes right before we save as a Redirect entity? Or earlier?

docroot/.htaccess
weitzman: Done

docroot/modules/custom/mass_utility/src/Commands/MassUtilityCommands.php
rbayliss: I think right here is good, since we'll have direct access to the path (rather than path + query string or whatever).

docroot/modules/custom/mass_utility/src/Commands/MassUtilityCommands.php
rbayliss: Actually, nah.  That's fine. It should be relatively limited in scope anyway, I'd hope.

docroot/modules/custom/mass_utility/src/Commands/MassUtilityCommands.php
rbayliss: I don't think that's right - here's a screenshot of a crazy example of the source field:
Screen Shot 2020-03-03 at 5 27 06 PM
Meanwhile the destination field is a single value:
Screen Shot 2020-03-03 at 5 29 04 PM

rbayliss:
This looks good to me!  This will just make the Drush commands available - once they make it to prod, they'll be runnable with:
```
drush @prod copyd2d
drush @prod copyLegacy
```

weitzman:
Thanks. Lets run the commands on stage first, and make sure we are happy before proceeding.

PR #166, opened 2020-02-26
Update aliases

ygannett:
Update update self.site.yml location in .circleci/config.yml.


weitzman:
Fixed elsewhere yesterday

PR #170, opened 2020-02-27
[TEST] DP-17484 info details section container

ygannett:
**Description:**

- Switch the Tableau template to use MF template figure.twig.
- Clean up templates for rich text content to avoid nested `div.ma__rich-text`.

**Jira:**
https://jira.mass.gov/browse/DP-****

**To Test:**

**Screenshots/GIFs:**

---


docroot/themes/custom/mass_theme/templates/paragraphs/paragraph--rich-text.html.twig
```diff
@@ -36,8 +36,8 @@
  * @ingroup themeable
  */
 #}
-{% embed "@organisms/by-author/rich-text.twig" %}
+{# {% embed "@organisms/by-author/rich-text.twig" %} #}
```
rbayliss, line 39: Huh... I'm worried about how this change might ripple out to other content types.  This paragraph type is used pretty heavily, right?  Maybe we should do some visual regression testing to figure out the scope of the change.

rbayliss (commented):
Hey Yukiko, 
In general, I love this approach.  Super simple, and seems to work beautifully.  My only concern is below.

docroot/themes/custom/mass_theme/templates/field/mass_content_tableau_embed.html.twig
```diff
@@ -12,13 +12,24 @@
...
+      "src": url
+    },
+    "title": "",
+    "caption": "This is an optional caption that can wrap to multiple lines."
+  }
+} %}
+{# <div class="ma_tableau_container" style="max-width:100%">
   <div id="{{ randId }}"></div>
 </div>
 <script type="text/javascript">
```
rbayliss, line 20: Oh right - I think we need to keep the tableau embed code in Drupal, also.  My preference would be for Mayflower to allow us to inject raw HTML into the figure template, as I outlined in my response to massgov/mayflower#954

rbayliss: In my proposed change on the other PR, here's how it would look on the Drupal side:
```twig
{% set tableau %}
<div class="ma_tableau_container" style="max-width:100%">
  <div id="{{ randId }}"></div>
</div>
<script type="text/javascript">
  window.addEventListener("DOMContentLoaded", function() {
    const url = "{{ url }}"
    new tableau.Viz(document.getElementById("{{ randId }}"), url);
  });
</script>
{% endset %}
{% include "@atoms/09-media/figure--dataviz.twig" with {
  "figure": {
    "align": "right",
    "content": tableau
    "title": "",
    "caption": "This is an optional caption that can wrap to multiple lines."
  }
} %}
```

PR #173, opened 2020-02-27
DP-17564 fix GitHub tagging issue

yriahi:
**Description:**
This PR fixes how we read the `changelog-body.txt` by adding new lines `\n` before building the JSON object used in the curl POST. Previously, this was causing a JSON parsing error.

**Jira:**
https://jira.mass.gov/browse/DP-17564

**To Test:**
- [ ] Generate a GitHub token for testing purposes and remember to delete it later!
- [ ] Overwrite the content of `tag-release-github` with the code below
```
#!/usr/bin/env bash

# exit when any command fails
set -e

GITHUB_MASSGOV_BOT_TOKEN="**********************************"

# Use AWK's output record separator (ORS) to avoid JSON parsing issue.
# https://linux.die.net/man/1/awk
cat ./scripts/changelog-body.txt |  awk '{print}' ORS='\\n' > /tmp/changelog-body.txt

# Create post data
generate_post_data()
{
  cat <<EOF
  {
  "tag_name": "${TAG}",
  "target_commitish": "master",
  "name": "${TAG}",
  "body": "$(cat /tmp/changelog-body.txt)",
  "draft": false,
  "prerelease": false
  }
EOF
}

# Create a release on GitHub for the tag that was just created.
curl -X POST --data "$(generate_post_data)" --fail -u massgov-bot:"${GITHUB_MASSGOV_BOT_TOKEN}" -H "Content-Type:application/json" https://api.github.com/repos/<YOUR_GITHUB_ID>/<YOUR_GITHUB_PROJECT>/releases
```

**Screenshots/GIFs:**

---

- [ ] Execute the script 
- [ ] This should have created a release tag on one of your projects.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,7 +83,7 @@ commands:
         type: string
         default: "/var/www/code/drush/sites/self.site.yml"
     steps:
-      - run: "curl -f -o << parameters.destination >> -L https://$GITHUB_DS_INFRASTRUCTURE_TOKEN@raw.githubusercontent.com/massgov/massgov-internal-docs/master/self.site.yml"
+      - run: "curl -f -o << parameters.destination >> -L https://$GITHUB_MASSGOV_SELF_SITE_YML_TOKEN@raw.githubusercontent.com/massgov/massgov-internal-docs/master/self.site.yml"
 
 # See https://medium.com/labs42/monorepo-with-circleci-conditional-workflows-69e65d3f1bd0
 parameters:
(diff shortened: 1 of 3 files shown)


rbayliss (approved):
This looks like a good quick fix.  I think ultimately, we probably shouldn't be doing this work with Bash, though.  Consider what happens when the changelog contains a `"` character - it's going to cause a termination of the JSON string.  Anyway, I'm hopeful that this will get us through the next release, and we can consider how to make it more bulletproof.

PR #183, opened 2020-03-05
DP-17535 Remove ability to create document on IEF form in List item document paragraph

danstafford:
**Description:**
The "List item document" paragraph which appears on the Curated List node has an entity reference field that references document media entities. This field uses the inline-entity-form widget to reference these entities. I was able to uncheck "create new media entities" on the widget settings on this page `/admin/structure/paragraphs_type/list_item_document/form-display` but [this patch](https://www.drupal.org/project/inline_entity_form/issues/2913571) was needed to allow disabling of editing of referenced entities.

As a note this work was done on the "List item document" paragraph and not specifically on the Curated List node. So this change will take effect everywhere this paragraph is used which includes other paragraphs and node types. If we think this is not acceptable or may cause a problem lets close this PR and think of another way to handle this (maybe with CSS or something?). 

Patch used: https://www.drupal.org/project/inline_entity_form/issues/2913571

**Jira:**
https://jira.mass.gov/browse/DP-17535

**To Test:**

1. - [ ] Edit a curated list node
2. - [ ] Add or edit a "List - Manual" paragraph
3. - [ ] Verify you can add/remove the document media entities under List Item Documents 
4. - [ ] Verify you can NOT create/edit the document media entities under List Item Documents 
5. - [ ] Verify the forms save correctly
6. - [ ] Verify the documents display as expected on the page

**Screenshots/GIFs:**

---

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -319,7 +319,8 @@
             },
             "drupal/inline_entity_form": {
                 "Fixes field label when a new entity is created inline (https://www.drupal.org/project/inline_entity_form/issues/2842744)":
-                "https://www.drupal.org/files/issues/inline_entity_form-no_label_required_field_with_no_entries-2842744.patch"
+                "https://www.drupal.org/files/issues/inline_entity_form-no_label_required_field_with_no_entries-2842744.patch",
+                "Add a setting to enable/disable inline editing of existing entities": "https://www.drupal.org/files/issues/2018-10-11/ief_edit_existing_2913571_7.patch"
             },
             "drupal/key": {
                 "Stop adding 'extensions' cache tag on unrelated config objects (https://www.drupal.org/project/key/issues/3080971)": "https://www.drupal.org/files/issues/2019-09-12/no-tag_1.patch"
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -319,7 +319,8 @@
             },
             "drupal/inline_entity_form": {
                 "Fixes field label when a new entity is created inline (https://www.drupal.org/project/inline_entity_form/issues/2842744)":
-                "https://www.drupal.org/files/issues/inline_entity_form-no_label_required_field_with_no_entries-2842744.patch"
+                "https://www.drupal.org/files/issues/inline_entity_form-no_label_required_field_with_no_entries-2842744.patch",
```
weitzman: The patch specifics should be on one line. I know this problem predates this PR but lets fix it now please.

composer.json
```diff
@@ -319,7 +319,8 @@
...
+                "Add a setting to enable/disable inline editing of existing entities": "https://www.drupal.org/files/issues/2018-10-11/ief_edit_existing_2913571_7.patch"
```
weitzman: Please link to the comment where this file is uploaded. See other patches in this file for an example of how we usually do this. I know the info is in the filename but for consistency lets add it to the left hand side as well.

weitzman (approved):
Just 2 super minor comments about `patches` standards. Lets merge once thats done. The functionality changed here is Approved.

PR #194, opened 2020-03-13
DP-17784: Fixes for media downloads

rbayliss:
**Description:**
This PR shortens the browser cache lifetime for media download URLs to 60 seconds, including all 404s received on a path that looks like a media download URL.  This is to prevent end users from getting stuff stuck in their browser for a really long time, and does not affect the edge lifetime. 

This has the side effect of shortening the cache lifetime for other types of content (content that is not one of our known "static" types, or served through Drupal).  We're not worried about this, because static content still gets a really long lifetime, and the edge lifetime isn't changed at all.

**Jira:**
https://jira.mass.gov/browse/DP-17784

**To Test:**
- [ ] Visit `https://wwwcf.digital.mass.gov/media/12312312321312/download` in your browser and verify that the browser cache lifetime (max age value on the `Cache` header) is 60 seconds, despite being a 404 page.  Do the same test for a valid media download like `https://wwwcf.digital.mass.gov/media/1/download`

**Screenshots/GIFs:**

---

### DIFF
--- a/changelogs/DP-17784.yml
+++ b/changelogs/DP-17784.yml
@@ -1,0 +1,41 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
+# #############################################################################
+# You can use one of the following types:
+#  - Added: For new features.
+#  - Changed: For changes to existing functionality.
+#  - Deprecated: For soon-to-be removed features.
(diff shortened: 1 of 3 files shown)


a-username (approved):
Looks good to me!

weitzman (approved):

PR #199, opened 2020-03-16
DP-17809: Resolve media download caching on initial creation

rbayliss:
**Description:**
This PR fixes an issue where the media download URL for the aliased path is not purged immediately when a document is first created. This turned out to be just an issue with hook ordering - pathauto fired after our hooks, so we didn't have access to the aliased path yet.  

**Jira:**
https://jira.mass.gov/browse/DP-17809

**To Test:**

**Screenshots/GIFs:**

---


a-username (approved):
that all makes sense to me!

PR #211, opened 2020-03-20
DP-17810 Adjust json output to fix alert links to ids

danstafford:
**Description:**
The link was not pointing at the correct html id on the page because the json output was not correct. Apparently the id of the paragraph is under `drupal_internal__id` and not simply id. 

If you visit [this endpoint](https://www.mass.gov/jsonapi/node/alert?page[limit]=250&sort=-changed&include=field_target_pages_para_ref,field_alert&filter[status][value]=1&fields[node--alert]=title,changed,entity_url,field_alert_severity,field_alert,field_target_pages_para_ref,field_alert_display&fields[paragraph--emergency_alert]=id,changed,field_emergency_alert_timestamp,field_emergency_alert_message,field_emergency_alert_link,field_emergency_alert_content&fields[paragraph--target_pages]=field_target_content_ref) used by the alerts block you will see that there is no id in the json output.

alert_endpoint

If you visit [this endpoint](https://www.mass.gov/jsonapi/paragraph/emergency_alert) for the emergency_alert paragraph you will see that the id is actually called `drupal_internal__id`. 

paragraph_endpoint

So I adjusted the endpoint on the alerts block to use the correct field and adjusted the `ajax-pattern-alerts.js` file to use the correct field.

**Jira:**
https://jira.mass.gov/browse/DP-17810

**To Test:**
- [ ] Visit the homepage or any page that is showing the alerts bar
- [ ] Verify the links that point to the alerts page no longer have "undefined" as an id. They will have a paragraph id, something like this http://mass.local/alerts#1455786
- [ ] If you click the link you should link to the alerts page and be positioned down the page to that id

**Screenshots/GIFs:**

---

### DIFF
--- a/changelogs/DP-17810.yml
+++ b/changelogs/DP-17810.yml
@@ -1,0 +1,41 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
+# #############################################################################
+# You can use one of the following types:
+#  - Added: For new features.
+#  - Changed: For changes to existing functionality.
+#  - Deprecated: For soon-to-be removed features.
+#  - Removed: For removed features.
+#  - Fixed: For any bug fixes.
+#  - Security: In case of vulnerabilities.
+#
+# Format
+# #############################################################################
+# The format is crucial. Please follow the examples below. For reference, the requirements are:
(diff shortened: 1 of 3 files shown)


ygannett (approved):
LGTM.

PR #217, opened 2020-03-23
Fix broken tests and use loginTrait.

weitzman:
Unblock uid=1 before trying to login.

**Jira:**
https://jira.mass.gov/browse/DP-17924

**To Test:**
- [ ] None. If its green its good.

---


ygannett (approved):
LGTM.

PR #225, opened 2020-03-26
DP-17863 private documents

a-username:
**Description:**
Uploads document files to the private filesystem.

**Jira:**
https://jira.mass.gov/browse/DP-17863

**To Test:**
- [x] Go to http://mass.local/media/add/document                                    
- [x] Enter a title and upload a document to the field labeled "Select a file"        
  - Note: if you get an error stating the file cannot be uploaded this is likely because the ownership of your local private filesystem isn't correct (i.e. mine was defaulting to root:www-  data when creating the directories). To fix this run `ls -l` inside sites/default/files and check what the ownership is set to for the public directory; this should be something like `sudo  chmod -R [user-from-public-directory]:www-data`. After resetting the user, verify you can upload a document.
- [x] After uploading the document, use the link that is generated in the field to open the document in a new tab.
- [x] Verify this takes you to the private filesystem; the path should start with "system/files/documents"                                                        
- [x] Open the file URL in a private window. Verify you are unable to see the document.
- [x] After completing steps 4 and 5, save the media entity as a draft.                                      
- [x] Go to the media entity and repeat steps 4 and 5 using the `/download` link to the document on the draft.
- [x] Edit the draft media entity and set it to "Published". Save the entity.    
- [x] From the published media entity use the `/download` link to the document to verify the file is accessible from the public filesystem.
- [x] Edit the published media entity and upload a new document. Repeat the steps from 3 to 7 to ensure that during and after the editing process while the media entity is still a draft, any associated documents remain in the private filesystem.

**Screenshots/GIFs:**

---


weitzman (changes requested):
I think that this merits a test.

I think we are at or near a complexity tipping point. Maybe we should switch to private files everywhere after the covid traffic subsides? /cc @rbayliss

docroot/modules/custom/mass_caching/mass_caching.module
```diff
@@ -130,3 +155,17 @@ function mass_caching_entity_insert(EntityInterface $entity) {
...
+ *
+ * @param string $uri
+ *   A stream, referenced as "scheme://target" or "data:target".
+ *
+ * @return bool
+ *   TRUE if the uri is private FALSE if it is anything else.
+ */
+function _mass_caching_uri_is_private($uri) {
+  $file_system = \Drupal::service('file_system');
+  return strpos($file_system->uriScheme($uri), 'public');
```
rbayliss: Isn't this the opposite condition from what we want (returns true if it's public)?  Also, let's do this the official way:
```
$stream_wrapper_manager = \Drupal::service('stream_wrapper_manager');
$destination_scheme = $stream_wrapper_manager::getScheme($file->getFileUri());
return $destination_scheme === 'private';
```

rbayliss:
@weitzman - I think this PR is not quite as complex as it looks. Most of the changes here are from shifting code from `mass_utility` to `mass_media`, where it belongs.  I do agree it would be great to add a test for this functionality, along two axes: 
1. Public file cache purging (already partially implemented - just need to add an assertion that a private file doesn't trigger invalidation).  
2. Media entity file management (not implemented yet).

I don't want to slow down what's a pretty high priority fix though, so I suggest that we manually test this one, then come back right after and implement the tests, which should be pretty quick.

rbayliss:
Also, re: private files - I don't think we should switch wholesale.  Just to be clear about what's happening in this PR, this is just a continuation of the work to make unpublished media files private - in order to reliably implement that, we had to set the "default" for media to be private, but they're still made public the moment they're published.

We could test, but I think it would be disastrous for us to send all of the public file traffic through Drupal. Right now, origin file access doesn't even hit PHP at all.  If we make them private, I suspect we'd end up with no-cache headers on the responses, and a full bootstrap of Drupal for every file access.

docroot/modules/custom/mass_caching/mass_caching.module
weitzman: Done

weitzman:
> 
> 1. Public file cache purging (already partially implemented - just need to add an assertion that a private file doesn't trigger invalidation).
> 1. Media entity file management (not implemented yet).

1. is now implemented
2. can't easily be an ExistingSiteTest because we generally would create the file and media entities programmatically and thus we "test the test", and instead of testing form based upload system. I notice that we have a Behat test at https://github.com/massgov/openmass/blob/develop/features/documents.feature. I could add a check that the uploaded document is private. I dont see any Behat tests that test content. moderation so not sure how to test that subsequent drafts stay private ... Also, https://github.com/massgov/openmass/blob/develop/features/documents.feature#L46 is passing with this branch. Is that right?, or did we miss a is_private() check?

docroot/modules/custom/mass_media/mass_media.module
```diff
@@ -151,3 +153,84 @@ function mass_media_media_access(EntityInterface $entity, $operation, AccountInt
...
+function mass_media_file_move(FileInterface $file, FileInterface $source) {
+  // Add temporary logging due to https://jira.mass.gov/browse/DP-16025
+  $file_system = \Drupal::service('file_system');
+  $msg_file = sprintf('File, Id: %s, Uri: %s Filename: %s, Realpath: %s', $file->id(), $file->getFileUri(), $file->getFilename(), $file_system->realpath($file->getFileUri()));
+  $msg_source = sprintf('Source, Id: %s, Uri: %s Filename: %s, Realpath: %s', $source->id(), $source->getFileUri(), $source->getFilename(), $file_system->realpath($source->getFileUri()));
+  Drupal::logger('mass_media')->debug("$msg_file.\n\n$msg_source");
+
+  // Delete the original if it's not in use elsewhere.
+  $delete_source_document = FALSE;
+  $file_references = file_get_file_references($file);
```
weitzman, line 219: My understanding is that the file usage API is not reliable. Not sure what else to do though - https://www.drupal.org/project/drupal/issues/2821423

rbayliss: Yeah.  This is old logic though - we've been using it for over a year now.  In practice, there's a 1-1 relationship between a media entity and its file, so I don't think we're likely to run into issues with it.

rbayliss:
Tests look good to me.  @weitzman, I'm assuming you've tested the update hook and the file upload capability as part of this PR.  If so, I think we can go ahead and merge.

PR #226, opened 2020-03-26
Attach workspace to fix githug tagging during release

weitzman:
**Jira:**
https://jira.mass.gov/browse//DP-17959

**To Test:**
- [ ] Not possible until after this is in develop

---

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -188,8 +188,8 @@ jobs:
     <<: *container_config
     working_directory: /var/www/code
     steps:
+      - attach_workspace: {at: /var/www}
       - *no_host_check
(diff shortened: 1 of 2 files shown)


a-username (commented):
This makes sense to me, the only thing I guess I have question about is why we would need to keep the `working_directory` variable in the `tagging_github` step if we're not checking out to it any more? It's entirely possible that I'm missing something obvious in which case I'll go ahead and approve this if that's not something that matters.

weitzman:
pretty sure we still need working_directory. attach_workspace means "mount all the files from the last job onto my job's container". working_directory specifies the start directory when running steps for the job.

PR #231, opened 2020-03-27
DP-17972 Add new banner to info_details page

danstafford:
**Description:**
I added a new banner image field to the info_details page. When an image is uploaded an illustrated-header style header is used. The title and short-description fields appear on this header along with the new image. 

If an image is NOT uploaded to this new banner image field, then the standard page-header style header is used. 

This image field only appears for admins and content administrators (content_team role).

**Jira:**
https://jira.mass.gov/browse/DP-17972

**To Test:**
- [ ] Login as an admin or content administrator
- [ ] Visit an info_details page
- [ ] The header should appear exactly as before (it still should use the page-header template)
- [ ] Edit this info_details page
- [ ] You should see a Banner Image field - upload an image here and save the node
- [ ] Now there should be a new illustrated-header style header being used
- [ ] Now logout and login with a user without admin or content administrator roles
- [ ] You should not see the new Banner Image field
- [ ] You should still be able to edit and save the node

**Screenshots/GIFs:**

---

### DIFF
--- a/changelogs/DP-17972.yml
+++ b/changelogs/DP-17972.yml
@@ -1,0 +1,41 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
+# #############################################################################
+# You can use one of the following types:
+#  - Added: For new features.
+#  - Changed: For changes to existing functionality.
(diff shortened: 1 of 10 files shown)


ygannett (approved):
LGTM.

PR #236, opened 2020-03-30
Change config for TFA module to disallow skipping setup

lmirabile:
**Description:**
Change the "skip validation" configuration option for TFA from 3 to 0

**Jira:**
https://jira.mass.gov/browse/DP-17994

**To Test:**
Check out this branch on a local environment, where you can use mailhog (just go to mailhog.mass.local in a different tab & voila, you'll see any mail sent by your local environment).
- [ ] Create a new user with the role "editor" or "author" using any email address (no real email will be sent) and make sure to check the "notify user" option
- [ ] Check mailhog.mass.local and copy the link from the new user email
- [ ] Open the link in an incognito tab or a different browser
- [ ] Set a password
- [ ] Confirm you successfully log in and can go to your user profile's security tab to set up TFA - but don't set it up
- [ ] Try to log in as that user again and confirm you aren't able to 
- [ ] Repeat the process, this time with a user that had been blocked but that you unblock

**Screenshots/GIFs:**

---


docroot/modules/custom/tfa_unblock/tfa_unblock.module
```diff
@@ -0,0 +1,6 @@
+<?php
```
weitzman: A .module is not needed as of Drupal 8. Lets add one when we need it.

docroot/modules/custom/mass_admin_pages/src/Controller/AuthorHomePageController.php
```diff
@@ -16,6 +16,15 @@ class AuthorHomePageController extends ControllerBase {
    *   A simple renderable array.
    */
   public function authorHome() {
+    $user_data = \Drupal::service('user.data');
+    if (empty($user_data->get('tfa', \Drupal::currentUser()->id(), 'tfa_totp_seed'))) {
```
weitzman, line 20: Is there not an API function for getting this info? Getting it from user_data is very low level. Its rarely the right way.

arthurbaghdas: This is the official documentation for that https://www.drupal.org/node/1852360

docroot/modules/custom/tfa_unblock/tfa_unblock.module
arthurbaghdas: Ok, will remove now

weitzman (approved):

weitzman (approved):
1. @joeg8612 This PR will add 3 similar changelog entries to our release emails and release notes. Is that OK?
1. Are there any emails that need to be changed, such as the welcome email that is sent upon registration?

Neither of thee are dealbreakers so this is approved.

joeg8612:
> 1. @joeg8612 This PR will add 3 similar changelog entries to our release emails and release notes. Is that OK?
> 2. Are there any emails that need to be changed, such as the welcome email that is sent upon registration?
> 
> Neither of thee are dealbreakers so this is approved.

The three changelog entries are fine.  I have reviewed all emails.

PR #245, opened 2020-04-01
DP-18030 Directions link text capitalization

ygannett:
**Description:**
Replace the link text in address of contact in org page with text value "Directions" with uppercase **D** originally the field value "drections" with lowercase **d**.

**Jira:**
https://jira.mass.gov/browse/DP-18030

**To Test:**
- [ ] Go to /orgs/qag-digital-services.
- [ ] Find the link text "directions" in the address component of contact has uppercase **D** instead of lowercase **d**.

**Screenshots/GIFs:**
Dropdown

_QAG_Digital_Services___Mass_gov

---


weitzman (approved):
Its not a big deal, and this is approved for merge as is.. But FYI these are other ways to meet the requirement. The advantage of the other ways is that they dont override the field label, so a developer who changes the field label is not surprised when it has no effect on the UI.


1. [Capitalize via CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform) 
1. [Capitalize via Twig](https://stackoverflow.com/questions/31199959/twig-capitalize-makes-other-letters-small#33260533).

ygannett:
@weitzman Thank you so much for the note.  The CSS option would affect other contact information links other than address, so I used a twig filter.

PR #254, opened 2020-04-06
DP-17976 stop unpublished nodes from appearing in the page flipper

a-username:
**Description:**
Keeps unpublished nodes out of the page flipper/pager on info details pages.

**Jira:**
https://jira.mass.gov/browse/DP-17976

**To Test:**
- [x] Find an existing binder to test with
- [x]  Open one of the pages in the middle of the binder
- [x] Scroll to the bottom where the binder pager is displayed.
- [x] Open both the "Previous" and "Next" pages referenced in different tabs
- [x] On one of them, set the publication status to "Unpublished"
- [x] Reload the page and verify the unpublished node is no longer displayed on the other pages where it was either "previous" or "next" before
- [x] Re publish that page; you should now have 3 nodes referenced by the binder open in tabs
- [x] Edit the binder and set its publication status to "Unpublished"
- [x] Visit the three open pages, refresh them, and confirm that they don't reference the binder any more at the top of the page, and don't show prev/next links at the bottom. 
--------
Cache-tags:
cache-tags for nodes referenced in the binder pager have been added to the node. To test:
- [x] Go to an info details page and load it with the network tab open
- [x] After it has loaded, select the document from the list
- [x] Verify that in the "X-Drupal-Cache-Tags" response header, cache tags matching the pattern "node:[nid-of-previous-pager-node]" and "node:[nid-of-next-pager-node]" appear

**Screenshots/GIFs:**

---


michael-wojcik:
@a-username LGTM! Just need to fix the above « Error YAMLException »

michael-wojcik (approved):
LGTM, save for « Error YAMLException »

PR #265, opened 2020-04-10
DP-16911 Update private_files_download_permission from 2.0 to 2.2

danstafford:
**Description:**
I updated `private_files_download_permission` from 2.0 to 2.2. I was never able to reproduce the conditions needed for the 500 error to occur locally to test if updating this module actually solved the problem. But this is a minor update that addresses bugs in this module so I don't think it should be a problem if we merge this in and see if it fixes these 500 errors on the live site.

https://jira.mass.gov/browse/DP-16911

**To Test:**
- [ ] These testing steps are to verify the module continues to work correctly after this update
- [ ] Visit http://mass.local/admin/content/media and edit the first media entity
- [ ] Make sure this media entity is published
- [ ] Copy the path to the file. It should look something like this:
http://mass.local/sites/default/files/documents/path-to-file
- [ ] Open this file in an incognito window and verify you can access the file
- [ ] Now unpublish this media entity
- [ ] Copy the path to the file. It should looks something like this:
http://mass.local/system/files/documents/path-to-file
- [ ] Open this file in an incognito window and verify you get a 403 error because you shouldn't be able to access this file on this path
- [ ] And if you visit http://mass.local/sites/default/files/documents/path-to-file when the media entity is unpublished it should be a 404 because the file no longer resides in this path

copy_file_path

**Screenshots/GIFs:**

---

### DIFF
--- a/changelogs/DP-16911.yml
+++ b/changelogs/DP-16911.yml
@@ -1,0 +1,41 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
+# #############################################################################
+# You can use one of the following types:
+#  - Added: For new features.
+#  - Changed: For changes to existing functionality.
+#  - Deprecated: For soon-to-be removed features.
+#  - Removed: For removed features.
+#  - Fixed: For any bug fixes.
+#  - Security: In case of vulnerabilities.
+#
+# Format
+# #############################################################################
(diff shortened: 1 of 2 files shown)


weitzman:
This got done in another PR https://github.com/massgov/openmass/pull/284/files#diff-1da2c7edc898c70e5a79a9997c98ceccR7263

PR #267, opened 2020-04-13
DP-13246 decision tree buttons

ygannett:
**Description:**
Styled decision tree buttons below 620px width to line up vertically instead of horizontally.
Instead of flow them like horizontally, style similar to the callout link style below 620px, so the second button doesn't have left margin when it flows to the next line(See the "before" screenshot below).

https://jira.mass.gov/browse/DP-13246

**To Test:**
- [ ] Go to any decision tree page with buttons. Ex) /decision-tree/how-to-find-historic-court-records#s=456366

- [ ] Set the page width below 620px.

- [ ] Find the button display matches the "after" screenshot below.

**Screenshots/GIFs:**
Before
belore

After
after

Callout link below 620px
Mass_Gov

---

### DIFF
--- a/changelogs/DP-13246.yml
+++ b/changelogs/DP-13246.yml
@@ -1,0 +1,41 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
+# #############################################################################
+# You can use one of the following types:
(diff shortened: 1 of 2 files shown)


danstafford (approved):
Look great!

PR #268, opened 2020-04-13
[DP-14628] author message subject length

michael-wojcik:
**Description:**
Truncates node title as needed when contacting authors.

https://jira.mass.gov/browse/DP-14628

**To Test:**
- [ ] Edit a node with a long title that exceeds 100 characters (e.g., `/service-details/local-options-relating-to-property-taxation-cpa-meals-and-room-occupancy`)
- [ ] Click the "Contact the author" link in the top righthand corner of the form
- [ ] Confirm that the title in the subject line was automatically truncated
- [ ] Confirm that the form can be submitted without error

**Screenshots/GIFs:**
Demo of truncating functionality

---

### DIFF
--- a/docroot/modules/custom/mass_utility/mass_utility.module
+++ b/docroot/modules/custom/mass_utility/mass_utility.module
@@ -146,10 +146,18 @@ function mass_utility_form_node_form_alter(&$form, FormStateInterface $form_stat
 
   // Add link for contacting a user in the publish area.
   if (isset($form['meta']['author']['#markup'])) {
-    $title = $node->getTitle();
-    $author = $node->getOwner();
+    $title_prefix = 'Outreach about Mass.gov page "';
     $nid = $node->id();
-    $title_nid = "Outreach about Mass.gov page \"$title\" ($nid)";
+    $title_suffix = '" (' . $nid . ')';
+    $prefix_suffix_combo_lenth = strlen($title_prefix . $title_suffix);
+    // Fits title into 100 characters or less.
+    $title_max_length = 100 - $prefix_suffix_combo_lenth;
+    $title = $node->getTitle();
+    // Truncates title as needed when forming combined `$title_nid` string.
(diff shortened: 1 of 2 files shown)


docroot/modules/custom/mass_utility/mass_utility.module
```diff
@@ -146,10 +146,18 @@ function mass_utility_form_node_form_alter(&$form, FormStateInterface $form_stat
...
+    $nid = $node->id();
+    $title_suffix = '" (' . $nid . ')';
+    $prefix_suffix_combo_lenth = strlen($title_prefix . $title_suffix);
+    // Fits title into 100 characters or less.
+    $title_max_length = 100 - $prefix_suffix_combo_lenth;
     $title = $node->getTitle();
+    // Truncates title as needed when forming combined `$title_nid` string.
+    $title_nid = $title_prefix .
+      ((strlen($title) > $title_max_length) ? substr($title, 0, $title_max_length - 1) . '…' : $title) .
+      $title_suffix;
```
a-username, line 159: I'm wondering if there isn't a simpler way to do this. Maybe something like:
```
$title = $node->getTitle();
$nid = $node->id();
$title_nid = "Outreach about Mass.gov page \"$title\" ($nid)";

if (strlen($title_nid) > 100) {
   // we want a negative value here so substr works from the end
   $truncate_val = 100 - strlen($title_nid);
   $title_nid = 'Outreach about Mass.gov page "'. substr($title, 0, $truncate_val) . '..." ' . "($nid)";
}
```
where we just keep the definition of the first three variables and add the `if` statement. It seems a little less confusing to me over breaking the whole message subject apart to get a truncation length from each component subtracted from 100 vs. counting the overall length of the subject and subtracting that from 100 to get our value.

weitzman: I actually think the PR is is about as simple as your proposal. Merging.

PR #281, opened 2020-04-21
DP-17511: deleted referenced entity issue fix

arthurbaghdas:
Description:
The DCAMM org page began producing a 500 error on 2/12/2020. the page hasn't been edited since 1/16/20.

Jira:
https://jira.mass.gov/browse/DP-17511

To Test:
Org pages do not error when an item they reference has been removed.

### DIFF
--- a/docroot/modules/custom/mayflower/src/Prepare/Organisms.php
+++ b/docroot/modules/custom/mayflower/src/Prepare/Organisms.php
@@ -3,6 +3,7 @@
 namespace Drupal\mayflower\Prepare;
 
 use Drupal\Core\Datetime\DrupalDateTime;
+use Drupal\Core\Url;
 use Drupal\mayflower\Helper;
 use Drupal\Component\Utility\UrlHelper;
 use Drupal\Core\Entity\ContentEntityInterface;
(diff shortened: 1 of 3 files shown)


docroot/modules/custom/mayflower/src/Prepare/Organisms.php
```diff
@@ -552,14 +554,20 @@ public static function preparePressListing($entity, $field, array $options = [],
           }
         }
       }
+      // DP-17511: Solving issue when referenced entity is deleted.
+      elseif (isset($field_value->getValue()['target_id']) && !$field_value->getValue()['_loaded']) {
+        break;
```
danstafford: Hey @arthurbaghdas I have a quick question on this...
Do we really want to `break` out of looping through the `$field_values` array if we come across a deleted referenced entity? Maybe we should `continue`  the loop instead and finish looping through the remaining `$field_values` ?

arthurbaghdas: Sounds reasonable @danstafford . I will review it

danstafford: Looks good!

PR #286, opened 2020-04-23
DP-18429: adding field_template field for author and editor

arthurbaghdas:
**Description:**
Permission set to the the user roles editor and author. The functionality is still there it is controlled with permission 'set service page template'

https://jira.mass.gov/browse/DP-18429

**To Test:**
- [ ] Login as a user with role author and add/edit service node
- [ ] Login as a user with role editor and add/edit service node

### DIFF
--- a/changelogs/DP-18429.yml
+++ b/changelogs/DP-18429.yml
@@ -1,0 +1,41 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
+# #############################################################################
+# You can use one of the following types:
+#  - Added: For new features.
+#  - Changed: For changes to existing functionality.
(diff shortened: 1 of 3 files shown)


changelogs/DP-18429.yml
```diff
@@ -0,0 +1,41 @@
...
+# Changed:
+#  - description: Automating the release branch.
+#    issue: DP-10166
+#  - description: Second change item that needs a description.
+#    issue: DP-19875
+#  - description: Third change item that needs a description along with an issue.
+#    issue: DP-19843
+#
+Changed:
+  - description: Changing field_template permission to be visible for Author and Editor!
```
weitzman:
```suggestion
  - description: Changing field_template permission to be visible for Author and Editor.
```

PR #297, opened 2020-05-01
DP-18313 Allow alerts by organization

danstafford:
**Description:**

1. I added a "By organization” option to the Alert type field on the Alert node.
2. When "By organization” is selected a target organizations field appears where multiple organizations can be entered.
3. An error message appears prompting the user to enter an organization if "By organization” is selected and an organization is not entered.
4. The Target Organizations field visually appears only when "By organization” is selected.
5. The Target Organizations field is an entity reference field to a `target_organizations` paragraph type. This is architecture is identical to how we have the Target Pages paragraph setup. This allowed more re-use of code and hopefully better maintainability. 
6. The Target Organizations paragraph has a single entity reference field which can only reference organization nodes. 
7. I added necessary field and paragraph field queries to the alert JSON response for organizations. 
8. The datalayer on each node on the site contains an array of organization node ids that the node belongs to. I initially tried to use these organization node ids as a way to determine if the node has an organization alert that needs to be shown. But I found this required significant additions to the alert JSON repsonse and additional looping in the JS to make this comparison. I ended up adding the organization UUID to the datalayer of each node as well. This reduced the amount of JSON in the response and JS so it should be better for performance. This did end up altering the structure of the `entityField_organizations` array in the datalayer. I don't think this will have unintended consequences on other parts of the site but I am not 100% sure.
9. Many organization (`org_page`) nodes reference themselves with an organizations field but in some cases they do not. But we would still want an org alert to appear on org pages even if the self-reference was forgotten. I added some JS to account for this. 
10. I added a test for the correct message showing when an organization is not selected. 
11. Also I tried to add a functional test for the alert appearing on a node but ran into some problems which I noted in the file. I included this `OrganizationAlertsClientSideTest.php` file in the PR (but skipping the tests) for now and maybe we can get it working later. 

https://jira.mass.gov/browse/DP-18313

**To Test:**

- [ ] This might be worth testing on a Acquia feature environment as locally the alerts are pretty slow to appear
- [ ] Create or edit an alert node and select "By organization” the the Alert type
- [ ] Verify the Target Organizations field appears only when "By organization” is selected
- [ ] Verify you cannot save the node when "By organization” is selected and Target Organizations is not filled out
- [ ] Add several orgs to this Target Organizations field
- [ ] Check the help text on this form, note any adjustments needed
- [ ] Edit or create a node that belongs to one of the orgs you just referred to
- [ ] Verify the alert appears in the correct place with the correct styling
- [ ] Add a few target page alerts to this same node so you can see multiple page and org level alerts appear on the page 
- [ ] Add multiple Target Organizations to an alert then edit a node and make it a member of both orgs. Now verify the alert only appears once. 
- [ ] Verify organization alerts appear on organization (`org_page`) nodes even if these pages do not reference themselves in their own organizations field.
- [ ] Check some random pages and verify the org alert isn't popping up where it shouldn’t be
- [ ] Do some checks with the site-wide alerts and verify they work exactly as before
- [ ] Also check out the testing steps provided in the Jira ticket description 

https://jira.mass.gov/browse/DP-18313

**Screenshots/GIFs:**

---


conf/drupal/config/block.block.ajaxpattern_2.yml
```diff
@@ -18,7 +18,7 @@ settings:
...
   provider: mayflower
   label_display: '0'
-  ajax_pattern_endpoint: '/jsonapi/node/alert?page[limit]=250&sort=-changed&include=field_target_pages_para_ref,field_alert&filter[status][value]=1&fields[node--alert]=title,changed,entity_url,field_alert_severity,field_alert,field_target_pages_para_ref,field_alert_display&fields[paragraph--emergency_alert]=drupal_internal__id,changed,field_emergency_alert_timestamp,field_emergency_alert_message,field_emergency_alert_link,field_emergency_alert_content&fields[paragraph--target_pages]=field_target_content_ref'
+  ajax_pattern_endpoint: '/jsonapi/node/alert?page[limit]=250&sort=-changed&include=field_target_pages_para_ref,field_target_orgs_para_ref,field_alert&filter[status][value]=1&fields[node--alert]=title,changed,entity_url,field_alert_severity,field_alert,field_target_pages_para_ref,field_target_orgs_para_ref,field_alert_display&fields[paragraph--emergency_alert]=drupal_internal__id,changed,field_emergency_alert_timestamp,field_emergency_alert_message,field_emergency_alert_link,field_emergency_alert_content&fields[paragraph--target_pages]=field_target_content_ref&fields[paragraph--target_organizations]=field_target_content_ref'
```
weitzman, line 21: If I'm not mistaken, there are 2 'ajaxpattern' blocks related to alerts. They end point urls in both have to match or else we will get more cloudflare misses. This mismatch recently bit us (dont recall the details).

weitzman (changes requested):
> The Target Organizations field is an entity reference field to a target_organizations paragraph type. This is architecture is identical to how we have the Target Pages paragraph setup. This allowed more re-use of code and hopefully better maintainability.

I value the consistency here. Alas, we forgot to tell you that we are no longer adding paragraphs to the site. The consistency we could use here is to rip out the existing use of paragraphs. Beyond maintainability, this will help a lot with performance. We may be teetering on the edge of a stampede if this response gets any slower. I'm OK with merging this now if we open a follow-up ticket and slate that ticket to rip paragraphs out of alert content type. Or we do that here - I have no pref.

The ExistingSiteJavascript tests are not working I believe. We have an open ticket to fix that. Thanks for contributing one all the same.

danstafford:
Closing this PR out and starting a new branch that includes removing the extraneous paragraphs

PR #304, opened 2020-05-05
Force media download links on www to have 60 second TTL at Cloudflare…

weitzman:
Old value was `RESPECT_ORIGIN`. Now www matches edit.

https://jira.mass.gov/browse/DP-18608

**To Test:**
- [ ] ?

---

### DIFF
--- a/cloudflare/src/backends.js
+++ b/cloudflare/src/backends.js
@@ -112,7 +112,7 @@ export function www(token) {
       browserTTL = 60
     }
     else if(isMediaDownloadUrl(url)) {
-      edgeTTL = RESPECT_ORIGIN
+      edgeTTL = 60
       browserTTL = 60
     }
     else if(isStaticUrl(url)) {


rbayliss:
In theory, this should actually be worse for performance, as it'll take what would normally be a 15 minute backend cache lifetime and shorten it to 1 minute.  That being said, we don't actually seem to be getting 15 minutes out of it, so I'm game to give it a shot and see if it helps.

rbayliss:
We should be able to tell if it's working pretty quickly, just by watching the Cloudflare logs in Kibana. If we see origin requests for those media download URLs drop, we make it permanent.  If not, we should roll it back and come up with a new game plan. I'd be ok with temporarily applying the worker changes to prod directly from this branch in order to test it quickly (obviously, after testing in dev/test).

weitzman:
Closed in favor #308

PR #307, opened 2020-05-06
DP-18625: Adding padding to the bottom of decision tree callout

arthurbaghdas:
**Description:**
Added new css file for overrides

https://jira.mass.gov/browse/DP-18625

### DIFF
--- a/docroot/themes/custom/mass_theme/overrides/css/decision-tree-callout.css
+++ b/docroot/themes/custom/mass_theme/overrides/css/decision-tree-callout.css
@@ -1,3 +1,3 @@
-.ma__decision-tree-post {
-  padding-bottom: 60px;
+.ma__decision-tree-post .ma__callout-alert {
+  padding-bottom: 90px;
 }
--- a/docroot/themes/custom/mass_theme/overrides/css/decision-tree-callout.css
+++ b/docroot/themes/custom/mass_theme/overrides/css/decision-tree-callout.css
@@ -1,0 +1,3 @@
+.ma__decision-tree-post {
+  padding-bottom: 60px;
+}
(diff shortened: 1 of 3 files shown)


docroot/themes/custom/mass_theme/overrides/css/decision-tree-callout.css
```diff
@@ -0,0 +1,3 @@
+.ma__decision-tree-post .ma__callout-alert {
+  padding-bottom: 90px;
+}
```
ygannett, line 3: I see the ticket says "Add padding below decision tree disclaimer", but we generally use `margin` for this type of spacing.

Looking at other page types, it seems it makes more sense to add the bottom spacing to a parent level of `section` components. So, I suggest to  add bottom margin to `.ma__decision-tree-post` instead of `.ma__callout-alert`.  Then, if some other components are added to`.ma__decision-tree-post` in the future, spacing between multiple components could be managed with `margin-top` just like sections in .main-content.

I have a question about the amount of the spacing.  How did you get the **90px**?  If you're told to put 90px, that's what's supposed to be. If not, I think it's good to put the same amount of space above the feedback module through all page types. 

The decision tree page type's structure is different from other page types.  It's confusing.  The `.ma__callout-alert` seems to be a part of the main content, but it's in the `.post-content`.

The margin setting between `.main-content` and `.post-content` which contains the feedback module is as shown:
```
.marin-content {
   margin-bottom: 45px;
}
@media (min-width: 911px) {
   margin-bottom: 60px;
}
```
To match the amount of spacing, this could be apply to `.ma__decision-tree-post`.

Please let me know for any questions.

ygannett (changes requested):

docroot/themes/custom/mass_theme/overrides/css/decision-tree-callout.css
arthurbaghdas: Hi @ygannett  , changed paddinf to margin, re the 90px , initially i changed it to be 60px, Joe changed to 90px. So please review again and approve

ygannett (approved):

PR #312, opened 2020-05-08
DP-16946: Replacing urls in sitemap with media download urls

arthurbaghdas:
**Description:**
URL object passed to the sitemap generator changed to be /media/[media id]/download

https://jira.mass.gov/browse/DP-16946

**To Test:**
- [ ] Regenerate sitemap from here admin/config/search/simplesitemap
- [ ] And verify URLs changed here /sitemaps/1/sitemap.xml

### DIFF
--- a/docroot/modules/custom/mass_site_map/src/Plugin/simple_sitemap/UrlGenerator/DocumentPagesUrlGenerator.php
+++ b/docroot/modules/custom/mass_site_map/src/Plugin/simple_sitemap/UrlGenerator/DocumentPagesUrlGenerator.php
@@ -40,11 +40,11 @@ class DocumentPagesUrlGenerator extends UrlGeneratorBase {
     foreach ($files as $file) {
       // NOTE: We must pass URL object of the file url in our dataset, so that simple_sitemap module can apply baseurl
       // config settings to it.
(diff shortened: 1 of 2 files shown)


docroot/modules/custom/mass_site_map/src/Plugin/simple_sitemap/UrlGenerator/DocumentPagesUrlGenerator.php
```diff
@@ -40,11 +40,11 @@ public function processDataSet($entity) {
     foreach ($files as $file) {
```
weitzman: Are all $files always media entities? If not, we will be generating invalid links?

arthurbaghdas: We are not generating invalid links, because we are generating URLs not from the $file but from the host entity which is always Media entity type. I will add some checks to avoid issues in the future

arthurbaghdas: Added checks. Please review again.

PR #333, opened 2020-05-20
DP-18313 Allow alerts by org and replace alert target paragraph

danstafford:
**Description:**

1. Added two new entity reference fields on the alert node. A target page and a target org field. 
1. I disabled the paragraph reference field that we were previously using for target pages. We will have to remove the old paragraph and reference field on a subsequent release.
2. The new target page and target org fields appear conditionally depending on what alert type is selected. Also error messages appear when the required reference field is not filled out. 
3. I added a post update function that copies the target node references out of the paragraph and into the new field.
4. I adjusted the help text to include to what we had in a previous incarnation of this PR. Now there is help text that includes organization commentary. 
5. Adjusted the JSON response to include the new org field and new target page field. The target org paragraph was removed from the response. I upped the number of alerts in the response from 250 to 300. Also it looks like this JSON response file is going to be reduced in size. By removing the paragraphs the response is simplified and makes for an overall smaller file. In the future the file shouldn’t grow as fast as well because editors will have the option to set alerts by org and not need to reference so many target pages on alert nodes. When I look at the [live site JSON](https://www.mass.gov/jsonapi/node/alert?page[limit]=250&sort=-changed&include=field_target_pages_para_ref,field_alert&filter[status][value]=1&fields[node--alert]=title,changed,entity_url,field_alert_severity,field_alert,field_target_pages_para_ref,field_alert_display&fields[paragraph--emergency_alert]=drupal_internal__id,changed,field_emergency_alert_timestamp,field_emergency_alert_message,field_emergency_alert_link,field_emergency_alert_content&fields[paragraph--target_pages]=field_target_content_ref) it looks like the JSON response file is 1.8mb. When I look at the [new JSON file locally](http://mass.local/jsonapi/node/alert?page[limit]=300&sort=-changed&include=field_alert&filter[status][value]=1&fields[node--alert]=title,changed,entity_url,field_alert_severity,field_alert,field_target_page,field_target_organization,field_alert_display&fields[paragraph--emergency_alert]=drupal_internal__id,changed,field_emergency_alert_timestamp,field_emergency_alert_message,field_emergency_alert_link,field_emergency_alert_content) it looks around 750kb.
6. I adjusted the alerts JS to use the new fields. I adjusted the datalayer to include a node’s organization uuids for this as well.
7. For users without the `create site wide alerts` permission (which is on the `emergency_alert_publisher` role) I disabled the site-wide option in the `field_alert_display` field.

**Follow up work for a future release**
Remove unused paragraph `target_pages`
Remove paragraph reference field on the alert node: `field_target_pages_para_ref`
Remove references to these fields in `mass_alerts.module`

https://jira.mass.gov/browse/DP-18313

**To Test:**

- [ ] Test both the new target page and target org fields on alert nodes. Make sure they toggle visibility when the alert type is selected and show the user an error when they are not filled out and you try to save the node. 
- [ ] Verify the old paragraph reference field no longer displays on the alert node form
- [ ] Compare some target node alerts to the live site. These were migrated with a post update function so checking that they still appear on pages where they should is important
- [ ] Test adding multiple target orgs to an alert
- [ ] Verify org pages that do not reference themselves in their own org field still show org-level alerts
- [ ] Take a look at the help text again
- [ ] Login as a user without the `emergency_alert_publisher` role to verify the site-wide option in the alert type field is not available.
- [ ] Also check out the testing steps provided in the Jira ticket description but bear in mind this PR now includes work done to the target page functionality

**Screenshots/GIFs:**

---


docroot/modules/custom/mayflower/js/ajax-pattern-alerts.js
```diff
@@ -166,64 +173,83 @@
...
         }
+
+        // Loop through relevant alert fields.
+        alertReferenceFields.data.forEach(function (paraItem) {
+          if (exit) {
+            return;
+          }
+          // Only get data from the alert if has a relevant uuid to this current page.
+          if ($.inArray(paraItem.id, uuids) !== -1) {
+            var alertDetailParagraphIds = Drupal.behaviors.mayflower.getAlertParagraphIds(currentAlertItem);
```
weitzman, line 231: Are we still using paragraphs in other parts of the alerts content type? Or maybe these variable names should be changed. I havent looked too closely.

danstafford: Yes we still need to use the emergency alert paragraph for the alert content. This is a place where using a paragraph is necessary.

weitzman (approved):
Approved. Anyone may merge after ER passes

PR #335, opened 2020-05-21
DP-18611 Drupal API errors

tom-fleming:
**Description:**
I set a paging limit of 1,000 to the Documents by Contributor displays, which currently results in memory limit-related 500 errors when a request attempts to return too many results. After making this change, the drush commands returned unexpected output, and some related fixes are also included here.

https://jira.mass.gov/browse/DP-18611

**To Test:**
See Jira case.

**Screenshots/GIFs:**

---


danstafford (approved):
Looks good!

PR #336, opened 2020-05-21
Add a new hold to build_tag workflow for releasing without maint mode.

weitzman:
https://jira.mass.gov/browse/DP-18558

**To Test:**
- [ ] I think the way to test this is to merge it to develop, release using the regular hold, and then do a test hotfix after this code is on Prod. The stakes are low. I think the only thing that could wrong is that we do maint mode when we prefer not to (i.e our current process).

---


weitzman:
Reroll LGTM

PR #337, opened 2020-05-21
DP-17139: Change revision cleanup duration to 14 months

arthurbaghdas:
**Description:**
Changing default timestamp to be 14 months ago instead of 1 year for ma:queue-revision-cleanup

https://jira.mass.gov/browse/DP-17139

### DIFF
--- a/changelogs/DP-17139.yml
+++ b/changelogs/DP-17139.yml
@@ -1,0 +1,41 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
+# #############################################################################
+# You can use one of the following types:
+#  - Added: For new features.
+#  - Changed: For changes to existing functionality.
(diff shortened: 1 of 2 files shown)


changelogs/DP-17139.yml
```diff
@@ -0,0 +1,41 @@
...
+# Changed:
+#  - description: Automating the release branch.
+#    issue: DP-10166
+#  - description: Second change item that needs a description.
+#    issue: DP-19875
+#  - description: Third change item that needs a description along with an issue.
+#    issue: DP-19843
+#
+Changed:
+  - description: Changing default timestamp for the ma:queue-revision-cleanup drush command to be 14 months ago
```
weitzman:
```suggestion
  - description: Change duration for the ma:queue-revision-cleanup drush command to be 14 months ago
```

PR #352, opened 2020-06-02
Relationship indicator SQL speedup

weitzman:
Our second try with this ticket. We added binder nav test so we should not fail in same way.

**Description:**
We no longer inherit from QueryGeneratedEntityReferenceListUpdated as its better to break up this query.

DP-15829

**To Test:**
- [ ] Review related nodes for several different kinds of pages. The listed nodes should match whats shown on Prod. The order might be different, but the content should be the same. If they aren't the same, and there are more than 25 pieces of related content,  we need to decide which 25 to show.

---


docroot/modules/custom/mass_content/src/Field/RelatedNodes.php
```diff
@@ -130,15 +130,6 @@ protected function links() {
       }
     }
 
-    $types = $this->getSetting('types');
```
tom-fleming, line 116: @weitzman Removing because I don't think we need this query, which just loads all nodes by type, and it ultimately triggered the issue of the top binder nav not loading. This also seems redundant when we are limiting the query by ID or URI in the other getQuery() calls. It's likely a slight performance drain as well.

docroot/modules/custom/mass_content/src/Field/RelatedNodes.php
```diff
@@ -52,9 +52,9 @@ class RelatedNodes extends EntityReferenceFieldItemList {
    * {@inheritdoc}
    */
   public function computeValue() {
-    $nids_fields = $this->fields();
     $nids_links = $this->links();
-    $nids_total = array_merge($nids_fields, $nids_links);
+    $nids_fields = $this->fields();
+    $nids_total = array_diff($nids_links, $nids_fields);
```
tom-fleming: @weitzman In the old version of this, we had a condition in place that filtered out nodes that were manually set against this content. It was done in a confusing way where we created an array of NIDs in computeValue() and passed those values to query(). I think changing this to an array_diff() replicates that behavior, but I didn't really notice a difference in the TOC with or without this in place.

weitzman:
This passes Internal Review, and may be merged once it passes External Review. I cant actually Approve in github as I'm the PR author.

weitzman:
I forgot to say that it would be nice to improve the new binder nav test so that it catches the failure that this pr had.

PR #359, opened 2020-06-05
DP-18846 Alert system tests

danstafford:
**Description:**
I added new tests for organization based alerts and specific page alerts that verify these alerts appear on the correct pages. They require Ajax to work and thus extend the `ExistingSiteWebDriverTestBase` class. 

I adjusted the existing site-wide alert tests that were failing to get them working again. There can only be one site-wide alerts published at a time so I begin the test by unpublishing existing site-wide alerts to have a clean plate for testing. 

I added user tests for various user roles being able to access alerts or publish them. 

The added tests include:
- An organization alert correctly appears on a news node
- An organization alert correctly appears on a organization node that does not reference itself in the organizations field
- A specific page alert correctly appears on the three different nodes it targets
- A site-wide alert alert correctly appears on /alerts, the homepage, and a random page
- A specific page alert can be created by a user with the editor role
- A organization alert can be created by a user with the editor role
- A site-wide alert can be created by a user with the emergency_alert_publisher role
- A site-wide alert cannot be created by a user with the editor role

Also I ran into a problem where CircleCI was running out of memory while running the tests. You can see I upped the memory_limit in `settings.php` for `ci` environments only. 

https://jira.mass.gov/browse/DP-18846

**To Test:**
- [ ] If the tests cover the required conditions and pass circleCI then this is good.

**Screenshots/GIFs:**

---

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,7 +62,7 @@ executors:
           DTT_BASE_URL: http://mass.local
           DTT_API_URL: http://mass.local:9222
           BROWSERTEST_OUTPUT_DIRECTORY: /tmp
-#          MEMORY_LIMIT: 2GB
+          MEMORY_LIMIT: 2GB
       - image: << parameters.instance >>
         auth:
           username: $DOCKER_USER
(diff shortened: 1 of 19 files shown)


weitzman:
This looks fantastic.

I dont see SiteWideAlertsClientSideTest anywhere in the browser_output files. I do see the other client side tests. This makes me wonder if its running.

danstafford:
Hey @weitzman , I swapped out `$this->visit()` for `$this->drupalGet()` and it now generates tests in `browser_output`. They were running before as well, just not outputting tests. I think `drupalGet` calls `$this->htmlOutput` where `visit` does not.

PR #366, opened 2020-06-09
DP-18958 Add ma:backstop drush command

weitzman:
**Description:**
A new drush command for running Backstop at CircleCI.

https://jira.mass.gov/browse/DP-18958

**To Test:**
- [ ] Run locally `drush ma:backstop prod test --ci-branch=backstop-drush-command -vvv`. Verify that a build is created at Circle and that Backstop uses correct target and reference. Note that Backstop always fails so this experiment will give a red build.

---

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,6 +98,13 @@ parameters:
   ma-cf-deploy:
     type: boolean
     default: false
+  # These params for currently used in drush ma:backstop. target is declared later.
+  ma-backstop:
+    type: boolean
+    default: false
+  reference:
+    type: string
+    default: ""
 
   # These params for currently used in drush ma:release
   ma-release:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -109,7 +116,7 @@ parameters:
   skip-maint:
     type: string
     default: ""
-  # target also used by ma-cf-deploy
+  # target also used by ma-cf-deploy, ma-backstpo
```
tom-fleming: @weitzman Hope I'm not being too nit-picky, but ma-backstop is spelled wrong here

weitzman:
```suggestion
  # target also used by ma-cf-deploy, ma-backstop
```

weitzman: No problem. Fixed. FYI, the suggestions feature of Github is great for review comments like this. See https://thenextweb.com/dd/2018/10/16/github-launches-suggested-changes-to-make-it-easier-to-collaborate-within-pull-requests/

PR #373, opened 2020-06-16
Mayflower/DP-9097 mayflower version 9.52.0

ygannett:
Changed:
  - description: |-
    Updated Mayflower version to 9.52.0.
         - DP-15628: Make the parent container of TOC link function like a link by expanding clickable area. (MF)
         - DP-17612: Adjust z-index of open search box to fix style issue on Firefox.  (MF)
         - DP-18940: Adjusted the key-message template and styles to stop BG image display problem. (MF)
    issue: DP-19097

### DIFF
--- a/changelogs/DP-19097.yml
+++ b/changelogs/DP-19097.yml
@@ -38,8 +38,8 @@
 #
 Changed:
   - description: |-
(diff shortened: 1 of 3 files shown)


ygannett:
backstop failures are from content chagne.

PR #374, opened 2020-06-16
Enable upgrade_status and supporting modules.

weitzman:
**Description:**
Enables the modules needed to guide our effort to get to D9. These modules should be uninstalled once we don't need them.

https://jira.mass.gov/browse/DP-18249

**To Test:**
- [ ] Browse to Admin => Reports > Upgrade Status and Analyze the custom modules to see warnings and errors. Analyze contrib modules to see what needs updating. Note that analyze takes a long time.

---


danstafford (approved):
Looks great!

PR #378, opened 2020-06-18
DP-19089 Add noindex and follow to doc pages

danstafford:
**Description:**
Add noindex and follow to the robots metatag when "docs" is the second part of the url and "download" is not the last part. 

https://jira.mass.gov/browse/DP-19089

**To Test:**
- [ ] Load a page like this one http://mass.local/doc/reopening-massachusetts
- [ ] Verify the robots metatag appears as described

**Screenshots/GIFs:**

---


docroot/modules/custom/mass_metatag/mass_metatag.module
```diff
@@ -367,8 +367,17 @@ function mass_metatag_page_attachments(array &$attachments) {
   $current_uri = Url::fromRoute('<current>')->toString();
   $uri_parts = explode('/', $current_uri);
   $uri_end = end($uri_parts);
+  $noindex_and_follow = FALSE;
+
+  if ($uri_parts[1] === 'doc' && $uri_end !== 'download') {
```
weitzman: This looks brittle to me. Can we get the route and see if its a media entity (and not a download route). This API-based condition will be more robust that looking and uri parts.

danstafford: Thanks @weitzman , are you thinking something like this?
```
  if (\Drupal::routeMatch()->getRouteName() === 'entity.media.canonical' && $uri_end !== 'download') {
    $noindex_and_follow = TRUE;
  }
```

weitzman: Yes. And is `media_entity_download.download` the /download rooute? See docroot/modules/contrib/media_entity_download/media_entity_download.routing.yml

danstafford: Hmmm... when a user hits the `/media/{media}/download` route they are served a file, not an html page where metatags can be added. So I don't think `mass_metatag_page_attachments` even runs in that case. Maybe we don't need the "download" check altogether?

weitzman: That makes sense to me ... Maybe do a git blame on that line and see if you can find out what issue introduced the /download check.

danstafford: OK I removed the extraneous /download check. Now just media entities get the robots metatag and there are no fragile url part checks for this.

weitzman: Thanks. We should add a test IMO.

danstafford: Thanks @weitzman , I added a test for this when you get a minute.

PR #383, opened 2020-06-22
DP-19148 Fix robots on tasks pages

danstafford:
**Description:**
The `tasks` pages are `view_mode_page` pages unlike the `need-to-know`, `resources`, and `related` pages which are `views` pages and also need this specific robots metatag. So on `tasks` pages the URL returned without the `tasks` part and so the tag wasn't being added. I modified this to use the path alias which should do the trick for the use cases here. I added unit tests for this as well.

https://jira.mass.gov/browse/DP-19148

**To Test:**
- [ ] These pages should have a robots metatag with "noindex, follow"
http://mass.local/unemployment-insurance-ui-for-employers/tasks
http://mass.local/unemployment-insurance-ui-for-employers/related
http://mass.local/unemployment-insurance-ui-for-employers/resources
http://mass.local/unemployment-insurance-ui-for-employers/need-to-know
- [ ] Check some other pages that should not have this tag

**Screenshots/GIFs:**

---


docroot/modules/custom/mass_metatag/mass_metatag.module
```diff
@@ -364,16 +364,17 @@ function mass_metatag_page_attachments(array &$attachments) {
     'resources',
   ];
 
-  $current_uri = Url::fromRoute('<current>')->toString();
```
weitzman, line 367: I would think that this returns an aliased URL. Thats not what you are seeing?

danstafford: This is the issue I am trying to resolve. If I am on `http://mass.local/unemployment-insurance-ui-for-employers/tasks` then `Url::fromRoute('<current>')->toString();` returns just `/unemployment-insurance-ui-for-employers`. I was looking for a way to determine if this is a `view_mode_page` and get the "tasks" part but I could't figure out how to that.

weitzman: Oh. In that case 'tasks' is the is called an "argument" or a "param". You can get those from some service. The routematch service I think. I'll look around if you tried that already.

danstafford: Yeah I tried `\Drupal::routeMatch()->getParameter('node')` and got the node entity but I can't find the currently used view_mode in there. The view_mode is `all_tasks` which I assume is exclusively used for the `view_mode_page`

weitzman: Is it in -`>getParamaters()`?

danstafford: I tried that too. The only param is the node.

danstafford: Looking at this again I don't think I need to get the alias. I can probably just use `$current_path = \Drupal::service('path.current')->getPath();` . Even though it returns `/unemployment-insurance-ui-for-employers/tasks` for http://mass.local/unemployment-insurance-ui-for-employers/tasks and `/node/7426/related` for http://mass.local/unemployment-insurance-ui-for-employers/related
(/tasks is a view_mode_page and /related is a view-page)

danstafford: OK @weitzman I removed getting the alias like I mentioned in the previous comment.

weitzman (approved):

weitzman:
Test failure is unrelated

PR #386, opened 2020-06-22
DP-19159 update the Mayflower version to 9.52.1

ygannett:
Changed:
  - description:  |-
      Updated Mayflower version to 9.52.1.
          - DP-17150: Change the label of the secondary set. Change the width of each set. Adjust spacing. (MF)
          - DP-17404: Match spacing with the current prod(develop) version. (MF)
          - DP-17404: Align relationship the first terms in primary and secondary sets. (MF)
          - DP-17404: Position the TOC below the relationship indicator. (MF)
          - DP-17404: Put back missing mobileNav.js in index.js. (MF)
          - DP-19085: Modify google-map.twig to print googleMap.link.info value and set it visualy hidden as context info for screen reader users. (MF)
    issue: DP-19159

### DIFF
--- a/changelogs/DP-19159.yml
+++ b/changelogs/DP-19159.yml
@@ -1,0 +1,48 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
+# #############################################################################
+# You can use one of the following types:
+#  - Added: For new features.
+#  - Changed: For changes to existing functionality.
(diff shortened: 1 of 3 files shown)


ygannett:
Backstop errors are from the design change of the relationship indicator and the date rendering timing difference of the alert.

Hold off the merge due to the test failure (https://circleci.com/gh/massgov/openmass/14569).

PR #398, opened 2020-06-30
DP-19001 Invalidate node cache tags on form submit

danstafford:
**Description:**
Maybe invalidating the cache tags as soon as the node form is submitted will help.

https://jira.mass.gov/browse/DP-19001

**To Test:**
- [ ] Please see the ticket for this.

**Screenshots/GIFs:**

---


weitzman:
Please add a comment in the code explaining why this is needed. You can just link to the Jira ticket.

Drupal collects cache invalidation requests until after a page is fully rendered and then after render, it merges the tags clears them. So I suspect this code is not actually doing anything. Later code requests invalidation of the the same tag.

Regardless, we can merge it since you've observed that it fixes the bug.

PR #409, opened 2020-07-15
Add PHP Yaml parse validation for changelog.yml files in PRs.

weitzman:
See title. Also, this seems like a harmless PR to bundle in another small change ...

This PR ups our resource_class at CircleCI. Our tests are taking a very long time, and its hurting our team velocity. The default resource_class is no longer cutting it. This PR ups our class to xlarge. We have plenty of monthly credits for this change. We are halfway through the month and have used 1/10 of our credits. That makes sense since our team is much smaller now. Furthermore, our tests are running [3x faster with this PR](https://app.circleci.com/pipelines/github/massgov/openmass/3789/workflows/459e9642-c4eb-4acd-a748-54538f0c20c1/jobs/15912) so I'm not sure this will use more credits since we consume fewer minutes:

image

https://jira.mass.gov/browse/DP-18527

**To Test:**
- [ ] Make a PR with an invalid changelog file and see that it fails.

---

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -222,6 +222,8 @@ jobs:
       - run: composer validate
       # Check coding standards with php code sniffer. See phpcs.xml
       - run: phpcs -v -n
+      # Make sure changelogs are parseable by PHP (since build-changelog.php requires that).
+      - run: php scripts/directory-yaml-parse.php
       # Lint javascript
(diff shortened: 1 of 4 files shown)


tom-fleming:
LGTM! My test failed with an exception as expected, and tests completed in 15 minutes!

PR #422, opened 2020-07-28
Added back dom-crawler - DP-19497

tom-fleming:
**Description:**
Adds back dom-crawler to fix pages caught in nightcrawler

https://jira.mass.gov/browse/DP-19497

**To Test:**
- [ ] Check forms/covid-19-donation-program
- [ ] Run nightcrawler against an environment hosting these code change (e.g. feature4)

**Screenshots/GIFs:**

---


composer.json
```diff
@@ -210,6 +210,7 @@
         "mikeyp/google_tag": "dev-8.x-1.x#3caa4f0b7f008576936ab7501d483bfcf58a4764",
         "phlak/semver": "^2.0",
         "phpunit/phpunit": "^6",
+        "symfony/dom-crawler": "3.4.35",
```
weitzman: Seems like we could be more loose here - `^3` would be fine, no?

weitzman (commented):
Do you think its worth adding a test for a form? It could be as simple as creating a form page (or using a QAG page if one exists) and asserting that we got an OK response. If not needed thats fine.

composer.json
tom-fleming: @weitzman Yeah, I don't think a test would have caught this particular issue, since the library existed on CircleCI, and is needed for the tests to run. I made the dom-crawler requirement looser. Should be fine. I was originally just matching what core-recommended's components use, but it's in step with the latest security version.

PR #443, opened 2020-08-13
DP-19677: Investigate continued issue with slow builds in CircleCI

weitzman:
**Description:**

- Parallelize the Behat and PHPUnit tests. Advantages:
  1. Build is briefer. Just 11 min!
  2. DB/config changes made during phpunit tests dont impact Behat, and vice versa.
  3. When Circle brings back Insights UI, we will get duration reporting for each job separately. Until then, [there is an API for it](https://circleci.com/blog/announcing-new-insights-endpoints-in-circleci-s-api-v2/).
- Emit a JUnit XML for phpunit tests inaddition to Behat and save them as artifacts.
- Emit information to console about CPU and memory

https://jira.mass.gov/browse/DP-19677

**To Test:**
- [ ] If its green its good.

**Screenshots/GIFs:**

---

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -249,8 +249,10 @@ jobs:
       - run: dockerize -wait tcp://localhost:3306 -timeout 2m
       # Run drush updatedb and cim.
       - run: scripts/ma-refresh-local --skip-db-prep
-      # Get a status
+      # Get an overview
       - run: drush status
+      - run: cat /proc/cpuinfo
+      - run: cat /proc/meminfo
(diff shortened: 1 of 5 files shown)


tom-fleming:
@weitzman LGTM. Great point about DB/config, and in general, definitely in favor of having these two tests run in parallel. One thing I needed to change to make this build green was the branch protection rules so that the GitHub/CircleCI integration is looking for the test-behat and test-phpunit jobs rather than test. After I changed that, everything was good.

PR #450, opened 2020-08-26
DP-19391 iframe text wrapping

tom-fleming:
**Description:**
This PR does the following:
- Creates new fields on iframe paragraphs for display, alignment, wrapping, and captions to match the fields and field ordering on Tableau paragraphs
- Migrates existing display and caption fields to the newly created fields
- Removes old field values for display and captions from all displays
- Updates JS and Twig templates to include newly created fields and to copy Tableau field behavior

https://jira.mass.gov/browse/DP-19391

**To Test:**
- [ ] When running the site locally, the only set up required is to run `ahoy updatedb`. Running this should both import the new configurations and migrate the field data.
- [ ] Check that existing caption and display fields have been migrated to their newer counterparts (field_media_display to field_iframe_display_size and field_caption to field_iframe_caption)
- [ ] Confirm that old fields and their values do not appear in any displays, such as edit forms and page displays.

**Screenshots/GIFs:**

---


ygannett (approved):
@tom-fleming The change looks good.
I see the validate failure with circle ci.  It looks like a lint error. Once that's cleared, it's ready to merge.

PR #459, opened 2020-09-14
Run tests with fixed entity definitions (i.e. mismatches resolved)

weitzman:
**Description:**
- Brings devel_entity_updated into the codebase
- Runs it before running tests so we verify that the site works

https://jira.mass.gov/browse/DP-19748

**To Test:**
- [ ] None

---


weitzman:
We can remove the config.yml changes before or after merging this code. They are just there to prove green.

arthurbaghdas (approved):
Looks good to me

PR #464, opened 2020-09-16
Fix dashboard iFrame URLs.

adamcogbill:
**Description:**
Fixes the iFrame URLs for our web analytics Superset dashboards. Previously, the URLs included an incorrect filter ID, so the dashboards were not being correctly filtered.

https://jira.mass.gov/browse/DP-19935

**To Test:**
_Parent page dashboards:_
* Open any service, organization, topic, binder, or curated list
* Click the "analytics" tab
* The data in this tab should 1. be there (i.e. there aren't a bunch of blank charts) and 2. be only about _this_ page, not _all_ service, org, topic, binder, or curated lists. You can check Google Analytics to be super sure, but the "Previous URLs" chart should be a good indication. 

_Promo page dashboard_
* Same as above but with promotional pages

_Basic dashboard_
* Same as above but with a law library content type

**Screenshots/GIFs:**

---

### DIFF
--- a/changelogs/DP-19935.yml
+++ b/changelogs/DP-19935.yml
@@ -1,0 +1,41 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
+# #############################################################################
+# You can use one of the following types:
+#  - Added: For new features.
+#  - Changed: For changes to existing functionality.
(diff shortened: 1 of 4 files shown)


adamcogbill:
THIS IS VERY EXCITING, WATCH OUT DRUPAL WORLD

PR #480, opened 2020-10-02
DP 20094 help text change for data resource type field

joeg8612:
**Description:**
Changing help text for data resource type field since we are restricted options to a single one for the time being.

https://jira.mass.gov/browse/DP-20094

**To Test:**
- [ ] Edit an info detail or service details page. Check the data flag. Choose the data type "data resource" Verify help test for the data resource type field matches the Jira ticket text.

**Screenshots/GIFs:**

---

### DIFF
--- a/changelogs/DP-20094.yml
+++ b/changelogs/DP-20094.yml
@@ -1,0 +1,41 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
+# #############################################################################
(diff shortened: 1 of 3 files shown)


weitzman (approved):

weitzman:
Approved the change, but it has to be green before merge. If you push a simple text change or update .build_trigger file, you will get a new build which should pass. The last build happenned before the PR was created and that always fails (deliberately).

PR #491, opened 2020-10-15
Change admin theme in Drupal to MF v10 font Noto

joeg8612:
**Description:**
This is to preview the MF font change in Drupal. This branch will never be merged.

https://jira.mass.gov/browse/DP-20297

**To Test:**
- [ ] Edit a page in Drupal. May sure that the Noto font is used in the admin theme both while previewing a page and while not previewing it.

---


ygannett (commented):
The ticket says changing the fonts for the admin theme, but the change is made in docroot/themes/custom/**mass_theme**/templates/layout/html.html.twig, which is the Mass.gov site theme.  The admin theme is at docroot/themes/custom/**mass_admin_theme**/.  Do you still see the change in the admin area?

ygannett (approved):
Realized 2 things are taken care of here.  One for the Mass.gov theme and another for the admin theme.  It looks good.

PR #508, opened 2020-10-28
CDN token verification

weitzman:
DP-20164

**To Test:**
- [ ] Browse to https://edit.stage.mass.gov and verify that it works fine since you are routed through Cloudflare
- [ ] Map our Origin IP (aka EIP) to edit.stage.mass.gov via your local `/etc/hosts` file. Get the origin IP from bottom of this [Acquia UI page](https://cloud.acquia.com/a/develop/applications/ff8ed1de-b8bc-48a4-b316-cd91bfa192c4/environments/26646-ff8ed1de-b8bc-48a4-b316-cd91bfa192c4). 
- [ ] In a new browser go to https://edit.stage.mass.gov/. Accept the SSL warning and verify that you get a 403 whose text reads "Only requests sent through Cloudflare CDN are allowed.".  This happens because you are bypassing Cloudflare.

---

### DIFF
--- a/docroot/sites/default/settings.acquia.php
+++ b/docroot/sites/default/settings.acquia.php
@@ -3,8 +3,8 @@
 use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
 
 $is_prod = isset($_ENV['AH_NON_PRODUCTION']) && $_ENV['AH_NON_PRODUCTION'];
-$cli = (php_sapi_name() == 'cli');
-$is_mass_gov = !$cli && preg_match("/\.mass\.gov$/", $_SERVER["HTTP_HOST"]);
+$cli = php_sapi_name() == 'cli';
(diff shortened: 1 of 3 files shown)


docroot/sites/default/settings.acquia.php
```diff
@@ -2,6 +2,10 @@
 
 use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
 
+$is_prod = isset($_ENV['AH_NON_PRODUCTION']) && $_ENV['AH_NON_PRODUCTION'];
```
tom-fleming: @weitzman What we're checking here is that it's nonprod, right? I'm getting a little hung up on the variable name

tom-fleming:
LGTM

tom-fleming (approved):

PR #510, opened 2020-10-28
DP-19553: Improve authoring experience when adding existing media as download

arthurbaghdas:
**Description:**
Display none for the buttons.

DP-19553


tom-fleming:
@arthurbaghdas Did you explore doing this instead in something like a preprocess function? Specifically, I wonder if we could target every inline editing field that has the "Add media item" button and hide this button at that level, similar to what was done here: https://github.com/massgov/openmass/blob/15cd66ad83c5cc149b1bfa12b5ead38ead6ac525/docroot/modules/custom/mass_docs/mass_docs.module#L52-L58

arthurbaghdas:
> @arthurbaghdas Did you explore doing this instead in something like a preprocess function? Specifically, I wonder if we could target every inline editing field that has the "Add media item" button and hide this button at that level, similar to what was done here:
> 
> https://github.com/massgov/openmass/blob/15cd66ad83c5cc149b1bfa12b5ead38ead6ac525/docroot/modules/custom/mass_docs/mass_docs.module#L52-L58

I've refactored code with suggested approach with some changes and it works. Please test again. @tom-fleming

tom-fleming (approved):
LGTM!

PR #513, opened 2020-11-02
DP-19696: Security updates for OpenMass

arthurbaghdas:
**Description:**
Updated packages inside cloudflare folder.

https://jira.mass.gov/browse/DP-19696

### DIFF
--- a/changelogs/DP-19696.yml
+++ b/changelogs/DP-19696.yml
@@ -1,0 +1,41 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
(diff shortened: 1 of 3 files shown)


weitzman:
Has this been deployed to editcf environment? That would be the next step. You can do so by running `drush ma:cf-deploy cf --ci-branch=DP-19696-security-updates-dependabot` and completing the workflow that gets created. Then make web requests to `https://editcf.digital.mass.gov`. These requests run through the new worker proposed by this PR.

arthurbaghdas:
Done @weitzman , smoke tested the environment as well. Thanks

PR #532, opened 2020-11-16
DP-20113: Convert iframe whitelist to not be in code

arthurbaghdas:
**Description:**
Added config_ignore module and configured to ignore mass_utility.settings.yml file from configs

https://jira.mass.gov/browse/DP-20113


weitzman:
I think we want dev-2.x of config_ignore and config_filter. These have had important commits since last release.

In general I'm fine with this approach. I'd like to highlight some considerations:

1. Making config changes directly in Production is inherently destabilizing. The changes have not been tested in the rigorous way that we do now (full CI test suite, manual QA, formal release notes that all can view, etc.)
1. Config changes in Prod can easily cause broad cache clearing which can lead to performance problems. Do we allow these during the day? Do we provide notice in Chat beforehand (a la Google Analytics changes)? 
1. Whom do we allow to make config changes directly in Prod. People who have this permission need to know the ramifications of the changes they are making, both on site quality and site performance.

weitzman (approved):

weitzman (changes requested):
Lets remember to update the versions of filter and ignore packages.

arthurbaghdas:
Modules updated.

- config_ignore to dev version
- config_filter to 2.x version

I was not able to find place where the code can cause mass cache clearing. So I think we are good to deploy.
Each form has a separate permission created which can be granted to any role.

weitzman:
> I was not able to find place where the code can cause mass cache clearing. So I think we are good to deploy.

The scenario I am thinking of is to look at what cache tags get cleared when you submit the admin form. Then review a service page, a topic page, etc. and see if they have those tags associated. Thats how I would determine whether this PR has mass cache clearing potential.

arthurbaghdas:
Worth checking. Agree.

arthurbaghdas:
On form submission cache invalidated by context "user.roles:authenticated", which doesn't affect public facing part that contains "user.roles:anonymous" context. So i think we are good to deploy unless any other concerns.

composer.json
```diff
@@ -109,6 +109,8 @@
         "drupal/components": "^1",
         "drupal/composer_deploy": "^1.4",
         "drupal/conditional_fields": "1.x-dev",
+        "drupal/config_filter": "^2.2",
+        "drupal/config_ignore": "2.x-dev@dev",
```
weitzman: What does the `@dev` suffix do? I dont see it on any other packages in this file, and we use a lot of dev packages.

weitzman (commented):
Thanks for looking into cache clearing.

composer.json
arthurbaghdas: This is how it was written on drupal.org page. Not sure to be honest.

weitzman: Lets omit it please.

arthurbaghdas: Done

arthurbaghdas: Done

weitzman (approved):

PR #539, opened 2020-11-23
DP-20556: Alerts: author choice made with "message link" field not always shown on published page

arthurbaghdas:
**Description:**
Added Link type field to the endpoint request and the logic related to it.

https://jira.mass.gov/browse/DP-20556


tom-fleming:
Behat test is still failing but is addressed in another issue. Merging.

PR #551, opened 2020-12-02
DP-20418: Media bundle is missing file for media entity: XYZ

arthurbaghdas:
**Description:**
Changed error message from both throwing and reporting to watchdog to just reporting the error to watchdog.

https://jira.mass.gov/browse/DP-20418

**To Test:**
- [ ] Check that the error exists by loading a media entity page that does not have a file (I used media/1518421) and check that the error shows up in Portainer. Also verify that the media entity page itself still loads without error.
- [ ] Add a file to the media entity and save. Verify that you no longer see an error in Portainer, that the media entity page itself still loads without error, and that you can also see the "Last Updated" info about the file along with other expected fields, such as the link to download the file.
---

### DIFF
--- a/changelogs/DP-20418.yml
+++ b/changelogs/DP-20418.yml
@@ -1,0 +1,41 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
+# #############################################################################
+# You can use one of the following types:
+#  - Added: For new features.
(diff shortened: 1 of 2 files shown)


weitzman:
This removes the exception from web requests as well. That begs the question of is the exception useful? Maybe we should watchdog a message.

ygannett (approved):

PR #552, opened 2020-12-04
Dp 20498: Add service family key indicator fields and dashboard to PFML service pages

roberttstephens:
**Description:**
This is an MVP attempt to add key indicators to service families.

For this first round, the key indicators are only being added to two PFML "service families".

Since there is a one to one relationship between service families and service pages, the fields are being added to the two service pages.

 - /paid-family-and-medical-leave-benefits/edit
 - /paid-family-and-medical-leave-information-for-massachusetts-employers/edit

The key indicator related fields will be under a "Service Family Key Indicators" sub tab/group on the node edit page similar to the KPI related fields on promotional pages like /try-fishing/edit . The fields will not show up on any other service pages or while adding/creating a new service page.

A service family dashboard is also added as a sub tab under analytics -> service family. This is ONLY on the two PFML service pages. All other content types (like promotional pages) and service pages will have no change on the analytics tab. The dashboard has already been created and has a URL of https://dashboards.digital.mass.gov/superset/dashboard/service-family/?preselect_filters=%7B%221823%22%3A%7B%22node_id%22%3A%5B%22535101%22%5D%7D%2C%221542%22%3A%7B%22__time_range%22%3A%22Last%20quarter%22%7D%7D (the node ID varies).

This dashboard cannot be tested on the feature environments since none of the superset dashboards display.

 - /paid-family-and-medical-leave-benefits/analytics/service-family
 - /paid-family-and-medical-leave-information-for-massachusetts-employers/analytics/service-family

The service family key indicators are added to the API at /api/v1/content-metadata automatically. This work was done in https://jira.mass.gov/browse/DP-16429 . Any fields matching a pattern of field_kpi_ are automatically included and the existing API code will automatically divide the value by 100 if the field name contains _pct or _ctr.

The ETL code base will grab the KPI targets from the API at /api/v1/content-metadata the evening after the fields are populated in Drupal. That work is NOT part of this PR/ticket.

https://jira.mass.gov/browse/DP-20498

**To Test:**
This is currently up on http://massgovfeature4.prod.acquia-sites.com/ .
 - Create a new service page at /node/add/service_page. There should be no group for key indicator fields.
 - Edit an existing non-pfml service page such as /protect-yourself-or-your-loved-ones-from-problem-gambling/edit . There should be no group for key indicator fields.
 - Go to the analytics/dashboards section of an existing non-pfml service page such as /protect-yourself-or-your-loved-ones-from-problem-gambling/analytics/dashboards . There should be no "service family" tab.
 - Go to the analytics/dashboards section of a pfml service page. There should be a "service family" tab that is available. You can view either /paid-family-and-medical-leave-benefits/analytics/service-family or /paid-family-and-medical-leave-information-for-massachusetts-employers/analytics/dashboards .
 - Edit one of the two services pages. There should be a group for key indicator fields. You can edit either node/535101/edit or node/414121/edit .
 - Click on "SERVICE FAMILY KEY INDICATORS" and select more than 3 key indicators. You should see several validation errors.

-   - One for each field that was selected with a message of "A value must be added for this field".
-   - An additional error that there cannot be more than 3 key indicators.

 - Select up to three key indicators, add values, and click save.
 - Check the KPIs in the API: When adding key indicators to node/414121/edit I can see those values at 
 http://mass.local/api/v1/content-metadata?content_types=service_page&published=1&depth=1&offset=2000 if I ctrl+f for 414121 and find the kpis object. 

**Screenshots/GIFs:**

---


docroot/modules/custom/mass_admin_pages/mass_admin_pages.module
```diff
@@ -232,3 +232,36 @@ function mass_admin_pages_preprocess_page(&$variables) {
...
+  }
+}
+
+/**
+ * Implements hook_form_FORM_ID_alter().
+ *
+ * Hide the KPI tab and fields on all service pages except two PFML pages.
+ */
+function mass_admin_pages_form_node_service_page_edit_form_alter(&$form, &$form_state, $form_id) {
+  $pfml_service_family_nids = [414121, 535101];
```
tom-fleming, line 256: @roberttstephens Could we instead store these values via the State API?

roberttstephens: @tom-fleming I'm happy to make the change if you think it's best. What would be the benefit of using state instead of hard coding the nids for now?

These nids are also stored in https://github.com/massgov/openmass/pull/552/files#diff-4747375ceb5366a98cd74f71459192b08c0f3abfb7f0b97bd82126c3631d0629 , the config that determines if the service family analytics dashboard shows. It seems like if the nids did somehow need to change, other code changes would be necessary even if these nids are pulled from the state API.

tom-fleming: Oh, wow, I see what you mean. It looks like that contributed module was written for this project, too. I wish that it wasn't storing those values in config either 🙃, but I don't really see the point in rewriting the module for this task. So, I'm okay with this.

tom-fleming (approved):

PR #553, opened 2020-12-07
DP-20281 Translation support

ygannett:
**Description:**

https://jira.mass.gov/browse/DP-20281

**To Test:**

**Screenshots/GIFs:**

---


tom-fleming:
@weitzman One thing I'm realizing on this ticket is that it could use at least one test for the Translations tab. Please let me know if you think any tests might be needed elsewhere as well and I will add that in.

docroot/modules/custom/mass_translations/mass_translations.deploy.php
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Migrate document language fields.
+ */
+function mass_translations_deploy_language_terms(&$sandbox) {
+  $_ENV['MASS_FLAGGING_BYPASS'] = TRUE;
+
+  $english_target_id = 8876;
+
+  $query = \Drupal::entityQuery('media')
+    ->condition('status', 1)
```
weitzman: Unpublished media does not need migration?

docroot/modules/custom/mass_translations/mass_translations.routing.yml
```diff
@@ -0,0 +1,23 @@
...
+  requirements:
+    _permission: 'view Translations tab'
+    _custom_access: '\Drupal\mass_translations\Controller\NodeTranslationsController::access'
+mass_translations.controller_media_translations:
+  path: '/media/{media}/translations'
+  defaults:
+    _controller: '\Drupal\mass_translations\Controller\MediaTranslationsController::content'
+    _title: 'Translations'
+  options:
+    _admin_route: TRUE
```
weitzman, line 16: Any particular reason why this is an admin route but the one above is not?

weitzman (commented):
Looks nice and clean. IMO, this merits a test or 2 for the key use cases. Ideally we would use the forms to create content and translate it and assert that all looks good along the way.

docroot/modules/custom/mass_translations/mass_translations.routing.yml
tom-fleming: The one above is using _node_operation_route. I had to jog my memory, but _node_operation_route sets _admin_route

docroot/modules/custom/mass_translations/mass_translations.module
```diff
@@ -0,0 +1,123 @@
+<?php
+
+/**
+ * @file
+ * Contains mass_tours.module.
```
tom-fleming:
```suggestion
 * Contains mass_translations.module.
```

docroot/modules/custom/mass_translations/mass_translations.deploy.php
tom-fleming: It does. I added this as a migration. I should note too that I ended up adding a bypass environment variable for a presave function in mass_media (it skips moving files, which was failing the migration when files didn't exist...likely helped a little with performance as well)

tom-fleming: > Looks nice and clean. IMO, this merits a test or 2 for the key use cases. Ideally we would use the forms to create content and translate it and assert that all looks good along the way.

I also added tests for checking hreflang values on nodes (though I did this for only one content type), checking the Language value on Media pages, and checking that the tabs load on both.

docroot/modules/custom/mass_translations/mass_translations.module
```diff
@@ -0,0 +1,123 @@
...
+    }
+    // Since entity was found, no need to iterate further.
+    break;
+  }
+  // Apply updated caching information.
+  $cache->applyTo($attachments);
+}
+
+/**
+ * Implements hook_field_widget_multivalue_form_alter().
```
weitzman, line 71: Could we say what we are doing here and why? I'm not sure now, and this will even less clear in the future.

weitzman (approved):
Great work. Approved with a minor comment. Feel free to merge with or without it.

docroot/modules/custom/mass_translations/mass_translations.module
tom-fleming: Good point. This hides the "Use existing media" text that shows up by default, and I added a note.

PR #563, opened 2020-12-15
Mayflower/dp 20691 mayflower version 10.4.1

ygannett:
Changed:
  - description: |-
          Updated Mayflower version to 10.4.1.
              - DP-20682: Significantly reduced Noto Sans Latin variable font `.woff` file sizes, from ~900k to ~60k. (MF)
              - DP-20682: Add conditionals to import static fonts and variable fonts `.woff2`, in order to prevent both fonts to be downloaded. (MF)
              - DP-20681: Add polyfill for includes. (MF)
    issue: DP-20691

### DIFF
--- a/changelogs/DP-20691.yml
+++ b/changelogs/DP-20691.yml
@@ -1,0 +1,47 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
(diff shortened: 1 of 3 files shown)


ygannett:
Backstop failures are from rendering timing and MF change.

PR #565, opened 2020-12-16
Dp 20703 font preload

joeg8612:
**Description:**
Changed font preloads to use woff2 formats.

https://jira.mass.gov/browse/DP-20703

**To Test:**
- [ ] Load home page and inspect with network view. Verify that woff2 fonts are loaded but woff fonts are not loaded. There should be 2 font files loaded rather than 4.

**Screenshots/GIFs:**

---

### DIFF
--- a/changelogs/DP-20703.yml
+++ b/changelogs/DP-20703.yml
@@ -1,0 +1,41 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
+# #############################################################################
+# You can use one of the following types:
+#  - Added: For new features.
+#  - Changed: For changes to existing functionality.
+#  - Deprecated: For soon-to-be removed features.
+#  - Removed: For removed features.
+#  - Fixed: For any bug fixes.
+#  - Security: In case of vulnerabilities.
+#
+# Format
+# #############################################################################
+# The format is crucial. Please follow the examples below. For reference, the requirements are:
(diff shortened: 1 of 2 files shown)


weitzman:
FYI Drupal's new theme uses a similar technical and adds a . `type="font/woff2"` attribute. Not sure how important that is. See https://www.drupal.org/files/issues/2020-10-29/3176910-13.patch

tom-fleming:
@weitzman I think this is probably the best explanation I've found: https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content#Including_a_MIME_type. 

TL;DR - Setting the MIME type helps with preloading by allowing the browser to determine support first. 

With that said, pretty much all modern browsers support woff2, so I'm not too concerned that we're not including that, but it might help with some older browsers that don't support the format like IE11, though in my own test, the woff2 font did not load, which is what we want. IE11 must have another way to determine that it shouldn't load the font.

tom-fleming:
I will add though that it looks like in IE11 we are no longer falling back on an EOT-format font and are instead falling back on probably Helvetica. Either way, no fonts are loaded in IE11 with this change.

tkoleary:
@tom-fleming `"Noto Sans","Helvetica",Arial, sans-serif;` are the fallbacks in the CSS `font-family` attribute but there are also fallbacks in the @font-face rules for "Noto Sans VF" and "Noto Sans Italic VF"

PR #575, opened 2021-01-06
Remove inaccessible links in entity references

weitzman:
**Description:**
Add #access to inaccessible links (i.e. deleted or unpublished referenced content), taking care to also add cacheability metadata. Unfortunately, this makes the links uncacheable (max-age=-1) but thats not so bad - they will process via dynamic page cache. 

The `{% if item.content|render|striptags|trim %}` in twig files is needed or else #access is not honored as the individual links are rendered in the design system, not in drupal.

Todo - The bad ref at bottom https://www.mass.gov/info-details/the-community-reinvestment-act-cra-for-banks-and-credit-unions is vexing. This one is in a LinkList and those deeply drift away from Drupal rendering starting at the mass_theme_preprocess_node_info_details() function https://github.com/massgov/openmass/blob/develop/docroot/themes/custom/mass_theme/mass_theme.theme#L1514

https://jira.mass.gov/browse/DP-19228

**To Test:**
- [ ] Try the links in the Jira ticket and see that their inaccessible content has been removed while anonymous. They are removed from editors too if the content is actually inaccessible.

---

### DIFF
--- a/docroot/modules/custom/mass_content/src/Plugin/Field/FieldFormatter/DynamicLinkSeparateFormatter.php
+++ b/docroot/modules/custom/mass_content/src/Plugin/Field/FieldFormatter/DynamicLinkSeparateFormatter.php
@@ -93,8 +93,6 @@ class DynamicLinkSeparateFormatter extends LinkFormatter {
         $element[$delta]['#access'] = $access->isAllowed();
       }
 
-
-
       if (!empty($item->_attributes)) {
         // Set our RDFa attributes on the <a> element that is being built.
         $url->setOption('attributes', $item->_attributes);
--- a/docroot/modules/custom/mass_content/src/Plugin/Field/FieldFormatter/DynamicLinkSeparateFormatter.php
+++ b/docroot/modules/custom/mass_content/src/Plugin/Field/FieldFormatter/DynamicLinkSeparateFormatter.php
@@ -2,11 +2,13 @@
 
 namespace Drupal\mass_content\Plugin\Field\FieldFormatter;
 
+use Drupal\Core\Access\AccessResult;
+use Drupal\Core\Access\AccessResultForbidden;
+use Drupal\Core\Cache\CacheableMetadata;
 use Drupal\Core\Field\FieldItemListInterface;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\link\Plugin\Field\FieldFormatter\LinkFormatter;
 use Drupal\mass_content\Field\FieldType\DynamicLinkItem;
-use Drupal\mayflower\Helper;
 
 /**
  * Plugin implementation of the 'dynamic_link_separate' formatter.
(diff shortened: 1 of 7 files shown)


docroot/themes/custom/mass_theme/templates/field/field--node--field-related-links--curated-list.html.twig
```diff
@@ -54,19 +54,21 @@
...
+              "href": item.content['#url'],
+              "text": item.content['#title'],
+              "info": "",
+              "property": "",
+            },
+            "level": level,
+            "date": item.content['#extra']['date'],
+          }
+        } %}
+      {% endif %}
```
weitzman, line 71: Since this template passes raw item values into Mayflower press-teaser twig template, Drupal never renders and respects any #access that the link may have.

docroot/modules/custom/mass_content/src/Plugin/Field/FieldFormatter/DynamicLinkSeparateFormatter.php
```diff
@@ -78,6 +80,21 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
         '#extra' => $extra,
       ];
 
+      // Add #access and cacheability metadata to remove links to deleted and unpublished content.
+      if ($url->isRouted() && $url->getRouteName() == 'entity.node.canonical') {
+        if ($entity = \Drupal::entityTypeManager()->getStorage(key($url->getRouteParameters()))->load($url->getRouteParameters()['node'])) {
```
rbayliss, line 85: Oof - this may have a pretty major performance impact, especially for pages like curated list.

weitzman: You quoted the entity load line - is that what you are worried about, or is it the entity access check? I'm not too worried about the entity load. Those should be in entity cache. I dont think the access checks are cached, but they could be since all their contexts, etc are known.

weitzman: An alternative is to queue up deletion of bad entity refs upon entity save but thats gonna cause a lot of invalidation. See my recent comment at https://www.drupal.org/project/drupal/issues/2723323#comment-13956557

rbayliss: I'm worried about both loading and access, yeah.  Consider the curated list point for a sec - We're talking about hundreds or thousands of entities that on some of those lists.  Even if all of the entities are in the cache, they're probably not in the static cache, so we've got to pull a lot of data from cache all of a sudden.

I'd be a bigger fan of pruning, which would move the extra work to a separate process we could control.  Understand your point on the invalidation, but this site doesn't actually do that much deletion to begin with, so it might not be so bad.

weitzman: Pruning doesnt actually help us if referenced content gets unpublished or trashbin. We would want the reference to be temporarily removed until the reference is published again. I'll double check with Joe about this requirement.

rbayliss: Yeah - I'm not trying to block this. I'm not sure what the requirements or urgency are, and it's totally possible this is no big deal, but I'm concerned that even if this one isn't a big deal, it commits us to a direction that will eventually end up with these kinds of checks being much more widely used in the codebase, so we should be pretty sure it's scalable.

rbayliss: FWIW, I do recall quite a bit of work being done at one point to hide unpublished things from showing up. Not sure where you can find that or what form it's in these days, but I do know it was in place.

weitzman:
Abandoned for now. See Jira.

PR #590, opened 2021-01-26
DP-20824: Change fields in Drupal to disallow content types that shouldn't be referenced

arthurbaghdas:
**Description:**
Changed field configs to disallow certain content types

https://jira.mass.gov/browse/DP-20824


tom-fleming:
LGTM

PR #597, opened 2021-02-01
Exclude menu_link_content from sanitization - DP-21067

tom-fleming:
**Description:**
While working on DP-20281, we noticed that some pages, especially edit pages, did not load when the Language module was enabled and menu link content was sanitized from the database.

https://jira.mass.gov/browse/DP-21067

**To Test:**
- [ ] Locally, run `ahoy drush sql-sanitize --sanitize-entities` against the less sanitized database. 
- [ ] Enable the Language module. Try editing a node. Page should load.
- [ ] For comparison, enable the Language module using the super sanitized database. Try editing a node. The page will likely not load.

**Screenshots/GIFs:**

---

### DIFF
--- a/drush/Commands/SanitizationCommands.php
+++ b/drush/Commands/SanitizationCommands.php
@@ -49,23 +49,25 @@ class SanitizationCommands extends DrushCommands {
       return;
     }
     $types = array_keys($this->entityTypeManager->getDefinitions());
+
+    // Deleting menu link content sometimes results in rendering issues with
+    // super-sanitized databases, so excluding it from sanitization.
+    $idx = array_search('menu_link_content', $types);
+    if ($idx !== -1) {
+      unset($types[$idx]);
+    }
+
+    // This is a bit of a hack, but content moderation needs to go dead last
+    // so it can clean up after all the other stuff has been deleted.
+    $idx = array_search('content_moderation_state', $types);
+    if ($idx !== -1) {
+      unset($types[$idx]);
+      $types[] = 'content_moderation_state';
+    }
+
     foreach ($types as $idx => $type) {
-
-      // Deleting menu link content sometimes results in rendering issues with
-      // super-sanitized databases, so excluding it from sanitization.
-      if ($type == 'menu_link_content') {
-        unset($types[$idx]);
-      }
-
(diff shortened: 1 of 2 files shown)


drush/Commands/SanitizationCommands.php
```diff
@@ -49,15 +49,21 @@ public function sanitizeEntities($result, CommandData $commandData) {
...
-    if ($idx !== -1) {
-      unset($types[$idx]);
-      $types[] = 'content_moderation_state';
-    }
+    foreach ($types as $idx => $type) {
+
+      // Deleting menu link content sometimes results in rendering issues with
+      // super-sanitized databases, so excluding it from sanitization.
+      if ($type == 'menu_link_content') {
+        unset($types[$idx]);
```
weitzman: Does this actually accomplish the goal? Looks to me like the default entity sanitizer gets picked and its sanitize() still runs.

drush/Commands/SanitizationCommands.php
```diff
@@ -49,6 +49,14 @@ public function sanitizeEntities($result, CommandData $commandData) {
       return;
     }
     $types = array_keys($this->entityTypeManager->getDefinitions());
+
+    // Deleting menu link content sometimes results in rendering issues with
+    // super-sanitized databases, so excluding it from sanitization.
+    $idx = array_search('menu_link_content', $types);
+    if ($idx !== -1) {
```
weitzman, line 56: Why the -1 check in addition to the array_search?

tom-fleming: Mostly for consistency's sake in the end. See line [63](https://github.com/massgov/openmass/pull/597/files#diff-fa3dc0c7a267917545b458fa2f1bb1d7306f4da39e4199aa8b9ffd43455a6d3aR63) below this. I tried looking into why that check might be there as well, and my best educated guess is that they wanted to verify that the result was actually true and not set explicitly to -1. This was done either out of an abundance of caution or because they ran into an issue (though I cannot imagine a situation where this would actually occur), since array_search will evaluate -1 as true.

weitzman: OK

PR #598, opened 2021-02-01
Mayflower/dp 21074 mayflower version 11.1.3

ygannett:
Changed:
  - description: |-
        Updated Mayflower version to 11.1.3.
          - DP-19859: Fix the Relationship Indicator display at the show all state in IE. (MF)
          - DP-21059: Make a skip link target to be displayed only when its associated skip link is clicked. (MF)
    issue: DP-21074


ygannett:
backstop failures are from rendering timing.  They are not actual errors.

PR #588, opened 2021-01-25
DP-20580: WYSIWYG editor on Info Detail pages is not allowing for image uploads v2

arthurbaghdas:
**Description:**
Modified Basic HTML text format.

**Jira:** (Skip unless you are MA staff)
https://jira.mass.gov/browse/DP-20580

PR #602, opened 2021-02-03
Pin web server during file uploads in the WYSIWYG editor - DP-20580

tom-fleming:
**Description:**
See https://support.acquia.com/hc/en-us/articles/360004147834-Pinning-to-a-web-server-without-using-the-hosts-file#defineacookie

https://jira.mass.gov/browse/DP-20580

**To Test:**
- [ ] Go to add an info details node
- [ ] Inspect the page, and check that the cookie is not set
- [ ] In the Overview field, select the image upload icon
- [ ] Inspect the page again and verify that the cookie is set with a server name value (might be more validating to test in an Acquia environment in this case)
- [ ] Reload the page, and check that the cookie is no longer set

**Screenshots/GIFs:**

---

### DIFF
--- a/docroot/themes/custom/mass_admin_theme/mass_admin_theme.theme
+++ b/docroot/themes/custom/mass_admin_theme/mass_admin_theme.theme
@@ -227,10 +227,22 @@ function _add_description_display_before(&$variables) {
  */
 function mass_admin_theme_preprocess_form_element(&$variables) {
   _add_description_display_before($variables);
+
+  // Remove cookie when in the admin theme to ensure that the cookie is reset.
+  setrawcookie('ah_app_server', '', REQUEST_TIME - 3600, '/');
+
   if (isset($variables['element']['#id']) &&
     ($variables['element']['#id'] === 'edit-upload')
   ) {
     $variables['prefix'] = t('Please note, only one image can be embedded at a time.');
+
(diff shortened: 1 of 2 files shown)


docroot/themes/custom/mass_admin_theme/mass_admin_theme.theme
```diff
@@ -227,10 +227,22 @@ function _add_description_display_before(&$variables) {
  */
 function mass_admin_theme_preprocess_form_element(&$variables) {
   _add_description_display_before($variables);
+
+  // Remove cookie when in the admin theme to ensure that the cookie is reset.
+  setrawcookie('ah_app_server', '', REQUEST_TIME - 3600, '/');
```
weitzman, line 232: A more API-ish way to do this is https://drupal.stackexchange.com/a/293281/549 but not a big deal at all.

weitzman (commented):
Can we remove the dropzone patch added in https://github.com/massgov/openmass/pull/588/files? If so lets do that here.

docroot/themes/custom/mass_admin_theme/mass_admin_theme.theme
tom-fleming: Thanks. Adjusted.

PR #603, opened 2021-02-08
DP-21109: Add generic label feature to allow content tagging

arthurbaghdas:
**Description:**
Added field_reusable_label to the content types, configured form/view displays, created custom mg_labels metatag and mass_labels token.

https://jira.mass.gov/browse/DP-21109

**To Test:**
I would recommend to create a node of each type and see the page for mg_labels tag existence and visual issues.


weitzman:
Sorry for just dropping in here with limited knowledge about this feature. If I understand the requirement right (free tagging), Drupal has long had a good solution for this. See https://drupal.stackexchange.com/questions/211337/drupal-8how-to-add-tags-for-custom-content-types#211339. We already have a vocabulary called Tags which is unused. I would use that, and add same field to each content type which references terms in that vocab. Pick `autocomplete (Tags style)` and you have good UX when entering in tags.

arthurbaghdas:
We already had field_reusable_label used in several content types, so that field was reused with same options and same vocabulary.

weitzman:
In the future, lets use mass_fields for field related stuff like `getAllLabelsFromNode` and mass_metatag can be for strictly metatag stuff. We plan to move away from metatag module one day so this separation will be helpful.

PR #604, opened 2021-02-09
Mayflower/dp 21174 mayflower version 11.1.4

ygannett:
Changed:
  - description:  |-
        Updated Mayflower version to 11.1.4.
          - DP-20381: Fix left floated figure components and images to cover list style elements in rich text containers. (MF)
    issue: DP-21174


ygannett:
The Backstop failures are from rendering time difference when the screenshots were taken.  Not actual failures.

PR #611, opened 2021-02-11
DP-21191

tom-fleming:
**Description:**
Changes from DP-20580 caused issues on admin pages. This change in approach helps to address the issue and can be tested on feature4.

https://jira.mass.gov/browse/DP-21191

**To Test:**
- [ ] Open your browser's inspector and go to the Application tab and then Cookies
- [ ] Go to add a node (e.g. info details) and select the image upload, you should see an ah_app_server cookie added
- [ ] Complete adding the image
- [ ] Refresh the page and the cookie should no longer be set

**Screenshots/GIFs:**

---

### DIFF
--- a/docroot/modules/custom/mass_admin_pages/src/EventSubscriber/CookieSubscriber.php
+++ b/docroot/modules/custom/mass_admin_pages/src/EventSubscriber/CookieSubscriber.php
@@ -4,16 +4,9 @@ namespace Drupal\mass_admin_pages\EventSubscriber;
 
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Drupal\Core\EventSubscriber\FinishResponseSubscriber;
-use Symfony\Component\HttpFoundation\Response;
-use Symfony\Component\HttpFoundation\Request;
-use Drupal\Core\Cache\Context\CacheContextsManager;
-use Drupal\Core\Config\ConfigFactoryInterface;
-use Drupal\Core\Language\LanguageManagerInterface;
-use Drupal\Core\PageCache\RequestPolicyInterface;
-use Drupal\Core\PageCache\ResponsePolicyInterface;
 use Symfony\Component\HttpFoundation\Cookie;
-use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
-use Symfony\Component\HttpKernel\KernelEvents;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
 
 /**
  * Response subscriber to add cookies.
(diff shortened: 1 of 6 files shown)


docroot/modules/custom/mass_admin_pages/src/EventSubscriber/CookieSubscriber.php
```diff
@@ -0,0 +1,49 @@
...
+  protected function setResponseNotCacheable(Response $response, Request $request) {
+    parent::setResponseNotCacheable($response, $request);
+
+    if (\Drupal::service('router.admin_context')->isAdminRoute()) {
+      $response->headers->clearCookie('ah_app_server', '/');
+
+      // Set cookie to stick file uploads to one server when on edit pages to
+      // address issues with uploading files via the WYSIWYG.
+      // See https://support.acquia.com/hc/en-us/articles/360004147834-Pinning-to-a-web-server-without-using-the-hosts-file#defineacookie
+      if ($request->attributes->get('_route') == 'entity_browser.browse_files') {
```
weitzman: Do we need to clear the cookie? It will expire in one day anyway.

 Also, why match that route browse_files? Can you upload without hitting that route? We could set the cookie for all admin route requests.

tom-fleming: To answer both questions, I did it this way because of the Important disclaimer here: https://docs.acquia.com/cloud-platform/manage/files/broken/#pinning-temporarily-to-a-web-server. It's my understanding as well that this will effectively pin all requests for that session to the same server, not just file uploads. So, I wanted to do this with as little footprint as possible and hopefully avoid a future hot patch if for some reason one day one of our web servers start seeing too much activity and some authors end up stuck on that server, which, of course, we could work around.

weitzman: I think that disclaimer is important only if you are pinning all your requests. We are only pinning editors so no load issues apply IMO.

tom-fleming: Makes sense. I made the change, and this is ready for review again

docroot/modules/custom/mass_admin_pages/src/EventSubscriber/CookieSubscriber.php
```diff
@@ -0,0 +1,44 @@
...
+   * {@inheritdoc}
+   */
+  public function __construct(LanguageManagerInterface $language_manager, ConfigFactoryInterface $config_factory, RequestPolicyInterface $request_policy, ResponsePolicyInterface $response_policy, CacheContextsManager $cache_contexts_manager, $http_response_debug_cacheability_headers = FALSE) {
+    parent::__construct($language_manager, $config_factory, $request_policy, $response_policy, $cache_contexts_manager, $http_response_debug_cacheability_headers);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setResponseNotCacheable(Response $response, Request $request) {
```
weitzman: Is there a benefit to overloading setResponseNotCacheable() with our cookie logic? I propose we implement onAllResponds() method. Otherwise code looks good.

weitzman (approved):

docroot/modules/custom/mass_admin_pages/src/EventSubscriber/CookieSubscriber.php
tom-fleming: If we're just checking for admin paths, no, not really, beyond I guess just using $response directly. Updated the PR

PR #613, opened 2021-02-12
Enable New Relic measument of CLI PHP commands.

weitzman:
Once this goes out, I will visit most of the scheduled jobs and add: 

- ` PHP_INI_SCAN_DIR=~` to the fron of the command, thus loading the newrelic php extension
- Optionally set `--nrname=foo` to clarify what job name shows up at New Relic

DP-13674
---

### DIFF
--- a/drush/Commands/NewRelicCommands.php
+++ b/drush/Commands/NewRelicCommands.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace Drush\Commands;
+
+use Consolidation\AnnotatedCommand\CommandData;
+use Drush\Commands\DrushCommands;
+
+class NewRelicCommands extends DrushCommands
+{
+
+  /**
+   * @hook option *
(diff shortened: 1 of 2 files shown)


drush/Commands/NewRelicCommands.php
```diff
@@ -0,0 +1,35 @@
...
+use Consolidation\AnnotatedCommand\CommandData;
+use Drush\Commands\DrushCommands;
+
+class NewRelicCommands extends DrushCommands
+{
+
+  /**
+   * @hook option *
+   *
+   * @option nr-name New Relic transaction name.
```
tom-fleming: @weitzman Will the option need to be `--nr-name`, rather than `--nrname` described in the description, with this set?

weitzman: Fixed

PR #618, opened 2021-02-19
Update Drush, composer2, and Media Entity Download

weitzman:
**Description**
- Drush is upgraded to fix broken pm:security-php command which blocks PP checks.
- Composer is upgraded to v2 in the container to fix `ahoy comi` which tripped up Mingua and will trip up new people to the team. Its also way faster and uses less memory.
- I upgraded Media Entity Download because its patch would not apply using Composer2. That’s fixed in [later version of the patch](https://www.drupal.org/node/2928318).

Composer reroll: `composer2 update drush/drush cweagans/composer-patches drupal/migrate_tools drupal/devel drupal/media_entity_download --with-dependencies`

https://jira.mass.gov/browse/DP-21093

**To Test:**
- [ ] Run drush pm:security-php. If it finishes, we have fixed the problem in the Jira ticket.
- [ ] Run `ma:release` to a feature env. Make sure that CI job is green. This is due diligence on the Drush upgrade.
- [ ] Link to a media entity using the WYSIWYG. Make sure the resulting download link works for an anon user. As anon, click on any other media download links on the site and make sure they work.

**After merge**
- Post in SSR channel: It is recommended that you upgrade to Composer2 at your earliest convenience. Run `composer self-update`. In the near future this will be a requirement. If this feels burdensome, you can pull latest code, `ahoy down && ahoy up`, and then run composer commands inside the container. This is feasible with the lower memory footprint of Composer2.

---


tom-fleming:
@weitzman I'm getting this error when I try to run the security check command:
> Client error: `POST https://security.symfony.com/check_lock` resulted in a `403 Forbidden` response:
>  error code: 1020

Composer is installing drush 10.3.6. Do we actually want to be on 10.4.0?

weitzman:
Drush 10.4 is incompatible with migrate tools. I think we can just disable and remove migrate? That would help our D9 progress a lot.

tom-fleming:
> Drush 10.4 is incompatible with migrate tools. I think we can just disable and remove migrate? That would help our D9 progress a lot.

That makes sense to me, but it looks like we do have a couple custom modules that have migrate as a dependency: mass_moderation_migration and mass_migration. And then mass_content is also using migrate_plus, which has a migrate dependency. It's possible that whatever those modules were doing is no longer needed though?

weitzman:
Thanks. Some more research:

- mass_moderation_migration  is disabled. can be removed.
- mass_migration. defines migrate process plugins for migration into massdocs which I think we are not using anymore. So can be disabled and removed.
- mass_content. looks like migrate was used for prototyping 5 years ago. See https://github.com/massgov/mass/pull/9. This is all dead code AFAICT.

weitzman:
I also uninstalled migrate_tools and mass_migration

tom-fleming:
LGTM

tom-fleming (approved):

PR #620, opened 2021-02-22
DP-10401: Hit a 500 trying to view the descendant manager overview

arthurbaghdas:
**Description:**
Label method changed to be getLabel.

https://jira.mass.gov/browse/DP-10401


vordude:
Changes make sense and it works for me locally 

:+1:

PR #627, opened 2021-03-01
Mayflower/dp 21326 mayflower version 11.2.0

ygannett:
Changed:
  - description:  |-
        Updated Mayflower version to 11.2.0.
          - DP-21258: Add BrandBanner molecule in Patternlab and generate the HTML, CSS and JS for Mayflower core documentation. (MF)
    issue: DP-21326


ygannett:
All failures are from the alert date difference and the Tableau rendering timing difference.  Not actual errors.

PR #647, opened 2021-03-18
DP-21459: Fixing the missing Media Entity Download WYSIWYG button icon

vordude:
**Description:**
The Media Entity Download WYSIWYG Icon was not appearing because of an issue with the Drupal patch required to entirely implement this plugin.

https://jira.mass.gov/browse/DP-21459

**To Test:**
- [ ] Look a the basic HTML WYSIWYG and others, and confirm this icon appears
mediaentitydownload

Previous broken behavior was a blank button that was functional.


vordude:
Behat is running out of memory before completion.

vordude:
Let's no do this this way.

PR #652, opened 2021-03-22
DP-21464: Half-width image renders as full width in rich text with no alignment

weekbeforenext:
Moved the desktop width CSS from the **right-align** and **left-align** classes to **article[data-entity-embed-display-settings="embedded_half_width"]**.

**Description:**
CSS change to the mass_theme overrides CSS.

**Jira:**
https://jira.mass.gov/browse/DP-21464

**To Test:**
- [ ] Copy the Content Rich Text markup from https://massgovfeature4.prod.acquia-sites.com/info-details/what-to-do-if-you-received-a-massnotify-exposure-notification to an Info Detail node.
- [ ] Verify the image as half-width and alignment set to none is showing at 50% width.
- [ ] Verify the right-aligned and left-aligned images below are also still at 50% width.

**Screenshots/GIFs:**
Feature4:
feature4_half-width_align-none
Fixed:
local_half-width_align-none

---

### DIFF
--- a/changelogs/DP-21464.yml
+++ b/changelogs/DP-21464.yml
@@ -1,0 +1,41 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
+# #############################################################################
+# You can use one of the following types:
+#  - Added: For new features.
+#  - Changed: For changes to existing functionality.
+#  - Deprecated: For soon-to-be removed features.
+#  - Removed: For removed features.
+#  - Fixed: For any bug fixes.
+#  - Security: In case of vulnerabilities.
+#
+# Format
+# #############################################################################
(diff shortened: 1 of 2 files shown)


ygannett:
@weekbeforenext Your initial change looks good.  I noticed that horizontal margins with float are not consistent with 50% width.  As you see in the screenshot, the left aligned image has 50px right margin, but the right aligned image's left margin is not 50px.  It's not the initial change in the ticket, but can you fix that as well?

The left aligned one's right margin is set as inline.  Probably coming from JS somewhere, but I don't think it's necessary to complicate it.  I think it would be sufficient that adding another style to the same file to add 50px left margin to a right aligned image.

right-float-left-margin

ygannett (approved):
LGTM.

PR #658, opened 2021-03-29
DP-21300: Add Data Topic Taxonomy and Topic/Sub Topic fields

tom-fleming:
**Description:**
Adds a data topic taxonomy vocabulary along with topic and sub topic fields. 

https://jira.mass.gov/browse/DP-21300

**To Test:**
- [ ] After pulling down and updating your branch, you will need to create Data Topic Taxonomy terms here: http://mass.local/admin/structure/taxonomy/manage/data_topic/overview. This should be built with a two-level hierarchy like the below screenshot: 
image
- [ ] Check all content types for the topic and sub topic fields as described in Done When section of the [Jira ticket](https://jira.mass.gov/browse/DP-21300)

**Screenshots/GIFs:**

---


weitzman:
Could we say a bit more about why there are two Drupal fields, topic and subtopic? If we had just a subtopic field, could we not could derive the topic with $subtopic->get('parent').

tom-fleming:
@weitzman Yes, we could derive the topic from the subtopic. One reason this approach was taken was to create a specific user experience where the user first selects only one top-level topic and can then choose multiple subtopics, with each having their own help text, and this was the quickest and cleanest path I could see for implementation, especially after evaluating some of the contrib options available.

But also, when we create the views, which includes a top-level topic view of nodes and a subtopic view of nodes, each with their own set of exposed filters, I can see this approach simplifying the creation and underlying queries of these views somewhat, since I think we would only need to reference the taxonomy relationship and hierarchy directly for the subtopic filters.

If you'd like to touch base on the designs, let me know. I think if there's any uneasiness about this approach, we should address it in this ticket.

weitzman (approved):

PR #664, opened 2021-03-30
DP-19362: D9 upgrade packages - authoring/fields

arthurbaghdas:
**Description:**
Updating packages

https://jira.mass.gov/browse/DP-19362


weitzman:
Lets upgrade drupal/auto_entitylabel here, or spinoff a followup if needed. The current version breaks the `drush php` command (just type `$container` to see the failure)

arthurbaghdas:
We have a follow-up ticket for that @weitzman DP-21622

PR #682, opened 2021-04-07
DP:19203 Fixing Custom Module Code Deprecations

vordude:
**Description:**
This updates the custom module code according to the Update Status module.

This should across the board cease use of deprecated Drupal APIs and allow custom code to be executed in both a Drupal 8 and an Drupal 9 site.

There should be no functional changes, except for the update status module will show that the custom code on the site is ready for Drupal 9.

https://jira.mass.gov/browse/DP-19203

**To Test:**
This is the hard part, since there is nothing to test. There should be no changes in end user interactions or behaviours.

- [ ] Successfully complete all automated tests
- [ ]  Manual review of the pull request diff
- [ ] General smoke test of the site, viewing, adding, editing content. 
- [ ]  Find a better way to test this?


weitzman (approved):
Epic work here!

PR #684, opened 2021-04-12
DP-21034 login error message type

ygannett:
**Description:**
Override the message type "status" with "error" for the message "Unrecognized username or password."
The messages are configured in core.  Instead of touching core, override the type for the particular message.

https://jira.mass.gov/browse/DP-21034

**To Test:**
- [ ] Go to the login page and type in incorrect user name and password to cause the login error.
- [ ] Find the error message is in a red box for _error_ instead of green for _status_.

**Screenshots/GIFs:**
error-style

---

### DIFF
--- a/changelogs/DP-21034.yml
+++ b/changelogs/DP-21034.yml
@@ -1,0 +1,41 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
+# #############################################################################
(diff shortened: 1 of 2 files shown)


weitzman:
I would prefer to understand why this form error is resulting in a status message instead of an error message. We should aim to fix root cause. There may be more forms misbehaving like this. Another thought is that the new TFA has altered this form somehow.

PR #686, opened 2021-04-13
DP-21633: Create content inventory view org page

arthurbaghdas:
**Description:**
Created a view with data export.

https://jira.mass.gov/browse/DP-21633


tom-fleming:
LGTM!

PR #691, opened 2021-04-16
Added deployment for unpublished header media items - DP-21704

tom-fleming:
**Description:**
The migration of header media image wrapping fields missed migrating unpublished nodes. These fields are not editable by authors, so these fields should match the defaults for both published and unpublished content.

https://jira.mass.gov/browse/DP-21704

**To Test:**
- [ ] Pull down a fresh database. Do not run ahoy updatedb yet.
- [ ] Log in and go to to http://mass.local/info-details/summary-of-the-methodology-for-the-investigation-of-the-drug-laboratory-at-the-william-a-hinton-state-laboratory-investigative-process---unpublished?auHash=Pxu_czkyEaOaZ6hw0g6DkZXbF7Vc_Mm1YVYWP0SzqYM. Note that the header image does not extend the full width of the content area.
- [ ] Run ahoy updatedb.
- [ ] Check the page linked above again. Verify that the header image extends the full width of the content area.

**Screenshots/GIFs:**

---

### DIFF
--- a/docroot/modules/custom/mass_content/mass_content.deploy.php
+++ b/docroot/modules/custom/mass_content/mass_content.deploy.php
@@ -262,11 +262,11 @@ function mass_content_deploy_image_section_fields_followup(&$sandbox) {
 /**
  * Migrate followup for Header Media image fields.
  */
-function mass_content_deploy_header_media_images_followup(&$sandbox) {
+function mass_content_deploy_header_media_images_followup(&$sandbox, $status = 1) {
   $_ENV['MASS_FLAGGING_BYPASS'] = TRUE;
 
   $query = \Drupal::entityQuery('node')
-    ->condition('status', 1)
+    ->condition('status', $status)
     ->condition('type', 'info_details')
     ->condition('field_info_details_header_media.entity:paragraph.field_image.target_id', "", "!=");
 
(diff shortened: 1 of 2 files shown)


docroot/modules/custom/mass_content/mass_content.deploy.php
```diff
@@ -317,3 +317,10 @@ function mass_content_deploy_header_media_images_followup(&$sandbox) {
     return t('All Header Media image caption fields migrated.');
   }
 }
+
+/**
+ * Migrate unpublished Header Media image wrapping fields.
+ */
+function mass_content_deploy_header_media_images_unpublished(&$sandbox) {
+  mass_content_deploy_header_media_images_followup($sandbox, 0);
```
tom-fleming, line 325: I modified the previous deploy hook to allow a status argument. The only difference between both field migrations is status, and in this case we want to only modify unpublished content.

docroot/modules/custom/mass_content/mass_content.deploy.php
```diff
@@ -317,3 +316,10 @@ function mass_content_deploy_header_media_images_followup(&$sandbox) {
     return t('All Header Media image caption fields migrated.');
   }
 }
+
+/**
+ * Migrate unpublished Header Media image wrapping fields.
+ */
+function mass_content_deploy_header_media_images_unpublished(&$sandbox) {
+  mass_content_deploy_header_media_images_followup($sandbox);
```
tom-fleming, line 324: @weitzman For this ticket, this deploy hook needs to do mostly the same thing as the previous deploy hook except it needs modify all field values regardless of the content's publication status. I am slightly hesitant to just rename and modify the old deploy hook since those names get stored in key_value, but is there a strong case either way between going with the approach I'm taking here vs renaming an old deploy hook?

weitzman: I think this approach is fine, except `mass_content_deploy_header_media_images_unpublished` now operates on both published and unpublished. If that’s desired then perhaps rename `mass_content_deploy_header_media_images_unpublished`.

PR #718, opened 2021-04-28
DP-21745 Alias duplicates

tom-fleming:
**Description:**
Addresses issues with duplicate aliases when mass_translations is enabled.

https://jira.mass.gov/browse/DP-21745

**To Test:**
- [ ] Create a node with the same title as another piece of content. Check that the URL is automatically appended with -0 or something similar to make the URL unique.
- [ ] Repeat the same steps with a Document
- [ ] Create a node translation (e.g. Info Details). Check hreflang links and links in the Translations tab
- [ ] Create a document translation. Check hreflang links and links in the Translations tab
- [ ] Repeat the same steps with a node and document that has Language set to None. See http://mass.local/admin/config/search/path?sort=desc&order=Language

**Screenshots/GIFs:**

---


weitzman (approved):
Code looks good

PR #724, opened 2021-05-03
DP-21793: Add character countdown and warning when page titles exceed 70 characters

arthurbaghdas:
**Description:**
Patched Maxlength module, configured fields on the site.

https://jira.mass.gov/browse/DP-21793

**To Test:**
- [ ] Try adding content of each type and see how maxlength module behaves.


conf/drupal/config/core.entity_form_display.node.advisory.default.yml
```diff
@@ -83,7 +85,6 @@ third_party_settings:
       weight: 13
       format_type: tab
       format_settings:
-        label: Details
```
weekbeforenext: This seems unrelated. It doesn't seem to make a difference when looking at the form, as far as I can tell though.

conf/drupal/config/core.entity_form_display.node.advisory.default.yml
```diff
@@ -113,7 +114,6 @@ third_party_settings:
       weight: 11
       format_type: fieldset
       format_settings:
-        label: '  '
```
weekbeforenext: This might also be unrelated.

weekbeforenext (approved):
I noted a couple of odd config changes, but this looks good. If the config changes I commented on are expected, this is good to merge after resolving conflicts.

conf/drupal/config/core.entity_form_display.node.advisory.default.yml
arthurbaghdas: Fixed, thanks

conf/drupal/config/core.entity_form_display.node.advisory.default.yml
arthurbaghdas: Fixed, thanks

PR #742, opened 2021-05-10
DP-21034: Edit log in error message with wrong styles

arthurbaghdas:
**Description:**
Fixed the method called on custom validation function.

https://jira.mass.gov/browse/DP-21034

**To Test:**
- [ ] Try to login with wrong credentials

**Screenshots/GIFs:**
Screen Shot 2021-05-10 at 15 20 05

### DIFF
--- a/changelogs/DP-21034.yml
+++ b/changelogs/DP-21034.yml
@@ -1,0 +1,41 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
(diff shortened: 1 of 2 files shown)


weitzman (approved):
The change looks good, thanks for finding the root cause.

The function description says "Moves username inline form errors on login page to the top of the page." The function only adds an error, it doesn’t move anything. I'm still approving this since that bad description is a pre-existing issue.

PR #752, opened 2021-05-13
DP-21790: Add labels to feedback manager, fix multiselect

roberttstephens:
**Description:**

https://jira.mass.gov/browse/DP-21790

**To Test:**
- Add configuration to point to production feedback manager instance. Reach out to @serundeputy for details if needed
- Log in
- Go to /admin/ma-dash/feedback
- TODO

**Screenshots/GIFs:**

---


serundeputy:
@roberttstephens this is working on my local manual testing with filtering with `label_id`'s and various combinations.

There is a cirle failure on importing `oktokit` node module, but I did not see where that is happening via grep:

```bash
grep -Erin "ocktokit" docroot
```

Circle is not seem to be indicating which file the error is coming from. Do you think this is something we can easily fix up?

PR #757, opened 2021-05-18
DP-22007: SPIKE Perform assessment of entity usage module

weekbeforenext:
**Description:**
Discover work in progress.

https://jira.mass.gov/browse/DP-22007

**To Test:**
- Login and go to /doc/dcr-media-advisory-dcr-officials-welcome-returning-massachusetts-firefighters/mass-usage.
- Verify the tab is “Pages linking here”.
- Verify the table headers are:
- Entity
- ID
- Content Type
- Field name
- Status
- Verify the data looks accurate for each table heading.
- Verify there is one result (this should match the Descendant Manager “Pages linking here” on Production).
- Go to /orgs/massachusetts-department-of-revenue/mass-usage.
- Verify the tab is “Pages linking here”.
- Verify the table headers are:
- Entity
- ID
- Content Type
- Field name
- Status
- Verify the data looks accurate for each table heading.
- Verify the results match the Descendant Manager “Pages linking here” on Production.
- Verify the results are in alphabetical, ascending order by the Entity column.
- Verify the pager works as expected.

**Screenshots/GIFs:**

---


composer.json
```diff
@@ -319,6 +320,10 @@
             "drupal/entity_browser": {
                 "Fix duplicated image selection when click use selected button": "https://www.drupal.org/files/issues/prevent-duplication.patch"
             },
+            "drupal/entity_usage": {
+                "2954973 - Track absolute URLs in link fields": "https://www.drupal.org/files/issues/2021-08-17/2954973-63.patch",
+                "3007752 - Entity usage list does not scale": "patches/entity_usage-usage_list_scale-3007752-17.patch"
```
weitzman: I'm curious if there a reason why we are not contributing this and other patches?

docroot/modules/custom/mass_content_api/mass_content_api.links.task.yml
```diff
@@ -1,8 +1,9 @@
-mass_content_api.descendant_controller_linking_page:
-  route_name: mass_content_api.descendant_controller_linking_page
-  base_route: entity.node.canonical
-  title: 'Pages linking here'
-media_linking_page:
-  base_route: entity.media.canonical
-  route_name: mass_content_api.descendant_controller_media_linking_page
-  title: 'Pages linking here'
+# Pages linking here controller local task links.
```
weitzman, line 1: Do we intend to delete this commented out code? Uncomment later?

docroot/modules/custom/mass_content_api/src/DescendantStorage.php
```diff
@@ -138,6 +138,10 @@ public function getLinksTo(string $entity_type, int $entity_id): array {
     $query->condition('destination_id', $entity_id);
     $query->condition('relationship', self::LINKS_TO);
 
+    // Added for easier comparison to entity_usage.
+    $query->orderBy('source_type');
+    $query->orderBy('source_id', 'DESC');
```
weitzman: have we checked EXPLAIN to see if this operates on unindexed columns? Or maybe the result set is so small it doesnt matter?

docroot/modules/custom/mass_entity_usage/src/Controller/MassLocalTaskUsageController.php
```diff
@@ -0,0 +1,169 @@
...
+          $state_label,
+        ];
+      }
+    }
+
+    return $rows;
+  }
+
+  /**
+   * Recursive function to retrieve the parent node of a paragraph.
```
weitzman, line 143: This looks pretty useful in general. I cant think of a much better place for it though.

weitzman (changes requested):
Code looks good. I had just minor comments. It looks like there are no tests? I think we should test at least a vanilla workflow of this (create content with paras, entity refs, etc. assert that it is reported in the tab)

composer.json
weekbeforenext: That patch is contributed but in an MR. I just learned that the best practice for those is to copy the patch locally because someone else could change the MR in a way that could break your site.

docroot/modules/custom/mass_content_api/mass_content_api.links.task.yml
weekbeforenext: I can remove these links.

docroot/modules/custom/mass_content_api/src/DescendantStorage.php
weekbeforenext: This was added only to help when comparing the results of Descendant Manager with Entity Usage. If we get a thumb up for external review, we could technically remove those lines.

docroot/modules/custom/mass_entity_usage/src/Controller/MassLocalTaskUsageController.php
weekbeforenext: I copied and modified from some the original entity_usage code that was retrieving a link to the node for paragraph sources. I did something similar in docroot/modules/custom/mayflower/src/Helper.php, but wasn't sure if it was worth it to reference the same method.

composer.json
weitzman: that makes sense. Please add the URL to the comment which holds the patch. Or the URL to the MR.

docroot/modules/custom/mass_content_api/mass_content_api.links.task.yml
weekbeforenext: Uncommenting these because we want these tabs to be available for initial deployment.

docroot/modules/custom/mass_content_api/src/DescendantStorage.php
weekbeforenext: Removing these because comparison isn't easy for paragraph sources so it doesn't matter.

docroot/modules/custom/mass_entity_usage/src/Controller/MassLocalTaskUsageController.php
weekbeforenext: I think you mentioned on a call or somewhere that it's okay to leave this here as is. Let me know if I'm remember wrong and I can look again.

weekbeforenext:
> Code looks good. I had just minor comments. It looks like there are no tests? I think we should test at least a vanilla workflow of this (create content with paras, entity refs, etc. assert that it is reported in the tab)

I added tests to verify tracking is skipped properly and that the new "Pages linking here" page has the correct table caption, headings, and displays a usage record. Let me know if we need to add more.

weitzman (approved):

PR #759, opened 2021-05-18
Fix missing location listing hours

clairesunstudio:


clairesunstudio:
branch name change to avoid hotfix auto workflow, replaced by #760

clairesunstudio:
branch name change to avoid hotfix auto workflow, replaced by #760

PR #768, opened 2021-05-24
Ahoy supports native development environments

weitzman:
https://jira.mass.gov/browse/DP-21768

**To Test:**
- [ ] Setup a native dev env. Set the nev variable that’s in the PR. Note that you cant set it in the .env file. It has to be in the host OS (.bashrc or .profile files). Then try a few Ahoy commands.

---

### DIFF
--- a/native.ahoy.yml
+++ b/native.ahoy.yml
@@ -1,0 +1,32 @@
+ahoyapi: v2
+commands:
+  exec:
+    # These env variables come from https://github.com/wodby/docker4drupal/issues/273
+    cmd: bash -c "$@"
+  up:
+    # See https://laravel-guide.readthedocs.io/en/latest/valet/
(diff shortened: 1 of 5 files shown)


native.ahoy.yml
```diff
@@ -0,0 +1,32 @@
+ahoyapi: v2
+commands:
+  exec:
+    # These env variables come from https://github.com/wodby/docker4drupal/issues/273
+    cmd: bash -c "$@"
+  up:
+    # See https://laravel-guide.readthedocs.io/en/latest/valet/
+    cmd: valet up
```
tom-fleming: @weitzman I've installed Valet, but I'm not seeing the commands described in this file. What should these commands be?

weitzman: Fixed

PR #778, opened 2021-05-31
DP-22099: Add backstop job for post deployment checks

arthurbaghdas:
**Description:**
Configured Circle CI to have a  3rd parameter with default value.

https://jira.mass.gov/browse/DP-22099

**To Test:**

**Screenshots/GIFs:**

---


backstop/post-release.json
```diff
@@ -0,0 +1,22 @@
...
+  {"label": "InfoDetailsLanding", "url": "/info-details/qag-info-detail-with-landing-page-features"},
+  {"label": "NewsPressRelease", "url": "/news/qag-newspressrelease"},
+  {"label": "OrgQAGEOTSSGeneralOrg", "url": "/orgs/qag-executive-office-of-technology-services-and-security"},
+  {"label": "OrgElectedOfficial", "url": "/orgs/qag-test-elected-org-page"},
+  {"label": "RegulationEffectiveDate", "url": "/regulations/900-CMR-2-qag-regulation-title"},
+  {"label": "ServiceDetails", "url": "/service-details/qag-servicedetails"},
+  {"label": "Service1", "url": "/qag-service1"},
+  {"label": "ServiceGroupedLinks", "url": "/qag-service-page-with-grouped-links"},
+  {"label": "TopicPage1", "url": "/topics/qag-topicpage1"}
+]
```
weitzman: Github wants a new line here.

backstop/post-release.json
```diff
@@ -0,0 +1,22 @@
+[
+  {"label": "Document", "url": "/media/1268726"},
+  {"label": "HomepageWithAlert", "url":  "/", "showAlerts":  true},
```
weitzman, line 3: These 2 lines will be stale after a release so checking them doesn’t tell us much. We need to add them to the post-release purge if we want to backstop them https://github.com/massgov/openmass/blob/develop/drush/Commands/DeployCommands.php#L326-L330

weitzman (changes requested):

backstop/post-release.json
arthurbaghdas: Hey @weitzman , https://github.com/massgov/openmass/blob/develop/drush/Commands/DeployCommands.php#L340
This line seems to be adding homepage to the queue, correct me if I'm wrong. If that's the case I think i need to add another item to the array for media path.

weitzman: Yes you are correct. I didn’t notice that.

backstop/backstop.post-release.js
```diff
@@ -0,0 +1,97 @@
+// Determine the list of urls to use with backstop
```
weitzman: It looks like this file duplicates code in backstop.js. Since this is javascript, you can put shared code in a library (i.e. a shared file)

weitzman (changes requested):

weitzman:
Also, backstop should not have run on this PR. I wonder if we are introducing it into PR workflow by mistake.

arthurbaghdas:
I was just testing that it doesn't affect current workflows.

backstop/backstop.post-release.js
arthurbaghdas: Ok will take a look

arthurbaghdas:
Done @weitzman , ready for another review.

PR #784, opened 2021-06-07
DP-22225: Adding helper function to reduce duplicate code, adding sea…

serundeputy:
…rch, labels, to csv export api call.

**Description:**
* There were two separate files parsing query strings from drupal to send back to the feedback_api. I've abstracted those into a helper function so there is just one place to maintain the query processing. 
* Added the search and lables fields to the CSV export API call.

https://jira.mass.gov/browse/DP-22225

**To Test:**
* visit `/admin/ma-dash/feedback`
* perform a search that includes the labels field
* use the export link at the bottom of the search results
* form should export a text CSV file with the filtered results

---

### DIFF
--- a/changelogs/DP-22225.yml
+++ b/changelogs/DP-22225.yml
@@ -1,0 +1,41 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
+# #############################################################################
+# You can use one of the following types:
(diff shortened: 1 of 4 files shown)


adamcogbill:
@serundeputy, looked at this locally, and LGTM! I do see it failing that last circle test--any idea what that is?

PR #791, opened 2021-06-09
DP-22247 Suppress google tag scripts on admin pages.

oknate:
**Description:**
There's an error in the google_tag module where the description says not to add leading forward slash.  But this has changed in Drupal since the module has created.  The forward slash should be there.

https://jira.mass.gov/browse/DP-22247 

**To Test:**
- [ ] do a config import, notice the google tags are no longer loading on the admin pages

---


oknate:
this is superceded by upgrading google tag manager.

PR #794, opened 2021-06-09
DP-22004 - Return html on ajax call for alerts

ddiestra:
## Continues in #845 

**Description:**
- A new type of block was created for alerts to pull html directly and append it into the html instead of rendering with js.

https://jira.mass.gov/browse/DP-22004

**To Test:**
- [x] Replace mayflower ajax pattern block for alerts with the new Alert block

**Before Merge:**
- [x] Deploy new Cloudflare worker

**Screenshots/GIFs:**

---


tom-fleming:
@ddiestra These alerts blocks are definitely faster, especially the sitewide ones, which is great! I think one piece that is missing though is that this PR doesn't change the blocks in the layout to your blocks, which confused me at first. Could you include that configuration change to swap out these blocks? Thanks!

docroot/modules/custom/mass_alerts/mass_alerts.routing.yml
```diff
@@ -22,3 +22,19 @@ mass_alerts.user_autocomplete:
     _format: json
   requirements:
     _permission: 'access email address list for alerts config page'
+
+
+mass_alerts.site_alerts:
+  path: '/jsonapi/site-alerts'
```
tom-fleming: I'll be curious to see what others think about using /jsonapi as a base route. I would have probably set a route like /ma-alerts/site-alerts or something similar, but I think this is okay.

docroot/modules/custom/mass_alerts/src/Controller/ApiController.php
```diff
@@ -0,0 +1,229 @@
...
+      }
+
+      ksort($alerts);
+      $results['emergencyAlerts']['alerts'] = array_values($alerts);
+    }
+
+    $build = [
+      '#theme' => 'mass_alerts_sitewide',
+      '#emergencyAlerts' => $results['emergencyAlerts'],
+      '#cache' => ['max-age' => 0],
```
weitzman: We need full cache metadata before we merge this. Sitewide responses expire after 60 seconds and page/org responses get no expiration (they will expire after 30 min per cloudflare policy).

weitzman (changes requested):

conf/drupal/config/block.block.alertsblock_2.yml
```diff
@@ -0,0 +1,21 @@
...
+    - mass_theme
+id: alertsblock_2
+theme: mass_theme
+region: header
+weight: 0
+provider: null
+plugin: mass_alerts_block
+settings:
+  id: mass_alerts_block
+  label: 'Alerts block'
```
weitzman: Ideally these blocks have more descriptive name and label. Distinguish between sitewide and page-specific.

docroot/modules/custom/mass_alerts/mass_alerts.routing.yml
weitzman: Lets change this path. We are not using jsonapi nor returning JSON. I would suggest /alerts/sitewide and /alerts/page/[n]

docroot/modules/custom/mass_alerts/src/Controller/ApiController.php
```diff
@@ -0,0 +1,237 @@
...
+      '#cache' => [
+        'max-age' => Cache::PERMANENT,
+        'tags' => [
+          'node:' . $nid
+        ]
+      ],
+    ];
+
+    $output = $this->renderer->renderRoot($build);
+    $response = new Response();
```
weitzman: This needs to be \Drupal\Core\Render\HtmlResponse for both types of alerts. You can do ->setContent($build), and not call renderRoot().

docroot/modules/custom/mass_alerts/src/Plugin/Block/AlertsBlock.php
```diff
@@ -0,0 +1,111 @@
...
+
+  /**
+   * Drupal\Core\Routing\CurrentRouteMatch definition.
+   *
+   * @var \Drupal\Core\Routing\CurrentRouteMatch
+   */
+  protected $routeMatch;
+
+  /**
+   * CompanyBlock constructor.
```
weitzman: We are not a company :)

weitzman (changes requested):
- [x] I am getting a 500 error on the homepage with page specific request. Lets just not make this request for this page.
- [x] testEmergencyAlertAppearsInJson method needs to change to do assertions on the alerts HTML responses. It should also assert the 1 minute expiration and the cache tags headers. Ping me if its not clear how to do these things.
- [x] We have to fix current test failures

docroot/modules/custom/mass_alerts/mass_alerts.routing.yml
```diff
@@ -22,3 +22,19 @@ mass_alerts.user_autocomplete:
...
+  defaults:
+    _controller: 'Drupal\mass_alerts\Controller\ApiController::handleSiteRequest'
```
weitzman: I would call this AlertsController instead of ApiController.

docroot/modules/custom/mass_alerts/src/Controller/ApiController.php
weitzman: Actually, I think you do call renderRoot() and set the result to value of #markup.

docroot/modules/custom/mass_alerts/src/Controller/AlertsController.php
```diff
@@ -0,0 +1,253 @@
...
+      $results['headerAlerts'] = array_values($alerts);
+    }
+
+    $build = [
+      '#theme' => 'mass_alerts_page',
+      '#headerAlerts' => $results['headerAlerts'],
+      '#cache' => [
+        'max-age' => Cache::PERMANENT,
+        'tags' => [
+          'node:' . $nid,
```
weitzman: Not sure what $nid is. I think you can omit this line, and instead add: `$response->addCacheableDependency($node)`. Same for sitewide alerts.

docroot/modules/custom/mass_alerts/src/Controller/AlertsController.php
```diff
@@ -0,0 +1,253 @@
...
+          'node_list:alert'
```
weitzman:
```suggestion
          ''handy_cache_tags:node:alert';'
```

weitzman:
Lets remove AlertsJSONAPISubscriber.php from mass_alerts module as its no longer used.

docroot/modules/custom/mass_alerts/src/Controller/AlertsController.php
```diff
@@ -0,0 +1,261 @@
...
+        $item_id = $item->uuid() . '__' . $changed_date;
+        $timestamp = $item->get('field_emergency_alert_timestamp')->getString();
+        $unix_timestamp = strtotime($timestamp);
+        $timestamp = $this->dateFormatter->format($unix_timestamp, 'custom', 'M. jS, Y, h:i a');
+
+        $uri = $item->get('field_emergency_alert_link')->getString();
+
+        // For Test this could be empty.
+        if ($uri) {
+          $url = Url::fromUri($uri)->toString(TRUE)->getGeneratedUrl();
```
weitzman, line 105: Isn't getGeneratedUrl() being called on a string ?

docroot/modules/custom/mass_alerts/tests/src/ExistingSite/EmergencyAlertsTest.php
```diff
@@ -75,6 +75,98 @@ public function testEmergencyAlertAppearsInJson() {
...
+      ->execute();
+    $nodes = \Drupal::entityTypeManager()->getStorage('node')->loadMultiple($nids);
+    foreach ($nodes as $node) {
+      $node->moderation_state = MassModeration::UNPUBLISHED;
+      $node->save();
+    }
+
+    $related = $this->createNode([
+      'type' => 'service_page',
+      'title' => 'EmergencyAlertsClientSideTest Service Page',
```
weitzman: Not cllient side anymore

features/ct_alerts.feature
```diff
@@ -11,5 +11,3 @@ Feature: Alert Content type
   Scenario: Verify Alert JSON API
```
weitzman: Not JSON API anymore.

features/ct_alerts.feature
```diff
@@ -11,5 +11,3 @@ Feature: Alert Content type
...
     When I go to "/jsonapi/node/alert"
     Then the response status code should be 200
-    And the "node_list" cache tag should not be used
```
weitzman, line 14: `node_list` should not be appearing so lets not uncomment that. If its appearing we have a bug I think.

weitzman (changes requested):
Great job. Just some nits left. I will add the Cloudflare worker change.

docroot/modules/custom/mass_alerts/src/Controller/AlertsController.php
ddiestra: Yes and not exactly, if we don't do this way it will trigger some leaked metada on render

features/ct_alerts.feature
ddiestra: So this is on the JSON API,  because we remove the `AlertsJSONAPISubscriber.php`  then the node_list tag is not removed , do you want to instead restore this?

weitzman: Oh, its /jsonapi. I dont care about that anymore :). This is not an issue.

weitzman (approved):

conf/drupal/config/block.block.ajaxpattern.yml
```diff
@@ -1,6 +1,6 @@
 uuid: 3bfa4546-fcbf-4b3a-b468-cbcd60172bec
```
tom-fleming, line 1: @ddiestra Might be worth instead saving this for a cleanup ticket, but AFAICT we only use AjaxPatternBlock for these two alerts blocks, meaning we could probably safely remove this block, ajaxpattern_2, AjaxPatternBlock (see docroot/modules/custom/mayflower/src/Plugin/Block/AjaxPatternBlock.php), and its associated JS (docroot/modules/custom/mayflower/js/ajax-pattern-alerts.js), along with any other related code.

tom-fleming: I don't see keeping this here impacting anything on rollout though, so I'm good leaving it as-is for now

PR #795, opened 2021-06-09
DP-21655 Upgrade to PHP 7.4

weitzman:
Also updates to PHPUnit 7 and Drush 10.5.0

https://jira.mass.gov/browse/DP-21655

**To Test:**
- [ ] If its green its good.

---


docker-compose.yml
```diff
@@ -2,7 +2,8 @@ version: '3.1'
 
 services:
   drupal:
-    image: "massgov/drupal-container:1.0.17-dev"
+    #image: "massgov/drupal-container:php74-beta2-dev"
+    build: /Users/moshe.weitzman/reps/Drupal-Container/dev
```
oknate: Is this right, it looks like a local mac path?

oknate: should it be `image: "massgov/drupal-container:php74-beta2-dev"`?

oknate (changes requested):
docker-compose.yml needs update, see comment.

docker-compose.yml
weitzman: Good catch. Its now `massgov/drupal-container:1.0.19-dev` which contains today's bugfix for drupal-container. I will fix composer conflicts next.

weitzman:
I can confirm that https://github.com/massgov/Drupal-Container/pull/21 did its job and Xdebug defaults to Off again.

oknate (approved):
looks good now, local runs php 7.4; xdebug fix works, it's disabled by default now.

PR #796, opened 2021-06-10
dp-22082 Google Tag Manager module upgrade

vordude:
**Description:**
Switch to the D.O version of the Google Tag manager module. It will be ready for D9

https://jira.mass.gov/browse/DP-22082

**To Test:**
- [ ] GTM Embed Settings and Code looks right on the page.
- [ ] Module is ready for D9 according to the Upgrade Status page


weitzman:
Lets also disable the mikeyp/google_tag module in this PR.

vordude:
Bumped the lockfile, also did a composer remove on the mikeyp version of google_tag.

There will have to be a subsequent task to remove the patch from the current Drupal version of the module that won't install unless the mikeyp version's class is present.

weitzman (approved):
Code looks good.

PR #823, opened 2021-06-28
 DP-22215 - SPIKE make proof of concept for location listing view

ddiestra:
**Description:**
Add a new view for locations to replace the current locations page and also use leaflet map

https://jira.mass.gov/browse/DP-22215

**To Test:**
Go to any  organization, service, or location page and go to /locations.

Examples: 
- orgs/massachusetts-registry-of-motor-vehicles/locations
- locations/lowell-court-service-center/locations
- court-service-centers/locations

**Screenshots/GIFs:**

---


clairesunstudio:
@ddiestra  please update this branch and the associated mayflower package and deploy this to a feature environment

composer.json
```diff
@@ -106,9 +106,37 @@
                     "type": "zip"
                 }
             }
+        },
```
clairesunstudio: are these packages added intentionally?

clairesunstudio: ok these are added for handling ajax change --> 8a380f1

clairesunstudio:
Associated MF PR https://github.com/massgov/mayflower/pull/1430/files

clairesunstudio:
This branch is deployed to STAGE. But the location listing and the map are not rendered. 

Screen Shot 2021-10-14 at 3 04 16 PM

docroot/modules/custom/mass_map/mass_map.routing.yml
```diff
@@ -1,5 +1,5 @@
 mass_map.map_page:
-  path: '/node/{node}/locations'
+  path: '/node/{node}/locations-old'
```
tom-fleming, line 2: @joeg8612 Is this something we want to hang onto? This is the old map, which will still be accessible at locations-old. I would prefer that we just get rid of it.

PR #851, opened 2021-07-16
Fix tag invalidation for alerts page responses.

weitzman:
**Description:**
getValue() and getString() confusion broke tag invalidation. None of the conditions matched so no invalidation happens.

DP-22571

**To Test:**
- [ ] See Fionna bug reports in Teams last night and today (Friday). No 30 min wait needed since the issue was in Drupal and not varnish or cdn.

---

### DIFF
--- a/docroot/modules/custom/mass_alerts/mass_alerts.module
+++ b/docroot/modules/custom/mass_alerts/mass_alerts.module
@@ -32,7 +32,7 @@ const MASS_ALERTS_TAG_SITEWIDE = 'mass_alerts_sitewide';
  *   One of three options.
  */
 function mass_alerts_get_scope(NodeInterface $node) {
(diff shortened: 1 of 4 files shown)


docroot/modules/custom/mass_alerts/mass_alerts.module
```diff
@@ -32,7 +32,7 @@ const MASS_ALERTS_TAG_SITEWIDE = 'mass_alerts_sitewide';
  *   One of three options.
  */
 function mass_alerts_get_scope(NodeInterface $node) {
-  return $node->get("field_alert_display")->getValue();
+  return $node->get("field_alert_display")->getString();
```
weitzman, line 36: This is the actual bug fix. The rest is cleanup.

ddiestra (approved):

PR #856, opened 2021-07-20
DP-22538: Make paragraph for Board Members on org page

arthurbaghdas:
**Description:**

https://jira.mass.gov/browse/DP-22538

**To Test:**

**Screenshots/GIFs:**

---

### DIFF
--- a/docroot/modules/custom/mass_validation/mass_validation.module
+++ b/docroot/modules/custom/mass_validation/mass_validation.module
@@ -1191,6 +1191,9 @@ function mass_validation_form_node_org_page_form_alter(&$form, FormStateInterfac
  *   The form array passed by reference.
  */
 function mass_validation_org_page_alter_helper(array &$form) {
+  if (isset($form['field_boards'])) {
+    $form['field_boards']['#access'] = FALSE;
+  }
   $form['#attached']['library'][] = 'mass_validation/org_page_edit_form';
   $form['#validate'][] = '_mass_validation_form_node_org_page_form_validate';
 }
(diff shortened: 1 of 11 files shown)


docroot/modules/custom/mass_validation/mass_validation.module
```diff
@@ -1191,6 +1191,9 @@ function mass_validation_form_node_org_page_form_alter(&$form, FormStateInterfac
  *   The form array passed by reference.
  */
 function mass_validation_org_page_alter_helper(array &$form) {
+  if (isset($form['field_boards'])) {
```
weekbeforenext, line 1194: Would it be better to hide the field in the form display config since the field will eventually be removed?

docroot/sites/default/services.local.yml
```diff
@@ -1,7 +1,7 @@
 # Uncomment the lines below to enable debugging.
 # Do not commit the changes on this file.
-
-#twig.config:
-#  debug: true
-#  auto_reload: true
-#  cache: false
+#parameters:
```
weekbeforenext, line 3: This is nice! I've been looking for a spot to put this that will stick after running `ahoy comi`.

docroot/themes/custom/mass_theme/templates/paragraphs/paragraph--org-section-long-form--default.html.twig
```diff
@@ -1,14 +1,17 @@
-{% embed "@organisms/by-author/rich-text.twig" with {
-  "richText": {
-    "compHeading": {
-      "title": (content.field_section_long_form_heading|field_value)
-    }
-  }
+{% embed "@organisms/by-author/stacked-row-section.twig" with {
```
weekbeforenext, line 1: Glad you figured this out. It fixed all of the front-end issues I was seeing. ++

weekbeforenext (commented):
This looks good. One question noted and two compliments :)

docroot/modules/custom/mass_validation/mass_validation.module
arthurbaghdas: Yeah makes sense to me. I just decided to not deal with the configs for now, for the same reason, since we will be migrating it later. I can change this to work from the Form display, but we can keep this for now, since we will revisit it later during data migration work. @weekbeforenext

docroot/themes/custom/mass_theme/templates/paragraphs/paragraph--org-section-long-form--default.html.twig
arthurbaghdas: Thank you 😄

PR #864, opened 2021-07-21
DP-22263 implementing the custom multi image field formatter.

Javi-er:
implementing the custom multi image field formatter, setting up responsive images.

**Description:**
I found a partially complete implementation that I continued working on (mass_formatters module), there is a custom image field formatter for allowing to choose different image styles according to the breakpoints defined in openmass/docroot/themes/custom/mass_theme/mass_theme.breakpoints.yml .

I had to cleanup the paragraph preprocessing so it doesn't alter image output as well to format the data before passing it to mayflower on the paragraph template, since the field value at preprocess level is returning a single image, probably a bug with the custom formatter.

On the Mayflower template, a full picture object is rendered that it's absolutely positioned and object-fit set to cover so it has the same effect as the background image.

There is a corresponding mayflower pull request that's needed as well.

https://jira.mass.gov/browse/DP-22263

**To Test:**
- [ ] Go to: /qagcampaign-landing-with-image-key-message-header
- [ ] Make sure images are visible and that it uses a picture element.

**Screenshots/GIFs:**
Captura de Pantalla 2021-07-21 a la(s) 14 41 32

---


clairesunstudio:
mayflower testing branch DP-22263_implement-responsive-images-test deployed to feature2. 

See Joe's comment in the ticket

PR #881, opened 2021-07-30
DP-22653 Page Header Restructure

ddiestra:
**Description:**
Separate components to position the alerts block after the title

https://massgov.atlassian.net/browse/DP-22653


composer.json
```diff
@@ -214,7 +214,7 @@
         "enyo/dropzone": "5.7.2",
         "furf/jquery-ui-touch-punch": "dev-master",
         "geocoder-php/open-cage-provider": "^4.4",
-        "massgov/mayflower-artifacts": "11.11.0",
+        "massgov/mayflower-artifacts": "dev-DP-22653",
```
tom-fleming: This should be changed to match the latest Mayflower version in develop

PR #934, opened 2021-08-19
DP-22796 Invalidate alert responses based on old and new field values.

weitzman:
Add a foreach around the existing code (which is unchanged) so we do clears based on current field values and the prior field values. See Jira ticket for a case where that matters. 

DP-22796

**To Test:**
- [ ] If its green its good. We are adding more clears to what we had before.

---


ddiestra (approved):
LGTM

PR #943, opened 2021-08-22
DP-22808 Rollback mysql for now.

weitzman:
**Description:**
Get mysql build working again. We need to move to mysql 5.7 soon, but this tops bleeding.

DP-22808

  
**To Test:**
- [ ] Make sure tests pass.

---

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1021,7 +1021,7 @@ workflows:
           instance: massgov/mysql-sanitized:super
     triggers:
       - schedule:
-          cron: "42 15 * * *"
+          cron: "40 18 * * *"
           filters:
             branches:
               only:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -1021,11 +1021,12 @@ workflows:
           instance: massgov/mysql-sanitized:super
     triggers:
       - schedule:
-          cron: "42 15 * * *"
+          cron: "55 18 * * *"
```
tom-fleming, line 1024: @weitzman Is there a specific motivation for the cron change, or was this just for testing purposes?

weitzman:
Nope. It was needed to trigger a rebuild. We have always rebuilt this every
24 hours and never cared about when it kicks off exactly.

PR #983, opened 2021-09-03
DP-22657 Sending the number of items to the component.

Javi-er:
**Description:**
Pases the actual items to the action finder component so the top region is hidden if there aren't any.

DP-22657

**To Test:**
- [ ] Visit an organization page, there shouldn't be any visual differences on the "what would you like to do" paragraph.

---


Javi-er:
I'm merging this one since it's a pretty insignificant change.

PR #1030, opened 2021-09-23
Dp 22982 fix org section layout

clairesunstudio:
Mayflower PR https://github.com/massgov/mayflower/pull/1509


docroot/themes/custom/mass_theme/templates/paragraphs/paragraph--org-section-long-form--default.html.twig
```diff
@@ -1,22 +1,32 @@
...
-  {% set heading = {
-    "title": (content.field_section_long_form_heading|field_value),
-    "id": "",
-    "borderless": true
-  } %}
+  {% set title = content.field_section_long_form_heading|field_value %}
+{% endif %}
+
+{% set contained = false %}
+{% set paragraphType = content.field_section_long_form_content[0]["#paragraph"]["#values"]["type"] %}
```
Javi-er, line 7: @clairesunstudio the problem is that content.field_section_long_form_content contains 1 or more childs, field_section_long_form_content is each organization section sub paragraph, so by doing this: field_section_long_form_content[0] you are checking just for the first element, but if instead the second or third element is one of those, it will be ignored. In any case you need to add a wrapper to the subchilds, not the main wrapper (paragraph org section).

docroot/themes/custom/mass_theme/templates/paragraphs/paragraph--org-section-long-form--default.html.twig
```diff
@@ -1,22 +1,32 @@
...
   {% block stackedRowContentOverride %}
     {{ content.field_section_long_form_content }}
   {% endblock %}
 {% endembed %}
 
 {{ content.field_section_long_form_addition }}
+
+{{ dd(content.field_section_long_form_content[0]["#paragraph"]) }}
+{{ dd(content.field_section_long_form_content[0]["#paragraph"]["#values"]) }}
+{{ dd(content.field_section_long_form_content[0]["#paragraph"]["#values"]["type"]) }}
```
Javi-er, line 32: The problem with `dd()` is that it doesn't show which properties are protected, `["#values"]` is one of these and can't be accessed directly, instead you need to target each field or in this case if you want to know the type or bundle, you can do this:
`content.field_section_long_form_content[0]["#paragraph"].bundle()`

Javi-er:
Closing per discussion on DP-22982.

PR #1042, opened 2021-09-28
DP-22906: Create sitewide alert content type

arthurbaghdas:
**Description:**

DP-22906

**To Test:**

---


docroot/modules/custom/mass_alerts/mass_alerts.module
```diff
@@ -40,26 +40,36 @@ function mass_alerts_get_scope(NodeInterface $node) {
...
  */
 function mass_alerts_form_node_alert_form_alter(&$form, FormStateInterface $form_state, $form_id) {
-  $form['#validate'][] = '_mass_alerts_validate_site_wide_alert_creation';
   $form['#validate'][] = 'mass_alerts_validate_alert_placement';
   $form['#validate'][] = 'mass_alerts_validate_alert_content';
 
-  if (!\Drupal::currentUser()->hasPermission('create site wide alerts')) {
-    unset($form['field_alert_display']['widget']['#options']['site_wide']);
-  }
+  // @TODO Remove the code, since we are using new Sitewide Alerts content type.
```
tom-fleming, line 46: @arthurbaghdas Is there a reason we're not acting on these todo items now?

arthurbaghdas: Yeah, this is the deployment problem, since we need to migrate data from the field before deleting that option. SO the best and safest way is to delete it after deploy.

PR #1046, opened 2021-09-29
DP-22983 Adding new variable for wrapping call out links in some situations.

Javi-er:
**Description:**
Ads an optional wrapper  on the OpenMass call out links template. Then on the call out links paragraph preprocess, it sets this optional wrapper if it's on an Organization page. Also cleaned up some small things from previous tasks.

DP-22983

**To Test:**
Refer to the  Jira ticket.

---

### DIFF
--- a/docroot/themes/custom/mass_theme/templates/paragraphs/paragraph--what-would-you-like-to-do.html.twig
+++ b/docroot/themes/custom/mass_theme/templates/paragraphs/paragraph--what-would-you-like-to-do.html.twig
@@ -46,7 +46,6 @@ solution will need a refactor. This ID is currently being used as an
 anchor link in field--extra-org-nav-featured-services.html.twig. #}
     {% embed "@organisms/by-author/action-finder.twig" with {
       "actionFinder": {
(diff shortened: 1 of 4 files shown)


docroot/themes/custom/mass_theme/templates/paragraphs/paragraph--what-would-you-like-to-do.html.twig
```diff
@@ -46,7 +46,6 @@ solution will need a refactor. This ID is currently being used as an
 anchor link in field--extra-org-nav-featured-services.html.twig. #}
     {% embed "@organisms/by-author/action-finder.twig" with {
       "actionFinder": {
-        "id": "what-would-you-like-to-do",
```
weekbeforenext: I think removing this will break a link in the SubNav bar based on the comments above the embed code.

Javi-er: yes, you are right, just added that back. At some point this should be refactored so the id is unique if multiple paragraphs are added.

PR #1057, opened 2021-10-07
DP-22210 fix focus issue

clairesunstudio:


clairesunstudio:
@ddiestra  is this associated with the mayflower PR in the ticket? If so, please update this branch and the associated mayflower package and deploy this to a feature environment

tom-fleming (commented):
This looks good to me, and it is already merged into DP-22215

PR #1060, opened 2021-10-08
DP-23009 - Elimate false positivies

ddiestra:
**Description:**
Added a wait for sitewide alerts

[DP-23009](https://massgov.atlassian.net/browse/DP-23009)


backstop/scripts/ready.js
```diff
@@ -178,6 +178,10 @@ module.exports = async function(page, scenario, vp) {
     // in local environments.
     await page.waitForFunction('jQuery.active == 0');
 
+    // Wait for alerts to load
+    await page.waitForFunction("document.querySelector('.ma__emergency-alerts__content') !== null");
```
arthurbaghdas: See this part of the file https://github.com/massgov/openmass/blob/ab30616780a537e19b05278774ba269c97fb5a66/backstop/scripts/ready.js#L185-L195
Alerts should only show on the pages where showAlerts flag is set to true in the json file. And there is already a place where we set default Time/date for the alerts to avoid false positives.

arthurbaghdas (changes requested):

backstop/scripts/ready.js
beto-aveiga: Sure @arthurbaghdas I didn't write that, but based on your explanation makes sense to not preserve it.
I will remove that addition from this PR.

beto-aveiga: I will run backstop tests again to see what keeps failing.

PR #1066, opened 2021-10-13
DP-23066 Reduce 500 errors

weitzman:
**Description:**
- Fix getLangCode bug in mass_ultility
- Update drupal/username_enumeration_prevention  to latest to fix a D9 bug
- The redirect loops are very infrequent (a few a week) and unclear how to resolve so lets leave them for now. Also hoping they might stop happenning with the above langcode fix
- The `ports` change in docker gets us back a feature we had before where local DB tools can be used.

https://massgov.atlassian.net/browse/DP-23066

**To Test:**
- [ ] If its green its good

---


weekbeforenext (approved):
Looks good to me.

PR #1081, opened 2021-10-18
Dp 3188 a11y autocomplete for search

beto-aveiga:
**Description:**
Autocomplete for search working as described here.
https://rawgit.com/w3c/aria-practices/master/aria-practices-DeletedSectionsArchive.html#autocomplete

Except for this part, because Google CSE needs a letter to show suggestions (one option would be to have a default list of suggestions, although the ticket does not specify what to do in this case).
> With focus in an empty text box, press Down Arrow, Up Arrow, Alt + Down Arrow, or Alt + Up Arrow to display the entire list of choices. Focus remains in the text box and no choice is highlighted.

DP-3188

**To Test (applicable instructions from [this link](https://rawgit.com/w3c/aria-practices/master/aria-practices-DeletedSectionsArchive.html#autocomplete)) :**
- With the drop-down list of choices displayed, move the mouse pointer over an item in the list to highlight it. The text box value is not modified when the mouse is used to highlight a choice. Clicking on the highlighted choice will close the drop-down and update the text box with the selected choice.
- With focus in an empty text box, type any letter. If any of the available choices begin with the letter typed, those choices are displayed in a drop down. If the letter typed does not match any of the available choices the drop-down list is not displayed.
- With focus in text box with an existing value type additional letters. As the user types letters the list of choices is filtered so that only those that begin with the typed letters are displayed.
  - Until the user presses the arrow keys to highlight a particular choice, only the typed letters are displayed in the text box.
  - In an editable auto-complete, if no choices match the letter(s) typed, the drop down list closes.
  - In a non-editable auto-complete, any letters that do not result in a match from the list are ignored. The drop down list of choices remains static until the user presses:
    - Escape to clear the text field
    - Backspace to remove some of the letters previously typed
    - An additional letter that results in a valid list of choices.
   - Navigation through the list of choices and display of the highlighted choice in the text box works as described above.
- With focus in a text box, press Escape
  - If there is no text in the text box, pressing Escape closes the drop-down if it is displayed.
  - For an editable autocomplete that has text in the text box that was both typed by the user and auto-completed by highlighting a choice using the keyboard, the auto-completed portion of the text is cleared and the user typed characters remain in the text box. The drop-down list is closed. To completely clear the text box contents the user must use the backspace key to remove the typed characters. This is how the Google search box in the Firefox UI works. Recommend that pressing the Escape key again completely clears the text box rather than relying on only the backspace key.
  - For a non-editable auto-complete that has text in the text box that was both typed by the user and auto-completed by highlighting a choice using the keyboard, pressing Escape closes the drop-down list and leaves the current choice in the text box.
  - For an editable or non-editable auto complete with text in the text box that was typed by the user and the mouse is highlighting a choice in the drop down (keyboard navigation was NOT used), pressing Escape closes the drop down and leaves the typed text displayed in the text box. Need to consider if pressing Escape again should clear the typed text. The user must press the Down arrow or Alt + Down arrow or click the associated icon to invoke the drop-down list of choices again. 

---

### DIFF
--- a/docroot/modules/custom/mass_search/js/mass_search.google_cse.js
+++ b/docroot/modules/custom/mass_search/js/mass_search.google_cse.js
@@ -136,7 +136,7 @@
     $input.trigger('autocomplete:suggestionsUpdated', [results]);
   });
   $input.on('keyup', function (event) {
-    if (event.keyCode !== 13 && event.keyCode !== 38 && event.keyCode !== 40) {
+    if (event.key !== 'Enter' && event.key !== 'ArrowUp' && event.key !== 'ArrowDown') {
       ac.update(event.target.value);
     }
   });
@@ -146,7 +146,7 @@
     var scope = input.parents('.js-suggestions-container');
 
     $('[role="listbox"]', scope).remove();
-    $('[data-suggestions]', scope).html('<div role="listbox" id="suggestions-list"></div>');  
+    $('[data-suggestions]', scope).html('<div role="listbox" id="suggestions-list"></div>');
     $('.js-suggestions-help', scope).empty();
 
     var $listbox = $('[role="listbox"]', scope);
@@ -172,7 +172,7 @@
 
     $listbox.on('keydown', function (e) {
 
-      if (e.keyCode === 13) {
+      if (e.key === 'Enter') {
         e.preventDefault();
         e.stopPropagation();
         input.data('realValue', input.val());
@@ -181,7 +181,7 @@
         return;
       }
 
-      if (e.keyCode === 27) {
+      if (e.key === 'Esc') {
         input.val(input.data('realValue'));
         input.focus();
         $listbox.hide();
@@ -189,7 +189,7 @@
       }
 
       var newOption;
-      if (e.keyCode === 40) {
+      if (e.key === 'ArrowDown') {
         e.preventDefault();
         newOption = $('.selected', scope).next();
 
@@ -201,7 +201,7 @@
         }
       }
 
-      if (e.keyCode === 38) {
+      if (e.key === 'ArrowUp') {
         e.preventDefault();
         newOption = $('.selected', scope).prev();
 
@@ -245,7 +245,7 @@
       // hence we need to update the reference.
       $listbox = $('[role="listbox"]', scope);
 
-      if (e.keyCode === 27) {
+      if (e.key === 'Escape') {
 
         if (input.val() !== '' && ($listbox.length === 0 || !$listbox.is(':visible'))) {
           input.val('');
@@ -258,7 +258,7 @@
         return;
       }
 
-      if (e.keyCode === 38) {
+      if (e.key === 'ArrowUp') {
         e.preventDefault();
         $listbox.find('.selected', scope).removeClass('selected');
         $listbox.attr('tabindex', '0').focus();
@@ -267,7 +267,7 @@
         return;
       }
 
-      if (e.keyCode === 40) {
+      if (e.key === 'ArrowDown') {
         e.preventDefault();
         $listbox.find('.selected', scope).removeClass('selected');
 
@@ -285,7 +285,7 @@
         return;
       }
 
-      if (e.keyCode === 9) {
+      if (e.key === 'Tab') {
         $listbox.remove();
         return;
       }
--- a/docroot/modules/custom/mass_search/js/mass_search.google_cse.js
+++ b/docroot/modules/custom/mass_search/js/mass_search.google_cse.js
@@ -222,7 +222,6 @@
     });
 
     $('[role="option"]', scope).on('click', function () {
-      console.log('clicking');
       input.data('realValue', $(this).text());
       $('[data-suggest]', scope).val($(this).text())
         .focus();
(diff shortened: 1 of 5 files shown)


docroot/modules/custom/mass_search/js/mass_search.google_cse.js
```diff
@@ -167,57 +170,128 @@
...
-        input.val($('.selected').text());
-      }
-      if (e.keyCode === 9) {
-        $('[role="listbox"]', scope).remove();
-      }
-    });
+    $listbox.on('keydown', function (e) {
 
-    $('[role="listbox"]', scope).on('keydown', function (e) {
       if (e.keyCode === 13) {
```
ygannett: I know this is an existing part, but `keyCode` is deprecated:  https://caniuse.com/?search=keycode
It's working so far, but for future proof, can you replace all `keyCode` with `key` along with their values in the file?

ygannett:
In IE11, no options are displayed.  I assume the feature doesn’t support IE 11.

JAWS in FF and Voiceover in FF don’t tell the options available. 
Voiceover with Chrome and Firefox don't read `div.s-suggestions-help`.

Can you make change to be like below to see if user experience with screen readers mentioned above is improved?
- `aria-live` value
- `aria-autocomplete`
- `role`
- `aria-own`
-  `id`

To make the requested changes obvious, I added empty lines around them.  Let me know for any questions and concern.
```
<div class="visually-hidden js-suggestions-help" role="status"

  aria-live="assertive"><!-- Change it from "polite." -->

  There are 10 suggestions. Use the up and down arrows to browse.</div>
<label for="banner-search" class="ma__search-banner__label">Search terms</label>
<input id="banner-search" name="q" class="ma__header-search__input" autocomplete="off"

  aria-autocomplete="both"
  role="combobox"
  aria-own="suggestions-list"

  placeholder="Search Mass.gov" data-suggest="" type="text">
<div class="ma__suggestions" data-suggestions="">
  <div role="listbox"

    id="suggestions-list"

    tabindex="0" aria-activedescendant="banner-search-6" style="display: none;">
    <div role="option" tabindex="-1" id="banner-search-0" class="selected">water management act program</div>
    <div role="option" tabindex="-1" id="banner-search-1" class="">water resources commission</div>
    ...
  </div>
</div>
```

beto-aveiga:
Thanks for the review, @ygannett. Also, thanks for the detailed implementation details (when possible).

I'm not sure if IE11 is supported. If the autocomplete is not working at https://mass.gov on IE11, then it is for sure another issue.

I will add the features you are mentioning to the ticket and implement them as well.
Thanks!

ygannett (commented):
Thank you for making the change.  JAWS and Voiceover in Firefox mention combobox, so their users would know some options are available to explore. Voiceover announces `div.s-suggestions-help` in Chrome and Firefox.

I don't see the change in Home page.  It seems this is because the home page is independently set up in theming and its search is also hard-coded in the template, I believe.  Please let me know if you have a trouble finding the template.
home

docroot/modules/custom/mass_search/js/mass_search.google_cse.js
ygannett: I'm afraid that I was not clear about above comment.  I'm sorry about that.  What I meant was to change `keyCode` to `key`, so it would look like `e.key === "ArrowDown"` instead of `e.keyCode === "40"` for example.

beto-aveiga: Sure, @ygannett !

I just updated the script, replaced `keyCode` with `key`. There is some discussion about how to handle this on different browsers, which is why I preferred not to touch it. However, now that you mentioned it, I think that using `key` is our best option.

Last time I forgot to submit the template to change from polite to assertive... that was just me being me 😄

Same URL, remember to clear your browser's cache to see the changes: https://pr1081-w5wnwne5t8wlimwu1dush8dijsireres.tugboat.qa/

ygannett: Thank you for updating the key code.

For the home page search component, I don't see the change. You did it right for that part to the change search for the rest of the pages in the site.  You originally did include the theming template where the search is hard coded, not using the Mayflower template in the PR.  

So, I looked into which template is used for the home page. 
home
 It turned out the home page calls the Mayflower template for the search through a theming template.
`{% include '@organisms/by-author/search-banner.twig' %}`
home-search

I'm sorry that I was wrong about the home page search is hard-coded in the theming.  To reflect the changes, `organisms/by-author/search-banner.twig` in Mayflower needs some edits.  Let me know if you have any questions.

beto-aveiga: @ygannett 
I think it was my fault.
I just did an update because I forgot the other options, that is why you are not seeing them.
It will be ready on Tugboat in just 5 minutes for you to check.
I think you are right about the template being hardcoded.
I'm not sure where that other Mayflower template is used, or if is outdated.

ygannett (approved):
After the quick conversation with @beto-aveiga, confirmed the change.  The change was not shown on Tugboat when I tested earlier, but it is now. No MF work necessary.

docroot/modules/custom/mass_search/js/mass_search.google_cse.js
ygannett: Looks good!  I'll merge the PR once Circle CI job is done with the updating the branch with develop.

PR #1090, opened 2021-10-21
Dp 23210 fix backstop failures with org boards

beto-aveiga:
**Description:**
- Restore OrgBoards on BS tests
- Increase timeout on OrgBoards

DP-23210

**To Test:**
- [ ] Run backstop tests and ensure that the OrgBoards option in the backstop/all.json does not generate a timeout failure.

---


changelogs/DP-23210.yml
```diff
@@ -0,0 +1,3 @@
+Added:
+  - description: Restore OrgBoards & increase timeout
```
tom-fleming:
```suggestion
  - description: Restore OrgBoards
```

PR #1098, opened 2021-10-25
Mayflower/dp 23257 mayflower version 11.16.3

ygannett:
Changed:
  - description:  |-
        Updated Mayflower version to 11.16.3.
          - Update BrandBanner toggle button text (MF #1535)
    issue: DP-23257


ygannett:
Backstop failures are from adding the brand banner.

PR #1102, opened 2021-10-26
Disable and remove handy cache tags module in favor of core's bundle list tags.

weitzman:
- background: https://www.drupal.org/node/3107058
- handy_cache_tags has no .install file so it is safe to remove from composer and disable in same PR.

DP-22423

**To Test:**
- [ ] If its green its good

---


beto-aveiga:
@weitzman I saw these references for `handy_cache_tags` in the repo. Should we care about them?

image

weitzman:
I believe those got removed last night in a different PR

PR #1104, opened 2021-10-27
DP-23070: Fix error messages in Drupal

arthurbaghdas:
**Description:**

DP-23070

**To Test:**

---


docroot/themes/custom/mass_admin_theme/mass_admin_theme.theme
```diff
@@ -71,7 +71,7 @@ function mass_admin_theme_preprocess_file_upload_help(&$variables) {
  * @param array $variables
  *   Contains the form element data from $element['entities'].
  */
-function mass_admin_theme_inline_entity_form_entity_table(array $variables) {
+function mass_admin_preprocess_inline_entity_form_entity_table(array $variables) {
```
weekbeforenext: I'm concerned that this change is just disabling the function since the hook replacement should be the theme name 'mass_admin_theme'.

I pulled locally and dropped a breakpoint in to test using the Download field for an Executive Order. It didn't fire with the change, but did fire when I reverted the change. 

I would say that we should probably evaluate if this code is still necessary or dig deeper to determine why this function triggers the error.

arthurbaghdas: You are right, @weekbeforenext , since this is working fine, I think we can just remove this snippet. The only difference between the contrib function is here: https://github.com/massgov/openmass/pull/1104/files#diff-b17a7dcaf861f7f1970d9697b71e47ec4eda74bdeeecd607987fade0d3235204L122-L125  But I checked it is working fine without this code.

weekbeforenext (approved):
Looks good to me. Needs updates from develop before merge.

PR #1109, opened 2021-10-27
Dp 23065 change header on service page

beto-aveiga:
**Description:**
For service page header:
- Makes the Image optional.
- If there is no image, a blue background is used with full width and height becomes fluid.
- H1 and related style changes.
- Mobile display of header uses full width.

image

DP-23065

**To Test:**
- [ ] Edit an existing service with an image.
- [ ] Remove the image
- [ ] Save
- [ ] Verify changes on the description above.

---


arthurbaghdas (approved):
Drupal changes look good to me. I don't remember how we handle mayflower part here. Should we change mayflower version back to dev-develop before merging?
Can you get the answer to that question @beto-aveiga and fix the conflicts?

beto-aveiga:
@arthurbaghdas I'm fixing the conflicts with this branch. Thanks!
> Should we change Mayflower version back to dev-develop before merging?

Yes, but first the Mayflower branch has to be merged, AFAIK.

docroot/themes/custom/mass_theme/templates/content/node--service-page.html.twig
```diff
@@ -58,13 +58,13 @@
     {% endblock %}
   {% endembed %}
 
-  {% set bannerImageWide = file_url(node.field_service_bg_wide.entity.fileuri ?? '') %}
```
clairesunstudio, line 61: might be related to this line? @beto-aveiga

PR #1114, opened 2021-11-01
Dp 23222 modify front end to get alerts to load faster

beto-aveiga:
**Description:**
- Calls `alerts` function on `alerts.js` as soon as possible.
- Library `mass_alerts.behavior` part of the critical rendering part, included now in head.

image

DP-23222

**To Test:**
- [ ] Check requests timing for alert blocks using the developer tools in your browser (see image)

---


docroot/modules/custom/mass_alerts/src/Plugin/Block/AlertsBlock.php
```diff
@@ -88,21 +89,32 @@ public function build() {
...
+    // Needs to be cached, as rendering a block happens once.
+    // @see docroot/themes/custom/mass_theme/templates/layout/html.html.twig */
+    // @see mass_theme_preprocess_html()
+    $current_path = \Drupal::service('path.current')->getPath();
+    $cid = 'mass_alerts_blocks:' . $current_path;
+    $cache_entry = \Drupal::cache('render')->get($cid);
+    $data = $cache_entry->data ?? new stdClass();
+    $data->$path = ['#path' => $path];
+    \Drupal::cache('render')->set($cid, $data);
+
```
weitzman, line 113: Can you explain more about what we are doing here? Why its needed and why we always set the cache, why we change ->path and re-save it?

docroot/themes/custom/mass_theme/mass_theme.theme
```diff
@@ -4377,6 +4377,59 @@ function mass_theme_preprocess_field__person_org_role(&$variables) {
...
+  }
+}
+
+/**
+ * Data to fetch the alerts as soon as possible.
+ *
+ * @see docroot/themes/custom/mass_theme/templates/layout/html.html.twig
+ * @see Drupal\mass_alerts\Plugin\Block:build()
+ */
+function mass_theme_fetch_alerts_sooner(&$variables) {
```
weitzman: Lets rename this. This gets $variables, it doesnt do the fetch.

docroot/modules/custom/mass_alerts/templates/mass-alerts-block.html.twig
```diff
@@ -1,4 +1,5 @@
 {% if path %}
   {{ attach_library('mass_alerts/behaviors') }}
-  <div class="mass-alerts-block" data-alerts-path="{{ path }}"></div>
+  <div class="mass-alerts-block"></div>
+  <script type="text/javascript">alerts("{{ path }}", "{{ node_type }}", document.currentScript.previousElementSibling)</script>
```
weitzman, line 4: is there a reason why the data attribute approach will no longer work?

weitzman (changes requested):
It looks like the library is still going into aggregated js. How about we inline it so its available in the initial page load? In general, this PR looks more complex than what I am expecting. It seems like we need to fetch some HTML, and place it into an empty div. Maybe someone has ideas on how to get there?

docroot/modules/custom/mass_alerts/templates/mass-alerts-block.html.twig
beto-aveiga: Because the alerts function was already receiving parameters, hence we can send another parameter instead of looking inside the alerts script and trying to determine the path based on an external block.

docroot/themes/custom/mass_theme/mass_theme.theme
beto-aveiga: Sure! Will do.

docroot/modules/custom/mass_alerts/src/Plugin/Block/AlertsBlock.php
beto-aveiga: There are 2 blocks to set alerts for a few pages, for the rest, there is 1 block.
So we run this build process, for some pages, up to 2 times, once per alerts-block placed on the page.
However, the CID is built on the $current_path, which means the same CID can hold information for multiple alert blocks.
That is why I need to get the data first, to not override cached data from other block builds.
I use $data->$path to not override other paths and later process that data to make early requests on html.html.twig.
It is redundant to set a property with name $path and then set the '#path' with the $path value, but it is very easy to iterate that on JS and differentiate multiple blocks at the same time.

beto-aveiga (commented):
@weitzman I just answered your comments. I will push a new commit now, renaming a function.

tom-fleming:
Since we also now have PR #1171, I'ld like to close this one to avoid confusion.

PR #1117, opened 2021-11-01
Dp 23300 permission change for topic pages

beto-aveiga:
**Description:**
- Author & Editor role can edit any Topic Page
- Users that can Edit Any Topic Page but can't create a Topic Page can use the following fields when the field_restrict_link_management is not checked:
  - Intended audience
  - Organizations
  - Labels
  - Link groups
  - Workflow states.
  - Action (save/preview)
  
DP-23300

**To Test:**
- [ ] With author and editor role (or roles able to edit topic pages, but not create topic pages)
  - [ ] Edit any topic page, you should be able to edit only the fields mentioned above. 
- [ ] With an administrator (or any other role that can create topic pages) check the "Restrict link management to internal content team".
- [ ] With author and editor role
  - [ ] You shouldn't be able to edit any field. 

---


beto-aveiga:
Declining this PR in favor of https://github.com/massgov/openmass/pull/1186

PR #1135, opened 2021-11-05
Use ddev db snapshot instead of db container

weitzman:
DP-22872

**To Test:**

---


weitzman:
We are sticking with docker container with data.

PR #1143, opened 2021-11-09
DP-23215 Org page component positioning

ygannett:
**Description:**

- Add the stat paragraph to the org page preprocess function, mass_theme_preprocess_paragraph__org_section_long_form(), in mass_theme.theme.

DP-23215

**To Test:**
- [ ] Add stats in any org page.
- [ ] Test the stats are aligned vertically with other components within the page/column width.

Screen width 911 px +

Screen width 780px +

Screen width 620px

---


docroot/themes/custom/mass_theme/overrides/css/stacked-row-sections.css
```diff
@@ -0,0 +1,24 @@
+/**
+ * Align stat components in stacked row section (ex. org page) with other components.
+ */
+ .ma__stacked-row__section__stat-wrapper {
```
clairesunstudio: Instead of adding an additional override class, I wonder if we can leverage the existing classes that I think @Javi-er added for aligning content in openmass `om__contained-component`

docroot/themes/custom/mass_theme/templates/paragraphs/paragraph--org-section-long-form--default.html.twig
```diff
@@ -16,6 +16,14 @@
   "externalSidebar": true
 } %}
   {% block stackedRowContentOverride %}
+    {# Add a wrapper for stat component to align with other components. #}
+    {% set statComponent = "" %}
+    {% for org_component in org_components %}
+      {% set statComponent = org_component[0].field_stat ? "true" : "false" %}
```
clairesunstudio: This check is not working as intended, this class is being added to all rows
Screen Shot 2021-11-22 at 5 03 12 PM

@Javi-er have done similar fixes before, I will @ him on the ticket to chat about the solution.

clairesunstudio:
Please see comments on the ticket from external review.  I think we need to adjust the spacing on the stat for it to be able to inline with rich text
Screen Shot 2021-11-22 at 4 46 10 PM

Javi-er (approved):
I added a small commit for preventing those two "notices", rest looks good!

PR #1148, opened 2021-11-10
Dp 23353 allow all authors to see image header field for info details page

beto-aveiga:
**Description:**
Avoids field_banner_image to be editable or not based on the field_info_details_sections flag, on an info_details form.

DP-23353

---


weitzman (approved):
Approved. Not merging because this currently is marked as Draft.

PR #1149, opened 2021-11-10
DP-23335: Testing automatically assigning parents / child relationships

weekbeforenext:
**Description:**
Adding a deploy hook to assign parents to content from a custom table.

DP-23335

**To Test:**

---


weekbeforenext:
It looks like another feature branch that used this branch as a base was merged into the integration branch. This will cause the Danger checks to fail, because the changelog file is already merged into the integration branch.

PR #1171, opened 2021-11-18
Dp 23222 modify front end to get alerts to load faster  simplified

beto-aveiga:
**Description:**

DP-23222

**To Test:**

---


docroot/themes/custom/mass_theme/mass_theme.theme
```diff
@@ -4377,6 +4377,24 @@ function mass_theme_preprocess_field__person_org_role(&$variables) {
...
+ * Data to fetch the alerts as soon as possible.
+ *
+ * @see docroot/themes/custom/mass_theme/templates/layout/html.html.twig
+ */
+function mass_theme_build_data_from_alert_blocks(&$variables) {
+  $is_front_page = \Drupal::service('path.matcher')->isFrontPage();
+  /** @var Drupal\Core\Routing\CurrentRouteMatch */
+  $current_route_match_service = Drupal::service('current_route_match');
+  /** @var \Drupal\node\Entity\Node */
+  $node = $is_front_page ?: $current_route_match_service->getParameter('node');
```
weitzman: It looks $node will be true when we are  on the front-page. That’s unexpected to me.

weitzman (commented):
Thanks. I'd like for Tom to review this PR and the Mayflower PR. Perhaps he wil have some suggestions.

docroot/themes/custom/mass_theme/mass_theme.theme
beto-aveiga: I did this because we don't need to load the node if we are on the front page. But I agree, it is confusing. I will update that specific line. Thanks!

docroot/themes/custom/mass_theme/mass_theme.theme
```diff
@@ -4384,6 +4384,24 @@ function mass_theme_preprocess_field__person_org_role(&$variables) {
...
+function mass_theme_build_data_from_alert_blocks(&$variables) {
+  $is_front_page = \Drupal::service('path.matcher')->isFrontPage();
+  /** @var Drupal\Core\Routing\CurrentRouteMatch */
+  $current_route_match_service = Drupal::service('current_route_match');
+  /** @var \Drupal\node\Entity\Node */
+  $node = $current_route_match_service->getParameter('node');
+  // Defining paths to fetch alerts.
+  $paths = [];
+  $paths[] = ['#path' => '/alerts/sitewide'];
+  !$is_front_page ? $paths[] = ['#path' => '/alerts/page/' . $node->id()] : NULL;
```
tom-fleming: @beto-aveiga I think this might be a problem for non-node pages, which will not have a node parameter to match.

PR #1176, opened 2021-11-19
DP-23302: Prevent unpublish of a page with published children

weekbeforenext:
**Description:**
Added validation to nodes to check for parents with similar logic as deleting a parent node. It checks for unpublished or trash moderation state and checks for children. If children are found, it adds a violation.

DP-23302

**To Test:**
- Login and go to /admin/content?type_1=org_page.
- Edit one or more organizations to assign one other org as the “Parent page” value.
- Edit the parent organization.
- Change the status to “Unpublished” and save.
- Verify you get a form error that says you can’t unpublish the node because it has children with a list of the child titles.
- Change the status to “Trash” and save.
- Verify you get the same message.
- Change the status to “Published” and save.
- Verify the node saves without error.
- Edit an organization without children.
- Change the status to “Unpublished” and save.
- Verify the node saves without error.
- Change the status to “Trash” and save.
- Verify the node saves without error.

---

### DIFF
--- a/docroot/modules/custom/mass_validation/src/Plugin/Validation/Constraint/UnpublishParentConstraintValidator.php
+++ b/docroot/modules/custom/mass_validation/src/Plugin/Validation/Constraint/UnpublishParentConstraintValidator.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace Drupal\mass_validation\Plugin\Validation\Constraint;
+
+use Drupal\Core\StringTranslation\PluralTranslatableMarkup;
+use Drupal\entity_hierarchy\Information\ChildEntityWarningBuilder;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+
+/**
+ * Checks if an entity has a parent before unpublishing.
+ */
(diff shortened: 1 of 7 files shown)


docroot/modules/custom/mass_validation/src/Plugin/Validation/Constraint/UnpublishParentConstraintValidator.php
```diff
@@ -0,0 +1,44 @@
...
+   * {@inheritdoc}
+   */
+  public function validate($entity, Constraint $constraint) {
+    if (!isset($entity)) {
+      return;
+    }
+
+    $moderation_state = $entity->get('moderation_state')->value;
+    $unpublished_states = ['unpublished', 'trash'];
+    if (in_array($moderation_state, $unpublished_states)) {
```
beto-aveiga, line 25: I prefer early returns, and this line and the line below would be a perfect fit.
Anyhow, the code is easy to read. Good job!

weekbeforenext: Thanks :)

PR #1177, opened 2021-11-22
Change base DB image

weitzman:
**Description:**

- Use a minimally customized mariaDB image instead of our customized Percona based DB container. This mariaDB image is newer so does not give error on Drupal requirements page (i.e. status report). The only customization is commenting out the VOLUME instruction
- Build database containers for Arm in addition to our usual architecture. This lets our DB container run natively on M1 Macs. See README change for the file you have to uncomment. These containers are not yet tested by "should work". @weekbeforenext will test them, or I will do so when my Mac arrives in early Dec.

DP-23172

**To Test:**
- [ ] Review that [all populate jobs are passing](https://app.circleci.com/pipelines/github/massgov/openmass/14607/workflows/2382d141-3f6b-402c-b23b-9f2963314712)
- [ ] Run `ddev drush rq` before and after an `ahoy pull`. Verify that the there is no db error reported after the pull.
- [ ] Run `drush status` and assure that bootstrap succeeded. Request the homepage and assure that it returns as expected.

---


weekbeforenext (commented):
Trying on my M1 local, the database system version looks right. Looking at the containers, ddev-mass-backstop and ddev-mass-chromedriver are still amd64. This may be intended. Otherwise, the site looks good on my M1 (yay!)

Trying this on my non-M1 local, the database system version error didn't change after `ddev start` and `ahoy pull`. Should this one also be fixed?

weitzman:
I think what's happening is that develop overwrote what we pushed for
latest so it won't be fixed for non m1 fully until this gets merged.

PR #1179, opened 2021-11-23
DP-23360: Display current breadcrumb for authors below the parent field

arthurbaghdas:
**Description:**

DP-23360

**To Test:**

---

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -322,7 +322,8 @@
                 "Fix duplicated image selection when click use selected button": "https://www.drupal.org/files/issues/prevent-duplication.patch"
             },
             "drupal/entity_hierarchy": {
-                "Hidden Weight field causes extra space in the form": "patches/entity_hierarchy-weight-widget-space-3245879-2.patch"
+                "Hidden Weight field causes extra space in the form": "patches/entity_hierarchy-weight-widget-space-3245879-2.patch",
+                "Enable entity_hierarchy_breadcrumb module to work on node edit routes": "patches/entity_hierarchy_breadcrumb_for_admin_routes.patch"
             },
             "drupal/geocoder": {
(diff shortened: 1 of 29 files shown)


composer.json
```diff
@@ -322,7 +322,8 @@
                 "Fix duplicated image selection when click use selected button": "https://www.drupal.org/files/issues/prevent-duplication.patch"
             },
             "drupal/entity_hierarchy": {
-                "Hidden Weight field causes extra space in the form": "patches/entity_hierarchy-weight-widget-space-3245879-2.patch"
+                "Hidden Weight field causes extra space in the form": "patches/entity_hierarchy-weight-widget-space-3245879-2.patch",
+                "Enable entity_hierarchy_breadcrumb module to work on node edit routes": "patches/entity_hierarchy_breadcrumb_for_admin_routes.patch"
```
weekbeforenext, line 326: Is this a patch from Drupal.org? Should it be? If it is, would be nice to include a link to the issue.

weekbeforenext (approved):
Nice work! Super cool. I just had the one comment about the patch, but this looks good to go.

weekbeforenext (changes requested):
Just reviewed on Tugboat and we need to disable the Breadcrumbs block in the Structure > Block layout page for the mass_theme. Looks good after that change.

composer.json
arthurbaghdas: This patch is not from Drupal.org. But we can make it happen. If we are going to introduce a patch for this to the community, I guess the best option will be to create a config form and store value in the config and depending on that value toggle visibility on admin pages. That would be a nice feature, but since this was not included in the scope of this ticket I've created a patch only for openmass.

PR #1182, opened 2021-11-24
DP-23356: Bulk parent update

tom-fleming:
**Description:**

DP-23356

**To Test:**

---


docroot/modules/custom/mass_admin_pages/mass_admin_pages.services.yml
```diff
@@ -4,3 +4,8 @@ services:
     tags:
       - {name: event_subscriber}
     arguments: ['@datetime.time', '@router.admin_context']
+
+  theme.negotiator.mass_admin_pages:
+    class: Drupal\mass_admin_pages\Theme\ThemeNegotiator
+    tags:
+      - { name: theme_negotiator, priority: -50 }
```
weekbeforenext, line 11: Probably need a blank line at the end of this file.

patches/DP-23411_node-hierarchy-async-support-save-only-changed.patch
```diff
@@ -247,6 +247,19 @@ index fc94005..87419e8 100644
...
+diff --git a/src/Plugin/Derivative/DynamicLocalTasks.php b/src/Plugin/Derivative/DynamicLocalTasks.php
+index bc8bd8a..e065096 100644
+--- a/src/Plugin/Derivative/DynamicLocalTasks.php
++++ b/src/Plugin/Derivative/DynamicLocalTasks.php
+@@ -80,7 +80,7 @@ class DynamicLocalTasks extends DeriverBase implements ContainerDeriverInterface
+       }
+       $this->derivatives["$entity_type_id.entity_hierarchy_reorder"] = [
+         'route_name' => "entity.$entity_type_id.entity_hierarchy_reorder",
+-        'title' => $this->t('Children'),
++        'title' => $this->t('Hierarchy'),
```
weekbeforenext, line 259: We might have been able to override this in a custom module but this patch works, too.

weekbeforenext (approved):
I think this looks good to go from code review perspective.

patches/DP-23411_node-hierarchy-async-support-save-only-changed.patch
beto-aveiga: @weekbeforenext I did try overriding it using a module. In fact, there is a file on this PR that should do this. However, it worked for the title on that route, but not for the local task when viewing the node. I spent at least 2 hours understanding why it wasn't changing, but I wasn't able to solve it 😞 ... So I add it to the patch.

docroot/modules/custom/mass_admin_pages/mass_admin_pages.services.yml
beto-aveiga: @weekbeforenext  sorry, I missed this comment before the merge.

PR #1186, opened 2021-11-25
Dp 23300 permission change for topic pages 4

beto-aveiga:
**Description:**
- Author & Editor role can edit any Topic Page
- Users that can Edit Any Topic Page but can't create a Topic Page can use the following fields when the field_restrict_link_management is not checked:
  - Intended audience
  - Organizations
  - Labels
  - Link groups
  - Workflow states.
  - Action (save/preview)
  
DP-23300

**To Test:**
- [ ] With author and editor role (or roles able to edit topic pages, but not create topic pages)
  - [ ] Edit any topic page, you should be able to edit only the fields mentioned above. 
- [ ] With an administrator (or any other role that can create topic pages) check the "Restrict link management to internal content team".
- [ ] With author and editor role
  - [ ] You shouldn't be able to edit any field. 

---


docroot/modules/custom/mass_fields/mass_fields.module
```diff
@@ -758,6 +758,64 @@ function mass_fields_form_node_info_details_form_alter(&$form, FormStateInterfac
...
+    'actions',
+  ];
+
+  foreach ($form as $key => &$element) {
+    if (in_array($key, $allowed_fields)) {
+      $element['#disabled'] = FALSE;
+      continue;
+    }
+
+    is_array($element) && (isset($element['#access']) || isset($element['#type'])) ?
```
weekbeforenext: It might be worth writing this conditional similar to the previous one, for ease of reading. Wouldn't affect functionality, so more of a nitpick.

weekbeforenext (approved):
One nitpick comment, but this seems good. Going to test on Tugboat before merge.

weekbeforenext (changes requested):
As an Author/Editor user on Tugboat, the field restrictions look right, but I can't save the node.

docroot/modules/custom/mass_fields/mass_fields.module
beto-aveiga: I just saw your comment. 
Sure, Thanks, I agree, indeed.
However, I think I changed that code already, and I can't find that pattern now.

beto-aveiga:
Thanks! @weekbeforenext it seems that when I moved to a new branch based on feature/DP-22668_topic-page-redesign, I lost a few things. I improved the test, and now it checks saving topic pages as well.

weekbeforenext (approved):

PR #1204, opened 2021-12-03
feature/DP-23537_parent_view_improvements2

joeg8612:
**Description:**
Change critera for content type for report "Pages with missing parents" so that authors can select multiple content types and can choose to include or exclude them.

DP-23537

**To Test:**
- [ ] View report and make sure you can filter for +1 content types and also exclude types.

---


weekbeforenext (approved):
Looks good to me.

PR #1208, opened 2021-12-06
DP-23549: move children message

tom-fleming:
**Description:**

DP-23549

**To Test:**
- [ ] Go to the Move Children tab of any node with children.
- [ ] Check that the following help text appears: "NOTE: Moving children in bulk is an experimental feature and currently is slow. It may be 1 minute or more before you see an indication that your move starts. In testing, we have seen it take two minutes to change the parent on 25 children. We will be evaluating this after release to production to assess performance and see if we can speed the process."

---

### DIFF
--- a/docroot/modules/custom/mass_hierarchy/mass_hierarchy.module
+++ b/docroot/modules/custom/mass_hierarchy/mass_hierarchy.module
@@ -1,0 +1,24 @@
+<?php
+
+/**
+ * @file
+ * Contains mass_docs.module.
+ */
+
+use Drupal\Core\Routing\RouteMatchInterface;
+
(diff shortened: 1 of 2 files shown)


docroot/modules/custom/mass_hierarchy/mass_hierarchy.module
```diff
@@ -0,0 +1,24 @@
+<?php
+
+/**
+ * @file
+ * Contains mass_docs.module.
```
tom-fleming:
```suggestion
 * Contains mass_hierarchy.module.
```

weekbeforenext (approved):

PR #1210, opened 2021-12-06
Add Akamai purger

weitzman:
**Description:**
To get this to work locally, you need to create a `setting.local.php` and add there `$config['akamai.settings']['disabled'] = FALSE;`. In addition, you need to add 3 env variables to your .env file (see .env.example changes). See Moshe for values.

Also adds handing of Akamai's True-Client-IP http header

DP-23555

**To Test:**
- [ ] Run `drush pqe` to empty the queue
- [ ] Edit a document node. Run `drush pqb` to browse the queue. Notice some URL validations. Those are the ones are configured  to go to Akamai.
- [ ] Run `drush pqw` to process the queue and send invalidations to Akamai (the tag ones are handled by our massgov_cloudflare purger but we have always swallowed those and not sent them along). If there were no shown errors, we succeeded. An outputted table shows this as well.
- [ ] Run `drush pqb` again and confirm that all invalidations are gone.

**To Do:**
- [ ] Add 3 values to global secrets file at Acquia.

---

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -6,6 +6,12 @@
 AC_API2_KEY=your-long-string-here
 AC_API2_SECRET=your-long-string-here
 
+#Akamai. Only needed if you are doing CDN work
+AKAMAI_ACCESS_TOKEN=
+AKAMAI_CLIENT_TOKEN=
+AKAMAI_CLIENT_SECRET=
+
+
 # New Relic API settings. Ask a release manager for the key.
 MASS_NEWRELIC_APPLICATION=46704875
 MASS_NEWRELIC_KEY=your-long-key-here
--- a/docroot/sites/default/settings.acquia.php
+++ b/docroot/sites/default/settings.acquia.php
@@ -29,6 +29,7 @@ if (file_exists($secrets_file)) {
  * Loads global secrets.
  *
  * - Used by TFA's encryption profile'.
+ * - Used by Akamai module's Key module integration.
  */
 $secrets_file_global = '/home/massgov/.app/secrets.settings.php';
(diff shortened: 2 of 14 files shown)


docroot/sites/default/settings.acquia.php
```diff
@@ -104,12 +112,17 @@
       $config['stage_file_proxy.settings']['origin'] = FALSE;
       $config['media_entity_download.settings']['external_file_storage'] = 0;
       $settings['mass_caching.hosts'] = ['edit.mass.gov', 'www.mass.gov'];
+
+      $config['akamai.settings']['disabled'] = FALSE;
+      $config['akamai.settings']['basepath'] = 'www.mass.gov';
```
arthurbaghdas: @weitzman in the  `akamai.setting.yml` file is see that the protocol is also included in the url `basepath: 'https://stage.mass.gov'` . Should we change it here as well?

.env.example
```diff
@@ -6,6 +6,12 @@
 AC_API2_KEY=your-long-string-here
 AC_API2_SECRET=your-long-string-here
 
+#Akamai. Only needed if you are doing CDN work
+AKAMAI_ACCESS_TOKEN=
+AKAMAI_CLIENT_TOKEN=
+AKAMAI_CLIENT_SECRET=
```
arthurbaghdas, line 12: @weitzman can you provide these values?

docroot/sites/default/settings.acquia.php
weitzman: Yes, good catch. Fixed.

PR #1211, opened 2021-12-06
DP-23552/fix-key-message-top

clairesunstudio:
https://github.com/massgov/mayflower/pull/1567


clairesunstudio:
Tested on mobile and looks good!

After:
IMG_0019

Before:
IMG_0020

PR #1215, opened 2021-12-07
Dp 23570 fixes to make sure children cant be dragged saved to an unallowed parent

beto-aveiga:
**Description:**
      - Disable the “save” button if an unallowed parent type is chosen.
      - Don’t allow published pages to be children of unpublished pages
      - Don’t allow people without the Create topic permission to change the parent of a topic page.
      - Don’t show any nodes that are not published in the hierarchy.
      - Move the help text in the hierarchy tab to be above the tabs and below the H1.  (Like Move Children tab)
      - Change Hierarchy help text.
      - Reorder the tabs when editing content so that Hierarchy is to the left of “Move Children”
      - Change “Move Children” to not allow:
        --- Move a published page to have an unpublished parent.
        --- Move a page to have a parent of a type that is not allowed for the child type.

DP-23570

---


tom-fleming (approved):
LGTM. Merge conflict needs to be resolved.

beto-aveiga:
Outdated. All work is on https://github.com/massgov/openmass/pull/1223

PR #1216, opened 2021-12-08
Make DB commands functional in DDEV

weitzman:
**Description:**
DDEV command s like `mysql`, `snapshot`, `import-db` will all start woking as expected with this PR. We accomplish this by changing the mysql datadir in our sanitized image. This is actually what we used to do before DDEV.

DP-23584

**To Test:**
- [ ] None. The `populate` code runs nightly at Circle and its annoying to mimic that locally. Reviewer should review code and _Approve_ this PR, and then ping me. I will merge it and then run the Circle mysql_rebuild workflow on the develop branch. This will ensure that anyone who does a pull will get a working DB even on same day.

---


beto-aveiga (approved):
LGTM

PR #1244, opened 2021-12-17
Dp 22872 change pages linking here to use entity usage module

beto-aveiga:
**Description:**

https://massgov.atlassian.net/browse/DP-22872

**To Test:**

---


docroot/modules/custom/mass_entity_usage/tests/src/ExistingSite/EntityUsageTest.php
```diff
@@ -0,0 +1,172 @@
...
+
+  private $user;
+
+  /**
+   * Create the user.
+   */
+  protected function setUp() {
+    parent::setUp();
+    $user = User::create(['name' => $this->randomMachineName()]);
+    $user->addRole('administrator');
```
joeg8612, line 28: Can we change this to "Content administrator" and "tester" role actually so more can test it?

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
```diff
@@ -0,0 +1,149 @@
...
+    // Get the moderation state.
+    $content_moderation_state = ContentModerationState::loadFromModeratedEntity($node);
+    if ($content_moderation_state instanceof ContentModerationStateInterface
+      && !$content_moderation_state->get('moderation_state')->isEmpty()) {
+      $state_name = $content_moderation_state->get('moderation_state')->value;
+      // These states should be not be tracked.
+      $skipped_states = [
+        'unpublished',
+        'trash',
+      ];
```
weitzman, line 72: 1. We have constants for each state so you dont have to use strings. See MassModeration class
1. It may make sense to make a method like  getPrepublishedStates in that same class for all the states we want here. Do this if you think we might need same grouping in the future.

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
```diff
@@ -0,0 +1,149 @@
...
+  $storage = \Drupal::entityTypeManager()->getStorage($entity->getEntityType()->id());
+  /** @var \Drupal\Core\Entity\ContentEntityBase */
+  $loaded_entity = $storage->load($entity->id());
+  $vid = $loaded_entity->getRevisionId();
+
+  // Delete reference to itself.
+  $delete = \Drupal::database()->delete('entity_usage');
+  $delete->condition('target_id', $entity->id());
+  $delete->condition('source_id', $entity->id());
+  $delete->execute();
```
weitzman:
```suggestion
  $delete = \Drupal::database()->delete('entity_usage')
    ->condition('target_id', $entity->id());
    ->condition('source_id', $entity->id());
    ->execute();
```
Same for the queries below.

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
```diff
@@ -0,0 +1,149 @@
...
+function mass_entity_usage_keep_only_current_vid($entity) {
+
+  if (!method_exists($entity, 'getRevisionId')) {
+    return;
+  }
+
+  // Only intervene on tracked entities.
+  $to_track = \Drupal::service('config.factory')
+    ->get('entity_usage.settings')
+    ->get('track_enabled_source_entity_types');
```
weitzman: Minor point, but it seems like this should be a method in entity_usage or in custom code. Going to config here is a bit too low level for my taste.

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
```diff
@@ -0,0 +1,149 @@
...
+/**
+ * Deletes usage data from revisions that are not the current revision.
+ *
+ * Entity usage also stores data from other revisions than the current one,
+ * but we don't want/need that data, we don't want to see that data in the
+ * entity_usage table either.
+ */
+function mass_entity_usage_keep_only_current_vid($entity) {
+
+  if (!method_exists($entity, 'getRevisionId')) {
```
weitzman:
```suggestion
  if (!$entity->getEntityType()->isRevisionable()) {
```

docroot/modules/custom/mass_entity_usage/tests/src/ExistingSite/EntityUsageTest.php
weitzman: This is test code so it wont affect anything that a web user would do. Does that answer your request?

weitzman (changes requested):
Code looks good. Only minor suggestions. 

Large patches these are hard to maintain over time, as the module can change and our patches no longer apply cleanly. It would be good if we had tickets in the backlog so someone follows up and works to get those committed back to entity_usage and media_entity_download @joeg8612

docroot/modules/custom/mass_entity_usage/tests/src/ExistingSite/EntityUsageTest.php
joeg8612: Yes

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
beto-aveiga: I wasn't able to find that method, however, I noticed that the way to get that config on the entity_usage code was doing this: `\Drupal::config('entity_usage.settings')->get('track_enabled_source_entity_types');`, which is the change I'm going to apply. Let me know if that is still correct for you.

weitzman: I guess if they do it that way then its fine for us to do same. No need to make this change unless you want to.

weitzman (approved):
Good to merge when green

PR #1250, opened 2021-12-22
Dp 23681 investigate issues related to parent field release reported by authors

beto-aveiga:
**Description:**
      - Avoids storing records for non-current revisions on the nested_set_field_primary_parent_node table.
      - To clear nested_set_field_primary_parent_node table from non-current revisions.
        - Adds an update hook.
        - Implements hook_entity_update & hook_form_alter.

See comment https://massgov.atlassian.net/browse/DP-23681?focusedCommentId=259229

DP-23681

---


docroot/modules/custom/mass_hierarchy/mass_hierarchy.install
```diff
@@ -0,0 +1,19 @@
...
+use Drupal\node\Entity\Node;
+
+/**
+ * Implementations of hook_update_N().
+ */
+
+/**
+ * Remove records on nested set table from non-current revisions.
+ */
+function mass_hierarchy_update_8105() {
```
arthurbaghdas: I think we should start from 8101 instead of 8105.

docroot/modules/custom/mass_hierarchy/mass_hierarchy.module
```diff
@@ -123,6 +146,15 @@ function mass_hierarchy_views_post_render(ViewExecutable $view) {
  * Implements hook_form_alter().
  */
 function mass_hierarchy_form_alter(&$form, FormState $form_state, $form_id) {
+
+  if (isset($form['#entity_type']) && $form['#entity_type'] == 'node') {
```
arthurbaghdas, line 150: Since we are trying to execute this code for the existing nodes I would suggest using hook_node_update

beto-aveiga: I tried that, but that won´t work, I did that but the application started to fail in different places because there was a lock on the table we wanted to modify. I made comments on the ticket about it, and also there must be Circle CI executions that you can see with that information.

docroot/modules/custom/mass_hierarchy/mass_hierarchy.install
beto-aveiga: Done, thanks!

docroot/modules/custom/mass_hierarchy/mass_hierarchy.module
tom-fleming: @beto-aveiga I know we discussed this some last week, but I can't recall the outcome. Is running cron enough to get us the results we want here?

beto-aveiga: @tom-fleming  Yeah, previously, cron wasn't working because I made a typo on the function name.
Now it works as expected. I even wrote a test to ensure that scenario.

PR #1261, opened 2022-01-03
Allow purge logging to New Relic, and avoid lock api locks during purge - requires a drupal patch

weitzman:
DP-23756

---


tom-fleming (approved):
LGTM

PR #1264, opened 2022-01-04
DP 23678 remove cloudflare, fix akamai purge in prod

weitzman:
**Description:**
- Akamai module purges its prod env when run in acquia prod
- Remove Cloudflare
- Remove a step in 2 Behat tests as that url purge is already tested by AutomatedPurgingTest in phpunit.
- The changes in field config are unrelated. They happen when you `drush config:export` so likely someone edited config file directly and now we have a persist difference between the sync dir and drupal's export. These changes resolve that difference.

DP-23678

**To Test:**
- [ ] No way to test Akamai change.
- [ ] If PR is green, Cloudflare should be removed properly from config.yml.
- [ ] Go to https://mass.local/admin/config/development/performance/purge and assure that Cloudflare is not listed in the Cache Invalidation section. This assures that its purger was removed.

---


weitzman:
I'm not sure why ci/circleci check is outstanding. This PR removes that check. I think we can proceed.

PR #1273, opened 2022-01-05
DP-23763 remove unpublished from the validation

beto-aveiga:
**Description:**
      - Unpublished children of a node are ignored when unpublishing a parent.
      - A node cannot be published if its parent is not published.

### Screnshot
image

DP-23763

**To Test:**

---


weitzman (approved):
Nice work. Thanks for the tests.

PR #1291, opened 2022-01-13
DP-23222 faster alerts and alert placement tests

beto-aveiga:
**Description:**

DP-****

**To Test:**

---

### DIFF
--- a/docroot/modules/custom/mass_alerts/js/alerts-block.js
+++ b/docroot/modules/custom/mass_alerts/js/alerts-block.js
@@ -67,25 +67,35 @@
           }
         }
 
-
         if (path) {
-          $.ajax({
-            type: 'GET',
-            url: path,
-            cache: true,
-            success: function (content) {
-              if (!content) {
-                $this.hide();
-                return;
-              }
-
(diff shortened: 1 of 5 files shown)


docroot/modules/custom/mass_alerts/js/alerts-block.js
```diff
@@ -67,25 +67,35 @@
...
+            $this.html(content);
+            if (removeContainer) {
+              $this.find('.ma__page-banner__container').removeClass('ma__page-banner__container');
             }
-          });
+            $(document).trigger('ma:AjaxPattern:Render', [{el: $this}]);
+          };
+
+          // Check if the data is already there.
+          if (document.prefetchAlertsData[path]) {
```
tom-fleming, line 85: @beto-aveiga  It seems like maybe this code living in a behavior is where we currently have a delay in seeing the sitewide alert at least. At the moment, I think we need to wait for the DOM to load, then jQuery, then the site's JS with its behaviors before we can call this JS here.

beto-aveiga:
Outdated.

PR #1301, opened 2022-01-18
Dp 23876 make fixes to entity usage to stabilize performance

beto-aveiga:
**Description:**
      - Entity usage: database updates happen at the end of the transaction so they don’t block anything.
      - Entity usage tracking queue: removed processing out of corn.
      - To process this queue run: `drush queue:run entity_usage_regenerate_queue --time-limit=60` where 60 is for 60 seconds of processing.

DP-23876

---


weitzman:
The cron patch is not in this PR yet.

weitzman:
Please also submit the cron patch to drupal.org. If that sounds onerous, use plugin_alter hook and no patch required.

beto-aveiga:
> Please also submit the cron patch to drupal.org. If that sounds onerous, use plugin_alter hook and no patch required.

Thanks, @weitzman, I was unsure about the plugin_alter hook to work for this situation, but it makes sense, I will try it.
I didn't submit a patch because for most of the scenarios processing entity usage queue items in cron will be just fine.

beto-aveiga:
@weitzman  done, thanks.
The code was straightforward, but it took me some time to verify it was working

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
```diff
@@ -146,3 +146,13 @@ function mass_entity_usage_entity_insert(EntityInterface $entity) {
...
   mass_entity_usage_keep_only_current_vid($entity);
 }
+
+/**
+ * Implements hook_queue_info_alter().
+ */
+function mass_entity_usage_queue_info_alter(&$queues) {
+  // Removes entity usage queue from processing when cron runs.
+  if (isset($queues['entity_usage_regenerate_queue'])) {
+    $queues['entity_usage_regenerate_queue']['cron']['time'] = 0;
```
weitzman:
```suggestion
    unset($queues['entity_usage_regenerate_queue']['cron']);
```
I didn't test this so please let us know if this works.

beto-aveiga: Right, based on `docroot/core/lib/Drupal/Core/Cron.php:171` is even better to unset instead of setting to 0.
I will test locally anyway. Thanks!

beto-aveiga: Tested locally. That works too.

weitzman:
In order to actually use the new code, we have to change our config file for this module `$config['entity_usage.settings']['background_tracking'] = TRUE;`.

beto-aveiga:
@weitzman great catch! Fixed. Many thanks!

weitzman:
Code looks good but tests are failing.

beto-aveiga:
@weitzman those tests started failing after a fix was made because the builds were failing. A week ago or 2.
As a side note, when I reviewed failing in my local, tests were passing. I think these issues are complicated to debug and fix. I can fix those issues but that should be done in another ticket.

See https://app.circleci.com/pipelines/github/massgov/openmass, these issues are already on develop.

weitzman:
But we cant verify that entity usage is working without those tests. I think we should fix the fails in this ticket. This PR is about enabling entity usage. Sorry.

beto-aveiga:
@weitzman no problem, that makes sense too. I will add that to the ticket description and start working on fixing those issues.

weitzman (approved):

PR #1305, opened 2022-01-19
Dp 23887 add batch 3 fields to integration branch for landing page features on service page

beto-aveiga:
**Description:**
Adding the following paragraphs to field.field.paragraph.service_section.field_service_section_content:
        - campaign_features
        - caspio_embed
        - featured_item_mosaic
        - key_message_section
        - pull_quote
        - stat
        - tableau_embed
        - video_with_section

DP-23887

**To Test:**

---


conf/drupal/config/views.view.all_documents.yml
```diff
@@ -1142,7 +1142,7 @@ display:
           exposed: true
           expose:
             operator_id: field_collections_target_id_op
-            label: 'Collections'
```
clairesunstudio, line 1145: these changes don't seem to be related. Please double-check if they are correct

clairesunstudio:
https://pr1479-s6wo0wsn4wnnkbhshjaopg4styqd1ktd.tugboatqa.com/the-new-service
I think we should enforce the link group heading H3. If the headings are not added, then the buttons are governed by the accordion title.
Screen Shot 2022-06-06 at 12 33 40 PM

conf/drupal/config/views.view.all_documents.yml
beto-aveiga: @clairesunstudio  They are not related but I also think that they aren't incorrect. After solving merge conflicts, I imported the changes and then exported them to see if the data was exactly the same, but I got these small differences. I think it is a good practice to update config files in this kind of situation instead of accumulating small differences over time. What do you think?

beto-aveiga:
Outdated.

PR #1310, opened 2022-01-24
Send edit.mass.gov URLs to Akamai purger

weitzman:
**Description:**
Keep this in _Draft_ until we understand the failure condition better.

If problem happens again, before clearing the purge queue
- Use `drush pqb` to browse the queue
- Enable all debugging `drush  p:debug-en`. 
- Run `drush pqw -vvv` and see if we can observe the Akamai request that is failing
- SSH into Prod and then capture full contents of the purge queue `drush sql:dump --tables-list=purge_queue_alt  --result-file=~/queue.sql`
- Disable debugging with `drush p:debug-dis`
- Now proceed with usual `drush pqe` to empty purge queue, and clear varnish via Acquia UI.

DP-23882

**To Do:**
- [ ] Add patch via composer

---


tom-fleming:
@weitzman It looks like there are some PHPCS issues that need to be addressed. Besides that, I think this PR looks good

PR #1315, opened 2022-01-26
Dp 23222 modify front end to get alerts to load faster  3

beto-aveiga:
**Description:**

DP-****

**To Test:**

---


docroot/modules/custom/mass_alerts/js/alerts-block.js
```diff
@@ -2,29 +2,137 @@
  * @file
  */
 
-(function ($, Drupal, drupalSettings) {
+var drupalLike = {behaviors: {massAlertBlocks: {}}};
+
+var jQueryLike = function (elemOrSelector, context) {
```
tom-fleming, line 10: @beto-aveiga I think both drupalLike and jQueryLike make sense, but I wonder if we should consider either some documentation or another name that might describe their intention a little more clearly. I just think future developers might look at this in its current state and wonder why we weren't loading Drupal and jQuery in the header.

beto-aveiga: Makes sense. I will add a comment explaining the purpose of those Objects. Thanks!

tom-fleming: Great, thanks!

PR #1290, opened 2022-01-12
DP-23669 Add "language bar" links for pages that have professional translations

Javi-er:
**Description:**

- Re orders the languages in the UI to match the list on the ticket.
- Adds a new `translationLinks` on node templates containing links to the different translations, if there are.
- New "prepare" function for a new MF component.
- Renders translation links into templates.

**Jira:** (Skip unless you are MA staff)
[DP-23669](https://massgov.atlassian.net/browse/DP-23669)

**To Test:**
- [ ] Refer to the ticket.

---

[Peer Review Checklist](https://github.com/massgov/openmass/blob/develop/docs/peer_review_checklist.md)

PR #1320, opened 2022-01-28
Revert "DP-23669 Add "language bar" links for pages that have professional translations"

weitzman:
Reverts massgov/openmass#1290. Develop was failing tests - particularly `composer validate` step in validate job.


weitzman:
Not needed. composer.lock has been fixed

PR #1322, opened 2022-01-31
Dp 23988 investigate performance issues with entity usage

beto-aveiga:
**Description:**
- Checks if paragraph must be tracked based on parent's node moderation state.

DP-23988

---


beto-aveiga:
No more time to keep researching on this topic.

PR #1329, opened 2022-02-02
Dp 24022 Disable entity usage and mass_entity_usage modules

beto-aveiga:
**Description:**

DP-24022 

**To Test:**

---


weitzman (changes requested):
Lets disable mass_entity_usage module as well.

beto-aveiga:
@weitzman I removed the entity_usage_module. Thanks!

weitzman:
We generally do not disable a module and remove its code in same deployment. The reason is that Drupal runs uninstall routines and it cant do so if the code is gone. If you do this you get (minor) error messages due to lack of cleanup. So, I think we need a PR that only uninstalls these 2 modules (i.e. just core.extensions.yml) and then a follow-up to do the rest. Sorry I did not think of this earlier.

weitzman:
Sorry, I mean that this PR can do evrything except for 
1. Dont yet remove entity_usage from composer
2. Dont yet remove mass_entity_usage from the codebase

beto-aveiga:
@weitzman makes total sense. No problem.

weitzman (approved):

PR #1338, opened 2022-02-14
Dp 24080 change entity usage to use a queue

beto-aveiga:
**Description:**

DP-24080

---


docroot/modules/custom/mass_entity_usage/tests/src/ExistingSite/EntityUsageTest.php
```diff
@@ -0,0 +1,248 @@
...
+use weitzman\DrupalTestTraits\ExistingSiteBase;
+use weitzman\LoginTrait\LoginTrait;
+
+/**
+ * Class EntityUsageTest.
+ */
+class EntityUsageTest extends ExistingSiteBase {
+
+  use LoginTrait;
+  use MediaCreationTrait;
```
weitzman, line 21: Lets use https://github.com/drupaltest/queue-runner-trait/. That package is already in our project. Should simplify our test.

docroot/modules/custom/mass_hierarchy/tests/src/ExistingSiteJavascript/HierarchyTest.php
```diff
@@ -118,27 +118,4 @@ public function testCronErasesOtherRevisions() {
     $this->assertSession()->pageTextContains($child1Node->label());
   }
 
-  /**
-   * Tests node edit erases other revisions other than the current.
-   */
-  public function testNodeEditFormErasesOtherRevisions() {
-    $this->drupalLogin($this->createRandomUser('content_team'));
```
weitzman, line 125: Why is this hierarchy test removed? I know the form alter is no longer needed but we still want to test that our queue worker does its just, and removed unneeded versions. Is that tested elsewhere?

weitzman (commented):
I havent finished reviewing this, but here are a few quick things. Also, I think we should use https://www.drupal.org/project/queue_unique for this queue because we are just shooting ourselves in the foot by adding duplicate queue items when while user's fixing typos and what not.

docroot/modules/custom/mass_entity_usage/tests/src/ExistingSite/EntityUsageTest.php
beto-aveiga: Awesome! I didn't know about that trait. Thanks.

docroot/modules/custom/mass_hierarchy/tests/src/ExistingSiteJavascript/HierarchyTest.php
beto-aveiga: I remove that test because that test was checking other revisions were deleted on Node Edit form, so this is not happening now. Only in cron.

beto-aveiga:
> I havent finished reviewing this, but here are a few quick things. Also, I think we should use https://www.drupal.org/project/queue_unique for this queue because we are just shooting ourselves in the foot by adding duplicate queue items when while user's fixing typos and what not.

Nice suggestion! Thanks.

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
```diff
@@ -0,0 +1,116 @@
...
+ * @param int $vid
+ *   The entity revision id.
+ *
+ * @return bool
+ *   If true, the content should be tracked. If false, do not track.
+ */
+function _mass_entity_usage_tracking_check($entity_type, $id, $vid) {
+  if ($entity_type == 'paragraph') {
+    // Verify the paragraph revision is in the paragraphs_item_field_data table.
+    $query = \Drupal::database()->select('paragraphs_item_field_data', 'pifd');
```
weitzman: Have we checked the performance if this query with EXPLAIN? Could cause slowness if there is no index.

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
```diff
@@ -0,0 +1,116 @@
...
+    }
+  }
+  return TRUE;
+}
+
+/**
+ * Implements hook_entity_insert().
+ */
+function mass_entity_usage_entity_insert(EntityInterface $entity) {
+  Drupal::queue('mass_entity_usage_delete_not_current_revision_usages')->createItem([
```
weitzman: What queues does this feature use and why does each exist? 

For example, I cant think of a reason why regenerate (i.e. the bulk recalculation) has to be a different queue from the entity add/edit queue.

Also, I cant know why mass_entity_usage_delete_not_current_revision_usages is queued in the first place. Didn't we develop a function which deletes all non-current revisions for any node? Maybe I'm misremembering that.

docroot/modules/custom/mass_entity_usage/tests/src/ExistingSite/EntityUsageTest.php
```diff
@@ -0,0 +1,216 @@
...
+    ]);
+
+    $node_org = $this->createNode([
+      'type' => 'org_page',
+      'title' => 'Test Organization',
+      'uid' => $this->user->id(),
+      'moderation_state' => MassModeration::PUBLISHED,
+    ]);
+
+    // Entity usage tracking is done at the end of the requests, so we need to
```
weitzman: This is no longer true. We process in a queue now. Not sure how that affects the code.

docroot/modules/custom/mass_entity_usage/tests/src/ExistingSite/EntityUsageTest.php
```diff
@@ -0,0 +1,216 @@
...
+   */
+  private function processEntityUsageQueues() {
+    $this->runQueue('entity_usage_tracker');
+    $this->runQueue('mass_entity_usage_delete_not_current_revision_usages');
+  }
+
+  /**
+   * Assert that usage records are tracked properly.
+   */
+  public function testEntityUsageTracking() {
```
weitzman, line 140: We should test that we dont track UNPUBLISHED nodes. Also our paragraph handling is complex. Lets test that a bit deeper.

weitzman:
I gave more minor feedback.

If we wanted to proceed a bit more cautiously, we could forego the queueing added by this PR and only use the "massive tracking" that the module already provides. I imagine we would enqueue a full rebuild of usage tracking every saturday morning. It would run over the weekend and then not change untuil the following Saturday. If that goes smoothly, we could layer on the on-save enqueue that this PR adds. I dont have a strong feeling about this.

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
beto-aveiga: I think you are referring to mass_hierachy mass_hierarchy_delete_all_non_current_revisions_from_nested_set_table, that executes on mass_hierarchy_cron. That is not about usages, but hierarchy.

Regarding this queue `mass_entity_usage_delete_not_current_revision_usages`, is to delete usages of non-published revisions for nodes as we don't need that information and it was showing misleading references in our entity usage reports.

docroot/modules/custom/mass_entity_usage/tests/src/ExistingSite/EntityUsageTest.php
beto-aveiga: It doesn't affect the test. I will remove the comment.

docroot/modules/custom/mass_entity_usage/tests/src/ExistingSite/EntityUsageTest.php
beto-aveiga: Right. I forgot to test UNPUBLISHED as well.
I will add a more complex paragraph example.
Thanks!

beto-aveiga:
@weitzman I think also when site usage is low the initial entity usage tracking can be done, probably every day in a specific hour range.

> I imagine we would enqueue a full rebuild of usage tracking every Saturday morning

I think this is only needed one time when this code is deployed.

Regarding the on-save enqueue, this can be measured and tested easily in the console, if time/performance looks okay, I mean, if the process is fast enough, we can configure the scheduled task.

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
beto-aveiga: I haven't. But it makes sense to create an index on id and revision_id on the paragraphs_item_field_data, as insertions will be fewer compared to the number of reads.
I will create an index for that. Thanks!

tom-fleming:
@beto-aveiga This looks good to me. We should remove this line in settings.acquia.php: https://github.com/massgov/openmass/blob/00f80d45472fe77d7a8662fe8adaa4b0534e8b3a/docroot/sites/default/settings.acquia.php#L116

docroot/modules/custom/mass_entity_usage/mass_entity_usage.info.yml
```diff
@@ -0,0 +1,8 @@
+name: Mass.gov Entity Usage Customization
+type: module
+description: Customizations for entity_usage contrib module.
+core: 8.x
+core_version_requirement: ^8 || ^9
```
deviantintegral: Given this is a custom module, any reason to still allow ^8?

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * Implements hook_entity_usage_block_tracking().
+ */
+function mass_entity_usage_entity_usage_block_tracking($target_id, $target_type, $source_id, $source_type, $source_langcode, $source_vid, $method, $field_name, $count) {
+  // Block tracking of entities with a node source that is unpublished or trash.
+  if (!_mass_entity_usage_tracking_check($source_type, $source_id, $source_vid)) {
+    return TRUE;
+  }
+  return FALSE;
```
deviantintegral:
```suggestion
return !_mass_entity_usage_tracking_check($source_type, $source_id, $source_vid);
```

docroot/modules/custom/mass_entity_usage/src/Controller/MassLocalTaskUsageController.php
```diff
@@ -0,0 +1,167 @@
...
+        ];
+      }
+    }
+
+    return $rows;
+  }
+
+  /**
+   * Recursive function to retrieve the parent node of a paragraph.
+   *
```
deviantintegral: This looks nearly identical to `Helper::getParentNode()` which is called in this PR. Any reason we're not reusing that?

https://github.com/massgov/openmass/pull/1338/files#diff-03c1e10b2da0d6902c3aaced183fdf4206f16b6b5fd7159f6038cd93c5e2b198R53

docroot/modules/custom/mass_entity_usage/src/Plugin/QueueWorker/DeleteNotCurrentRevisionUsages.php
```diff
@@ -0,0 +1,94 @@
...
+    $storage = \Drupal::entityTypeManager()->getStorage($entity_type);
+    /** @var \Drupal\Core\Entity\ContentEntityBase */
+    $entity = $storage->load($id);
+
+    if (
+      !$entity ||
+      !$entity->getEntityType()->isRevisionable()
+    ) {
+      return;
+    }
```
deviantintegral: Just noticing the different formatting compared to typical Drupal code and what's below. Either this is a "something is wrong with our code standards CI job" or "we really don't care about code formatting for conditionals".

docroot/modules/custom/mass_entity_usage/tests/src/ExistingSite/EntityUsageTest.php
deviantintegral: Looks like this is still todo?

docroot/modules/custom/mass_entity_usage/tests/src/ExistingSite/EntityUsageTest.php
```diff
@@ -0,0 +1,274 @@
...
+/**
+ * Class EntityUsageTest.
+ */
+class EntityUsageTest extends ExistingSiteBase {
+
+  use LoginTrait;
+  use MediaCreationTrait;
+  use QueueRunnerTrait;
+
+  private $user;
```
deviantintegral, line 29: Should we add a docblock so this gets autocomplete?

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
```diff
@@ -0,0 +1,116 @@
...
+ * @param int $id
+ *   The entity id.
+ * @param int $vid
+ *   The entity revision id.
+ *
+ * @return bool
+ *   If true, the content should be tracked. If false, do not track.
+ */
+function _mass_entity_usage_tracking_check($entity_type, $id, $vid) {
+  if ($entity_type == 'paragraph') {
```
deviantintegral: Given this is new code, I'm curious why it's a "private" module function instead of a set of one or two classes. In particular, seeing the branch between paragraph and node entity types makes me think this should be two methods or classes. Perhaps that would let us write a unit test around the conditionals on the node content type.

docroot/modules/custom/mass_entity_usage/mass_entity_usage.install
```diff
@@ -0,0 +1,39 @@
...
+
+use Drupal\Core\Database\Database;
+
+/**
+ * Add index to paragraphs_item_field_data on id and revision id.
+ */
+function mass_entity_usage_update_8106() {
+  $spec = [
+    'fields' => [
+      'id' => [
```
deviantintegral: I'm a little surprised to see these. Aren't there already indexes on this, since it's a primary key? And I would have expected `id, revision_id` to be a key too. If not, should we file an issue upstream?

deviantintegral (commented):
I haven't tested this out yet, but I have a few questions that I figured I'd get in before I closed this up for the day.

docroot/modules/custom/mass_entity_usage/src/Plugin/QueueWorker/DeleteNotCurrentRevisionUsages.php
beto-aveiga: I will put everything into one line. The split is unnecessary. Thanks!

docroot/modules/custom/mass_entity_usage/mass_entity_usage.install
beto-aveiga: Right, those 2 keys are primary keys by themself, but I'm creating a new index based on those 2 keys.
Not sure if we have to post a fix/patch in the issue upstream.
AFAIK an index means you get faster reads, but slower inserts.

docroot/modules/custom/mass_entity_usage/tests/src/ExistingSite/EntityUsageTest.php
beto-aveiga: Agree. Will do.

docroot/modules/custom/mass_entity_usage/src/Controller/MassLocalTaskUsageController.php
beto-aveiga: Good cacth! I will use `Helper:getParentNode` only, remove the other function and see if tests pass. Thanks!

beto-aveiga: Tests are passing locally 👍🏼

docroot/modules/custom/mass_entity_usage/tests/src/ExistingSite/EntityUsageTest.php
beto-aveiga: The suggestion is/was implemented. Thanks.

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
beto-aveiga: @deviantintegral  I'm not clear on this one. I get it to refactor it to 2 methods instead of one, but using a class or 2... wouldn't be too much for 2 functions? Would you elaborate more on this, please?

docroot/modules/custom/mass_entity_usage/mass_entity_usage.info.yml
beto-aveiga: "Coping/pasting" is the reason haha. I will change it. Thanks!

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
deviantintegral: We generally recommend defaulting new code to an object-oriented style rather than a procedural style. Ideally, a custom module file just contains hook implementations that delegate to services and classes. It's less obvious in this example where there is only one function in the module file. Usually, by the time a module file becomes too large to be easy to maintain and understand, there also isn't the appetite for a major refactoring.

We'd usually use [Typed Entity](https://www.lullabot.com/articles/drupals-bundle-classes-empower-better-code) to allow us to encapsulate the business logic. That also makes it easier for new developers to learn what each bundle type can do, without having to search through many module files. See the article for details on the limitations of core's bundle classes. With that architecture, we'd end up with something like:

1. A new interface such as `\Drupal\mass_entity_usage\TrackableEntityInterface`:
  - It would contain one method, just `isTrackable(): bool`.
2. We would implement the interface on the classes for each bundle.
  - The paragraph block and the node block in the current if statement would move to each of those classes.
3. `\mass_entity_usage_entity_usage_block_tracking` would then be:

```php
$entity = $entityStorageManager->loadRevision($vid);
if ($entity) {
  $wrapped = $repository_manager->wrap($entity);
  if ($wrapped instanceof TrackableEntityInterface) {
    return $wrapped->isTrackable();
  }
}

return FALSE;
```

There's similar ways to do this without bundle classes or typed entities too. But this is a big topic, so I'll stop here for now!

docroot/modules/custom/mass_entity_usage/mass_entity_usage.install
deviantintegral: Ooo, I think I realized why this index isn't in core.

The revision ID is already a **unique key**. `\Drupal\Core\Entity\EntityStorageInterface::loadRevision()` takes just a revision ID, not an entity ID. In other words, adding the primary key doesn't to the query doesn't actually do anything! The only way you'd need this index is if revision IDs were not unique and started from `1` for each entity.

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
beto-aveiga: Thanks for the advice and the article @deviantintegral! Much appreciated.

docroot/modules/custom/mass_entity_usage/mass_entity_usage.install
weitzman: Lets use our existing bundle classes for now, and skip Typed Entity. We waited this long to intro bundle classes because we want to stick to vanilla core in this case.

deviantintegral: @weitzman @beto-aveiga what do you think about my comment on the unique key?

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
weitzman: Thanks. Could we have a query that runs once a day and removes non-current revisions? If it cant delete immediately, it could queue up the deletes. The improvement is that we dont do any enqueue on each entity CRUD.

beto-aveiga: I think we are not processing queues on the entities' CRUD. We are just adding items to the queue. And because we are using queues we have to manually add a scheduled task to do this, which gives us control over its execution time and frequency. @weitzman

weitzman: Yes, that makes sense. But my suggestion is more robust. Consider what happens if we have to empty the deletion queue for some reason. With the current solution, will have keep non-current revisions in the table for a very long time (until each one of them is next edited). With my suggestion, it is no problem. We will re-enqueue the non-current revisions immediately.

beto-aveiga: Oh! I get it. Makes sense.

docroot/modules/custom/mass_entity_usage/mass_entity_usage.install
beto-aveiga: @deviantintegral interesting! I think you are correct. Which also means that this condition can be removed...  https://github.com/massgov/openmass/pull/1338/files#diff-d04531c7b9896c1f1c59a10b38ae00d4c957267c33912dc7d7a41697dc31423fR63  Right?

deviantintegral: Yes, I think so!

weitzman:
Also, needs a reroll for https://github.com/massgov/openmass/pull/1420 (composer.lock changed there)

PR #1343, opened 2022-02-15
DP-23509 CircleCI omit trigger

weitzman:
**Description:**
Slightly faster and cleaner pipeline without the "double run" trigger thing that was put in to handle testing of PRs from forks. We have only ever gotten 2 of those and one was from Melissa. That feature is not worth the complexity.

This PR does not use Dynamic Configuration from CircleCI as it got more complicated to have an Org, and main config.yml, and various other small config.yml files. 

DP-23509

**To Test:**
- [x] If its green its good

---


tom-fleming:
LGTM

PR #1350, opened 2022-02-17
DP-24111: Add bulk feature to add collection information to documents

arthurbaghdas:
**Description:**

DP-24111

**To Test:**

---


docroot/modules/custom/mass_views/src/Plugin/Action/AddCollectionsDocuments.php
```diff
@@ -0,0 +1,210 @@
...
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\taxonomy\Entity\Vocabulary;
+use Drupal\views_bulk_operations\Action\ViewsBulkOperationsActionBase;
+use Drupal\Core\Session\AccountInterface;
+use Drupal\Core\StringTranslation\StringTranslationTrait;
+
+/**
+ * Allows to add Collections field value.
+ *
+ * @see https://www.drupal.org/docs/contributed-modules/views-bulk-operations-vbo/creating-a-new-action#s-2-action-class
```
deviantintegral, line 19: 👏

docroot/modules/custom/mass_views/src/Plugin/Action/AddCollectionsDocuments.php
```diff
@@ -0,0 +1,210 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute($entity = NULL) {
+
+    $config = $this->getConfiguration();
+    $new_collection_id = $config['new_collection'];
+
+    $media_storage = \Drupal::entityTypeManager()->getStorage('media');
```
deviantintegral: Is it a convention on this project to not use dependency injection in custom plugins using `\Drupal\Core\Plugin\ContainerFactoryPluginInterface`?

docroot/modules/custom/mass_views/src/Plugin/Action/AddCollectionsDocuments.php
```diff
@@ -0,0 +1,210 @@
...
+          $entity->field_collections->appendItem($id);
+        }
+      }
+      else {
+        $entity->field_collections = $new_collection_id;
+      }
+    }
+
+    $entity->setNewRevision(TRUE);
+    $entity->setRevisionUserId(\Drupal::currentUser()->id());
```
deviantintegral: This block is duplicated in `\Drupal\mass_views\Plugin\Action\ChangeCollections`. Should we be refactoring this?

docroot/modules/custom/mass_views/src/Plugin/Action/AddCollectionsDocuments.php
```diff
@@ -0,0 +1,210 @@
...
+  public function execute($entity = NULL) {
+
+    $config = $this->getConfiguration();
+    $new_collection_id = $config['new_collection'];
+
+    $media_storage = \Drupal::entityTypeManager()->getStorage('media');
+    $vid = $media_storage->getLatestRevisionId($entity->id());
+    $create_draft = $vid != $entity->getRevisionId();
+
+    if (is_array($new_collection_id)) {
```
deviantintegral, line 88: What would cause this value to not be an array? It looks like if this is a scalar, the only thing we do is save the entity and related media item, without ever referencing the new collection ID again.

Should this variable be called `new_collection_ids`?

docroot/modules/custom/mass_views/src/Plugin/Action/AddCollectionsDocuments.php
```diff
@@ -0,0 +1,210 @@
...
+          }
+        }
+        else {
+          $media_latest->field_collections = $new_collection_id;
+        }
+      }
+      $media_latest->save();
+    }
+
+    return $this->t('Updated collections for') . ' ' . $entity->label() . ' - ' . $entity->id();
```
deviantintegral, line 125: 🤔 should this be using placeholders instead of concatenation? At the least, it doesn't look like this is (easily) exploitable XSS, as this string is still escaped when rendered as a message. `\Drupal\views_bulk_operations\Action\ViewsBulkOperationsActionInterface::executeMultiple` says we can return either a string _or_ translatable markup, so let's do that? If I haven't missed anything here, we should probably update `ChangeCollections` too (or file a new ticket for that).

docroot/modules/custom/mass_views/src/Plugin/Action/AddCollectionsDocuments.php
```diff
@@ -0,0 +1,210 @@
...
+    if (!empty($target_bundles)) {
+
+      $vocabularies = Vocabulary::loadMultiple($target_bundles);
+
+      $form['#list'] = $this->context['list'];
+
+      $form['actions']['submit']['#value'] = $this->t('Add collections');
+
+      $form['new_collection'] = [
+        '#type' => 'checkbox_tree',
```
deviantintegral, line 184: TIL'ed about the term_reference_tree module. Nice!

docroot/modules/custom/mass_views/src/Plugin/Action/AddCollectionsDocuments.php
```diff
@@ -0,0 +1,210 @@
...
+    $items = _term_reference_tree_flatten($element, $form_state);
+    $value = [];
+    if ($element['#max_choices'] != 1) {
+      foreach ($items as $child) {
+        if (!empty($child['#value'])) {
+          // If the element is leaves only and select parents is on,
+          // then automatically add all the parents of each selected value.
+          if (!empty($element['#select_parents']) && !empty($element['#leaves_only'])) {
+            foreach ($child['#parent_values'] as $parent_tid) {
+              if (!in_array([$element['#value_key'] => $parent_tid], $value)) {
```
deviantintegral, line 214: I often go by a rule of "limit nested blocks to 3 levels", and if you have more it's a signal that a new private method should be refactored out. I don't see it as a blocker for this PR, unless we end up having to change these lines anyways.

deviantintegral (commented):
This is looking good! I tested locally as I wanted to make sure I could actually run the site, and it looks to be working as advertised. I've left a few notes and questions, of which many may be "Andrew's first review" type questions and not actually issues with the code.

docroot/modules/custom/mass_views/src/Plugin/Action/AddCollectionsDocuments.php
arthurbaghdas: I think I've grabbed the code form the contrib module,I didn't think of that. Do you think it's critical and worth attention?

docroot/modules/custom/mass_views/src/Plugin/Action/AddCollectionsDocuments.php
arthurbaghdas: hmm, interesting. What's your suggestion?

docroot/modules/custom/mass_views/src/Plugin/Action/AddCollectionsDocuments.php
arthurbaghdas: This was also duplicated from other custom modules.Not sure exactly about the condition that we need to check if it's array or not.

docroot/modules/custom/mass_views/src/Plugin/Action/AddCollectionsDocuments.php
arthurbaghdas: If you think that's an issue we have many places which return strings, should we address all of them and replace tstring with Translatable Markup?

docroot/modules/custom/mass_views/src/Plugin/Action/AddCollectionsDocuments.php
arthurbaghdas: Refactored the code.

docroot/modules/custom/mass_views/src/Plugin/Action/AddCollectionsDocuments.php
weitzman: **MW: woops, i'm reviewing the wrong issue**

@deviantintegral see question above

deviantintegral: We could pull it out into a trait that both collection classes use? Or perhaps just a new class. This would make a good case for typed entity / bundle class logic too.

In case I wasn't clear, this could totally be a followup or a new ticket. PHPStorm will still highlight the duplicated fragments (which it doesn't now, due to injecting dependencies just here!).

docroot/modules/custom/mass_views/src/Plugin/Action/AddCollectionsDocuments.php
deviantintegral: This seems like a great example of where unit tests and code coverage can help. I bet if we had a test documenting this case, we'd all know why it was important (or why it's not relevant anymore).

Anyways, perhaps we just file a ticket to investigate this. Nothing like having `if` statements that we don't understad!

docroot/modules/custom/mass_views/src/Plugin/Action/AddCollectionsDocuments.php
deviantintegral: Yeah, that seems like a good idea. @weitzman do you agree? If so, I think we could file a ticket to add a DrupalPractice check to identify these in CI to improve site security.

docroot/modules/custom/mass_views/src/Plugin/Action/AddCollectionsDocuments.php
deviantintegral: No, I would say leave it at this point unless we're making changes to these lines. But, let's keep it in mind for future work?

deviantintegral (approved):

docroot/modules/custom/mass_views/src/Plugin/Action/AddCollectionsDocuments.php
deviantintegral: Filed https://massgov.atlassian.net/browse/DP-24333

docroot/modules/custom/mass_views/src/Plugin/Action/AddCollectionsDocuments.php
deviantintegral: Filed https://massgov.atlassian.net/browse/DP-24334

PR #1352, opened 2022-02-17
Dp 24122 validation error on info details

beto-aveiga:
**Description:**
Avoids browser clientside validation because not all fields can be focused by the browser when the user tries to submit the form. 

For example, a required field can be temporarily hidden in a tab and empty, then the user tries to submit the form, but because of the clientside validation the browser stops the submission to show a message regarding the empty but required field, but because the field is hidden the browser can't focus the field, then it shows an error in the console, however the final user doesn't know what is going on.


weitzman:
If we are going to disable client side validation for all entity forms, then I recommend we disable the clientside_validation module. We dont have important non-entity forms AFAIK.

weitzman:
If we are going to keep clientside validation, then we would presumably do it on a "best effort" basis, and not throw js errors when the field cant be found.

tom-fleming:
@weitzman We're not using the clientside_validation module, but we are using inline_form_errors. In this case, it's the actual browser's validation that checks for attributes like required that's causing the issue being reported. Adding novalidate to the entity forms allows inline_form_errors to work more as expected: https://www.drupal.org/docs/8/core/modules/inline-form-errors/inline-form-errors-module-overview#s-known-limitations

weitzman:
Thanks. Luis also explained this in chat. +1 from me.

PR #1356, opened 2022-02-21
DP-23372: Add the full URL as a field in the document CSV download file

arthurbaghdas:
**Description:**

DP-23372

**To Test:**

---


conf/drupal/config/views.view.all_documents.yml
```diff
@@ -1351,23 +1351,24 @@ display:
...
           exclude: false
           alter:
-            alter_text: false
-            text: '<span title="{{ field_file_mime }}">{{ field_title }}</span>'
-            make_link: true
-            path: 'https://www.mass.gov/media/{{ mid }}/download'
-            absolute: false
-            external: true
+            alter_text: true
+            text: 'https://www.mass.gov{{ view_media_1 }}/download'
```
deviantintegral, line 1367: I see this is hardcoded to the public URL. Should this be a relative link instead, so it works across different environments? Even so, it would probably be OK to do this as a new ticket, since this is unrelated and how it works on production today.

deviantintegral (commented):
This looks close! Config changes look good (though I didn't do a line-by-line read). While I can see the new column, I'm not getting a downloaded file. Instead, after the batch completes I see the CSV in my browser (Safari). Given that's part of the ticket criteria, it probably needs to be fixed?

Screen Shot 2022-03-02 at 9 42 55 AM

conf/drupal/config/views.view.all_documents.yml
arthurbaghdas: Yeah this is site specific. /download is a special URL in the site. So i think leaving it with hardcoded env URL is fine. At least we used same approach for different views as well.

arthurbaghdas:
> This looks close! Config changes look good (though I didn't do a line-by-line read). While I can see the new column, I'm not getting a downloaded file. Instead, after the batch completes I see the CSV in my browser (Safari). Given that's part of the ticket criteria, it probably needs to be fixed?
> 
> Screen Shot 2022-03-02 at 9 42 55 AM

This issue is now fixed.

deviantintegral:
Great! Process wise, should I be testing this since I reported the bug, or is there someone who will do cross-browser testing to make sure it didn't somehow break it elsewhere?

arthurbaghdas:
> Great! Process wise, should I be testing this since I reported the bug, or is there someone who will do cross-browser testing to make sure it didn't somehow break it elsewhere?

arthurbaghdas:
> Great! Process wise, should I be testing this since I reported the bug, or is there someone who will do cross-browser testing to make sure it didn't somehow break it elsewhere?

I think you can post this in the ticket.

beto-aveiga (approved):
LGTM

PR #1370, opened 2022-02-28
Removes crufty HQ2 content

bryanhirsch:
**Description:**
This is a static website that was created when Massachusetts was
bidding on Amazon's second headquarters. Now it's outdated and
unmaintained, seems like it's time to delete it.

**To Test:**
- [ ] Before: Go to mass.gov/hq2, see proposal website
- [ ] After: Go to mass.gov/hq2, now you should get a 404 error


weitzman:
Thanks for the cleanup! PRs from forks always fail, as they are untrusted. So this is fine to merge.

PR #1384, opened 2022-03-04
DP-24227 fix paragraphs with conditional fields

beto-aveiga:
**Description:**

DP-24227

**To Test:**

---


weitzman (commented):
I am seeing two patches with the same node id number in the filename. Thats not usually how  we work. Please change one of the filenames or explain whats up here. Lets post a dev version at https://www.drupal.org/node/2902164 and even upgrade us to dev if thats helpful (optional).

Also, the most recent comment at https://www.drupal.org/node/2902164 suggests there is a bug. If we think that bug is real and will affect us, lets fix before merge.

PR #1385, opened 2022-03-04
DP-24238 adding foreach on IE11

beto-aveiga:
**Description:**

DP-****

**To Test:**

---


docroot/themes/custom/mass_theme/templates/layout/html.html.twig
```diff
@@ -32,16 +32,101 @@
 
     <script>
       document.prefetchAlertsData = {};
+
+      // See: https://gist.github.com/nekman/297ebda63d6b00380058cbb0114296aa#file-polyfill-js-L586
```
clairesunstudio: The polyfills look good and thanks for adding the comments! I wonder if we should move these custom polyfills into its own file and add it to mass_theme.libraries.yml or in alerts-block.js, instead of adding them directly to html.html.twig?

clairesunstudio: per discussion with Luis, although these are IE11 polyfills for alerts-block but they also make sense to be in the global context

> ... if we need IE11 support and we need them as soon as possible, the answer is yes, there is no other way I know.
Anyhow, moving those polyfills makes sense, although on the other hand those polyfills are adding things in the global context.

clairesunstudio (approved):

clairesunstudio:
pulling the scripts into separate files make the html template more organized. 
Also no console errors are displayed in IE11 with the latest deployment
Screen Shot 2022-03-07 at 1 44 22 PM

PR #1391, opened 2022-03-09
Dp 24285 topic page heading label followup

beto-aveiga:
**Description:**

DP-24285

**To Test:**

---


weitzman:
I dont think this is really the root cause fix but its not that important.

PR #1395, opened 2022-03-10
DP-24242 Stem search queries in All Collections view

deviantintegral:
**Description:**
This is very rough, but works! Before this branch searching for `content filters`:

Screen Shot 2022-03-09 at 8 17 17 PM

On this branch:

Screen Shot 2022-03-09 at 8 17 01 PM

It'd been a while since I wrote a Views plugin, so of course I started with "let's decorate the string filter plugin". Then I ran into the Views class hierarchy twice! Once, to make it work for the Combine filter, then again to make it work for views autocomplete filters. So really, I'd like this to be a contrib module, but it would require some work to make it work with all the different classes in the tree.

DP-24242

**To Test:**
- [ ] Play around with searches and compare to production. Let's see if they're better or worse.

---

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -253,7 +253,8 @@
         "phlak/semver": "^2.0",
         "symfony/dom-crawler": "^3.4 || ^4",
         "typhonius/acquia-php-sdk-v2": "~1.1",
-        "vlucas/phpdotenv": "^2.4"
+        "vlucas/phpdotenv": "^2.4",
+        "wamania/php-stemmer": "^3.0"
     },
     "require-dev": {
         "behat/behat": "^3",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "a39ebcb252caecdcb54eb169c850b945",
+    "content-hash": "bf5ef72269a826c17771a910ce7e4e89",
     "packages": [
         {
             "name": "akamai-open/edgegrid-auth",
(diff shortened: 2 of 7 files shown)


composer.json
```diff
@@ -253,7 +253,8 @@
         "phlak/semver": "^2.0",
         "symfony/dom-crawler": "^3.4 || ^4",
         "typhonius/acquia-php-sdk-v2": "~1.1",
-        "vlucas/phpdotenv": "^2.4"
+        "vlucas/phpdotenv": "^2.4",
+        "wamania/php-stemmer": "^3.0"
```
deviantintegral, line 257: `drupal/porter_stemmer` is hardcoded to require search module, and this package has many hundreds of thousands of installs.

deviantintegral:
Unrelated, but when I exported config, I got _many_ other unrelated changes:

```
 ...entity_form_display.node.topic_page.default.yml |  1 +
 .../core.entity_view_display.node.news.default.yml |  2 +-
 ...re.entity_view_display.node.topic_page.card.yml |  1 +
 ...entity_view_display.node.topic_page.default.yml |  1 +
 ...isplay.node.topic_page.link_and_description.yml |  1 +
 ...tity_view_display.node.topic_page.link_only.yml |  1 +
 ....entity_view_display.node.topic_page.teaser.yml |  1 +
 ...e.entity_view_display.node.topic_page.token.yml |  1 +
 conf/drupal/config/mass_utility.settings.yml       |  2 +-
 conf/drupal/config/views.view.content.yml          |  8 ++++----
 .../config/workflows.workflow.topic_page.yml       | 22 +++++++++++-----------
 12 files changed, 24 insertions(+), 22 deletions(-)
```

It doesn't look like this is an artifact of local development settings or modules. Is there a CI job in the mix to make sure config imports cleanly?

deviantintegral:
Replaced by https://github.com/massgov/openmass/pull/1399

PR #1404, opened 2022-03-18
Dp 24336 past event not disappearing from org page

beto-aveiga:
**Description:**
Adds a tag to events paragraphs to disable cache whenever an event changes.

DP-24336

**To Test:**

---


weitzman:
Needs testing steps. Also needs a re-run as Danger test is not passing for some reason.

beto-aveiga:
@weitzman The only way I came up to test events on org pages was by changing the query for the upcoming events, using the Drupal state to know when to change the query, instead of changing the time of the request or injecting the time on the Events Manager, as that would be more complex and I'm not clear how to proceed.

Let me know if you have a better approach for this. Thanks.

composer.lock
```diff
@@ -9290,17 +9290,17 @@
         },
         {
             "name": "drupal/scheduled_transitions",
-            "version": "2.2.1",
+            "version": "2.2.0",
```
weitzman: We dont want to downgrade this. We need the patch release.

weitzman (commented):
Sorry, I meant "testing steps" like steps for manually testing it. But an automated test is even better. Lets call upon @deviantintegral and @tom-fleming for input on mocking time service for functional tests like those used on this project. I see https://www.drupal.org/project/datetime_testing which looks like a more polished version of what you have done with State.

composer.lock
beto-aveiga: > Sorry, I meant "testing steps" like steps for manually testing it. But an automated test is even better. Lets call upon @deviantintegral and @tom-fleming for input on mocking time service for functional tests like those used on this project. I see https://www.drupal.org/project/datetime_testing which looks like a more polished version of what you have done with State.

No problem, Moshe. The `datetime_testing` module looks promising, I can try that for sure, but I'm interested in @tom-fleming  and @deviantintegral's opinion. Thanks!

deviantintegral: Nice find! I only looked at a few files, but they look good. Multiple independent implementations [reaching the same conclusion](https://git.drupalcode.org/project/datetime_testing/-/blob/8.x-1.x/src/TestTime.php#L72-81) is nice to see.

docroot/modules/custom/mass_content/src/EventManager.php
```diff
@@ -16,11 +16,26 @@
...
+  public function __construct(EntityTypeManagerInterface $entityTypeManager, StateInterface $state) {
     $this->entityTypeManager = $entityTypeManager;
+    $this->upcomingEndDateRelative = 'today';
+
+    // $time is "now" only when running tests because we can't wait
+    // one day to test if the event is gone. Injecting the relative
+    // upcoming end date is complex, as well as changing the request time.
+    if ($state->get('UPCOMING_END_DATE_RELATIVE', FALSE)) {
+      $this->upcomingEndDateRelative = 'now';
+    }
```
deviantintegral: I'm not sure we want to use State directly in this class. Once we mock the time service, will this still be needed? It may be clearer to to have this string be a parameter in the constructor.

I can think further on this if I'm not making sense, let me know!

docroot/modules/custom/mass_content/src/EventManager.php
```diff
@@ -82,10 +97,9 @@ private function getPastQuery(NodeInterface $parent) {
    */
   private function getUpcomingQuery(NodeInterface $parent) {
     $query = $this->getBaseQuery($parent);
-    $today = new \DateTime('today', new \DateTimezone('America/New_York'));
-    $today->setTimezone(new \DateTimezone('UTC'));
-
-    $query->condition('field_event_date.end_value', $today->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT), '>');
+    $upcomingDateEndDate = new \DateTime($this->upcomingEndDateRelative, new \DateTimezone('America/New_York'));
```
deviantintegral: Should this be `DrupalDateTime`?

Should we use the site timezone instead of hardcoding America/New_York?

docroot/modules/custom/mass_content/src/EventsRendererOrgPages.php
```diff
@@ -0,0 +1,127 @@
+<?php
```
deviantintegral, line 1: A suggestion: use strict types by default in all new classes?

https://architecture.lullabot.com/adr/20211201-use-strict-types-in-php-code/

docroot/modules/custom/mass_content/src/EventsRendererOrgPages.php
```diff
@@ -0,0 +1,127 @@
...
+  /**
+   * Returns a render array with the upcoming or past events.
+   */
+  public function render() {
+    // Gather events associated with this node.
+    $eventManager = \Drupal::service('mass_content.event_manager');
+    $upcoming = $eventManager->hasUpcoming($this->parentNode);
+    $render = $upcoming ? $this->nextEvents() : $this->pastEventsLink();
+    // To update if any event changes.
+    $render['#cache']['tags'][] = 'node_list:event';
```
deviantintegral, line 55: 🤔 Do we want to add the tag manually, or should we use the an existing API like https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21Renderer.php/function/Renderer%3A%3AaddCacheableDependency/9.4.x or `CacheableMetadata`?

docroot/modules/custom/mass_content/src/EventManager.php
beto-aveiga: Makes total sense, but I don't see an easy way to implement that.
I can also have a new parameter in the constructor for upcomingEndDateRelative, but then I will also have to alter the service parameters only when testing, which can be done enabling another module only for testing... but that seems also complicated.
I don't know how we would implement it without adding up more complexity.

docroot/modules/custom/mass_content/src/EventManager.php
beto-aveiga: I will change the references of DateTimezone to DrupalDateTime. Great suggestions as using DrupalDateTime there is no need to set the timezone when using relative dates. Thanks!

docroot/modules/custom/mass_content/src/EventsRendererOrgPages.php
beto-aveiga: Great suggestion, I've read it before but I forget it. This should raise a warning on PHPCS.

docroot/modules/custom/mass_content/src/EventsRendererOrgPages.php
beto-aveiga: For this particular case: yes. Any event can change and the org page should be invalidated. 
If an org page doesn't have any event, it wouldn't have this tag. Then if an event changes by adding a relationship to that org page, the org page cache entry will not be cleared.
Does it make sense?

deviantintegral: I was more referring to something like:

```php
    (new CacheableMetadata())->addCacheTags(['node_list:event'])
      ->applyTo($render);
```

On the one hand, it's more verbose, but it also means it reduces strings you can typo from three (`#cache`, `tags`, `node_list:event`) to just `node_list:event`.

Let's keep it as is for this PR. Thanks for thinking through it!

docroot/modules/custom/mass_content/src/EventsRendererOrgPages.php
deviantintegral: Filed https://massgov.atlassian.net/browse/DP-24449

PR #1412, opened 2022-03-21
DP-24353 Update drupal/devel so webprofiler works

deviantintegral:
**Description:**
Now with less FATALs! Fixes:

```
$ ddev drush en webprofiler -y
PHP Fatal error:  Class Drupal\webprofiler\Entity\Decorators\Config\ConfigEntityStorageDecorator contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (Drupal\Core\Entity\EntityStorageInterface::getEntityClass) in /var/www/html/docroot/modules/contrib/devel/webprofiler/src/Entity/Decorators/Config/ConfigEntityStorageDecorator.php on line 18

Fatal error: Class Drupal\webprofiler\Entity\Decorators\Config\ConfigEntityStorageDecorator contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (Drupal\Core\Entity\EntityStorageInterface::getEntityClass) in /var/www/html/docroot/modules/contrib/devel/webprofiler/src/Entity/Decorators/Config/ConfigEntityStorageDecorator.php on line 18
 [warning] Drush command terminated abnormally.
Failed to run drush en webprofiler: exit status 255
```

DP-24353

**To Test:**
- [ ] Builds passing!

---


deviantintegral:
We will need another PR after this, as it appears there is a conflict with `entity_hierarchy`. Since that module is important for end-user functionality, it will need real testing as compared to this development dependency.

tom-fleming:
This update looks good to me. I guess I would consider this a soft update at the moment until we address the patch issue, but I agree that this is an improvement.

PR #1420, opened 2022-03-23
Add and configure environment indicator module

weitzman:
- This also updates scheduled_transition module, to fix DP-24364
- Add a patch for debugging the error message thats spamming watchdog https://www.drupal.org/project/drupal/issues/3013802#comment-14456001.
- This stops our scaffolding from overwriting our improved development.services.yml (disables cache headers), and causing a local git diff. However, that fix caused our CI to start failing, because some tests expect check for these headers. So, our CI now adds back core's default development.services.yml file during the build job.

DP-24094

**To Test:**
- [ ] Confirm that header color and environment name are different on admin pages between this PR's Tugboat and [Feature3](https://massgovfeature3.prod.acquia-sites.com/admin/)

image

image

image

---

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,12 +30,12 @@ references:
     restore_cache:
       name: Restore Composer Cache
       keys:
-        - site-composer-v12-{{ checksum "composer.lock" }}
+        - site-composer-v13-{{ checksum "composer.lock" }}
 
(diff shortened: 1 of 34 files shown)


clairesunstudio:
Revert production nav color, and set non-production nav color to light gray
Screen Shot 2022-03-29 at 10 14 09 AM

PR #1422, opened 2022-03-24
Remove illustrated header min-height

clairesunstudio:


clairesunstudio:
Tugboat environments have been having issues. Deploying this to the DEV feature env to run backstop

PR #1286, opened 2022-01-12
DP-23508: Upgrade to PHP 8

arthurbaghdas:
**Description:**
Reroll:

```console
ddev composer upgrade -W \
  drupal/diff \
  drupal/acquia_connector \
  drupal/geocoder \
  drupal/geofield \
  drupal/memcache \
  drupal/monolog \
  palantirnet/drupal-rector \
  geocoder-php/open-cage-provider \
  monolog/monolog \
  typhonius/acquia-php-sdk-v2 \
  chillerlan/php-qrcode
```

**Jira:** (Skip unless you are MA staff)
DP-23508

**To Test:**
- [ ] Add steps to test this feature

---

[Peer Review Checklist](https://github.com/massgov/openmass/blob/develop/docs/peer_review_checklist.md)

PR #1426, opened 2022-03-29
DP-23508 Fix test failures in mass_alerts module

deviantintegral:
**Description:**
Pulling out changes in `mass_alerts` from https://github.com/massgov/openmass/pull/1286 that look to be unrelated to the PHP 8 upgrade.

DP-23508

**To Test:**
- [ ] Tests passing

---


docroot/modules/custom/mass_alerts/tests/src/ExistingSite/EmergencyAlertsTest.php
```diff
@@ -91,7 +85,6 @@ public function testEmergencyAlertResponseSitewide() {
     $this->assertStringContainsString($alert_message_text, $page->getText());
 
     $headers = $session->getResponseHeaders();
-    $this->assertStringContainsString('max-age=60', $headers['Cache-Control'][0]);
```
weitzman: We no longer test max-age? I might have written this but I dont recall why.

deviantintegral: The commit is over at https://github.com/massgov/openmass/pull/1286/commits/3bf807f5bebcbf7d0b37ff4729ec2576b4a55b19

I'll try reverting this and the logout call and see if tests still pass.

deviantintegral:
Tests pass locally for me with this reverted, but I'd like to double check in CI. Once the above commit has passed, I will push a revert to confirm, and then we can likely close this and I'll revert over in the other branch.

PR #1429, opened 2022-03-29
DP-23508 Move development dependencies to require-dev

deviantintegral:
**Description:**
I noticed several dependencies in require instead of require-dev. This PR moves and updates them.

Normally I'd put stage_file_proxy as a development dependency too, but it's enabled on production so we'll want to keep it for now.

DP-23508

**To Test:**
- [ ] Do we have a test that verifies the site bootstraps when running `composer install --no-dev`?

---


deviantintegral:
Once this is approved I will resolve conflicts.

weitzman:
In CircleCI, we build and push our codebase to Acquia. I think this will just build and push the development build. So this change has no practical effect yet AFACT.

typhonius/acquia-php-sdk-v2 is part of deploy process, so we always need it. upgrade_rector has a report thats useful on Prod.

deviantintegral:
> typhonius/acquia-php-sdk-v2 is part of deploy process, so we always need it.

The SDK is still in require. It just needed a `,` removed because JSON.

> upgrade_rector has a report thats useful on Prod.

This module isn't in core.extension.yml. Is it being enabled somewhere else for production?

weitzman:
We enable in config when we get within 6-12 months from a new major Drupal release. That day is coming soon.

deviantintegral:
I've moved drupal/upgrade_rector back to require, and fixed conflicts from develop.

weitzman:
So, whats the intention here? Should the build and build_with_mayflower jobs in ~/.circleci/config.yml use `composer install --no-dev`? That way only local env would be using the dev packages. If we do that, we need to keep friends-of-behat/mink-debug-extension in require as thats needed for test running.

I bring this up because those 2 jobs assemble a build that is then pushed to Acquia git (see push_acquia job in Circle). If we dont change those jobs then we are still sending dev and non-dev packages to Acquia.

deviantintegral:
> Should the build and build_with_mayflower jobs in ~/.circleci/config.yml use composer install --no-dev

I think not sending dev dependencies to production is a good idea, but one that requires actual testing so I think it could be a ticket to throw in the backlog. A common pattern I've used is to have dev dependencies on locals and test builds, but not on interactive environments like Tugboat or Acquia.

As is, this is really a "documentation" PR in that it makes clear what is needed for development purposes and what isn't. When a testing dependency is in require, it's a signal to new developers that somehow that code is actually needed to run the site.

PR #1436, opened 2022-03-31
Dp 24334 investigate conditions collection ids can be a scalar

beto-aveiga:
**Description:**

DP-****

**To Test:**

---


deviantintegral (approved):
Static review looks good to me!

Can we also do this in `\Drupal\mass_views\Plugin\Action\ChangeCollections`? If so, it doesn't matter to me if we do it here or in a new PR. But if we can do it, we should get it done before closing the Jira ticket.

PR #1438, opened 2022-04-01
DP-23508 Support switching PHP versions during Acquia deployments

deviantintegral:
**Description:**
The error I mentioned earlier was not related to CircleCI caches. Instead, it was:

1. I had switched the PHP version in stage's Acquia UI.
2. The deployment then tried to import the database using the old PHP 7.x code, as it deploys the new code _after_ importing the database.

I was able to test this by switching the deploy step around, and the deployment succeeded at https://app.circleci.com/pipelines/github/massgov/openmass/18330/workflows/d7ba3185-4d3b-4411-9aa5-d633323e9513/jobs/75632

However, we may be switching environments over time, and eventually we'll want PHP 8.1 and so on. This PR:

1. Switches the PHP version to the declared version after turning on maintenance mode, but before deploying code.
2. I noticed a possible NULL method call with `logger()`, and since I was adding new code with log messages I cleaned it up.

DP-23508

**To Test:**
- [x] Tests passing
- [x] Deploy this branch to stage, and it should properly downgrade to PHP 7.4.
- [x] Merge this branch into the PHP 8 branch, and when deployed it should go back to PHP 8.0.

---


drush/Commands/DeployCommands.php
```diff
@@ -22,6 +24,13 @@ class DeployCommands extends DrushCommands implements SiteAliasManagerAwareInter
 
   use SiteAliasManagerAwareTrait;
 
+  /**
+   * Set the PHP version to use when deploying to Acquia environments.
+   *
+   * @const string
+   */
+  public const PHP_VERSION = '7.4';
```
deviantintegral, line 32: This is the one thing I'm not sure about. I'm a little worried that in the future this constant could be lost. I'd thought about putting it in a separate YAML file somewhere, but that adds complexity.

weitzman (approved):
LGTM. I'm fine merging this into develop or merging it into the PHP8 branch. Your call.

deviantintegral:
Let's put this into develop, so we have the code to switch back an environment to 7.4 if needed. Thanks!

PR #1448, opened 2022-04-07
DP-24477: Add focal point to banner images on certain content types

arthurbaghdas:
**Description:**

DP-24477

**To Test:**

---


docroot/modules/custom/mass_content/mass_content.services.yml
```diff
@@ -13,3 +13,6 @@ services:
     tags:
       - { name: event_subscriber }
     weight: 200
+  mass_content.image_style_warmer:
```
deviantintegral, line 16: Have you seen https://www.drupal.org/project/warmer ? This could be nice as a plugin contributed to that project.

docroot/modules/custom/mass_content/mass_content.module
```diff
@@ -887,3 +888,25 @@ function mass_content_entity_bundle_info_alter(array &$bundles): void {
...
 
 }
+
+/**
+ * Implements hook_ENTITY_insert() for crop entity.
+ */
+function mass_content_crop_insert(EntityInterface $entity) {
+  if ($entity->get('entity_type')->value == 'file') {
+    if ($file = File::load($entity->get('entity_id')->value)) {
+      \Drupal::service('mass_content.image_style_warmer')->warmUp($file);
```
deviantintegral, line 1078: 👏 love the thin module code here!

docroot/modules/custom/mass_content/src/ImageStylesWarmer.php
```diff
@@ -0,0 +1,98 @@
...
+   */
+  public function warmUp(FileInterface $file) {
+    $initialImageStyles = ['action_banner_large_focal_point', 'hero1600x400_fp'];
+    $this->doWarmUp($file, $initialImageStyles);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function doWarmUp(FileInterface $file, array $image_styles) {
```
deviantintegral: Since this is only called by warmUp above, how about we make it private? Also, add `: void` as a return type hint?

```suggestion
  private function doWarmUp(FileInterface $file, array $image_styles): void {
```

docroot/modules/custom/mass_content/src/ImageStylesWarmer.php
```diff
@@ -0,0 +1,98 @@
...
+  public function __construct(EntityTypeManagerInterface $file_storage, ImageFactory $image_factory, EntityTypeManagerInterface $image_style_storage) {
+    $this->file = $file_storage->getStorage('file');
+    $this->image = $image_factory;
+    $this->imageStyles = $image_style_storage->getStorage('image_style');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function warmUp(FileInterface $file) {
```
deviantintegral: `: void` return type hint?

```suggestion
  public function warmUp(FileInterface $file): void {
```

docroot/modules/custom/mass_content/src/ImageStylesWarmer.php
```diff
@@ -0,0 +1,98 @@
+<?php
```
deviantintegral, line 1: Should we add `declare(strict_types=1);` here, since this is a brand new file?

```suggestion
<?php
declare(strict_types=1);
```

docroot/modules/custom/mass_content/src/ImageStylesWarmer.php
```diff
@@ -0,0 +1,98 @@
...
+   */
+  public function doWarmUp(FileInterface $file, array $image_styles) {
+    if (empty($image_styles) || !$this->validateImage($file)) {
+      return;
+    }
+
+    /* @var \Drupal\Core\Image\Image $image */
+    /* @var \Drupal\image\Entity\ImageStyle $style */
+
+    // Create image derivatives if they not already exists.
```
deviantintegral:
```suggestion
    // Create image derivatives if they do not already exist.
```

docroot/modules/custom/mass_content/src/ImageStylesWarmer.php
```diff
@@ -0,0 +1,98 @@
...
+  public function warmUp(FileInterface $file) {
+    $initialImageStyles = ['action_banner_large_focal_point', 'hero1600x400_fp'];
+    $this->doWarmUp($file, $initialImageStyles);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function doWarmUp(FileInterface $file, array $image_styles) {
+    if (empty($image_styles) || !$this->validateImage($file)) {
```
deviantintegral: How about we reverse the order of these conditionals, to match the method arguments? "First validate the file, then check the image styles".

```suggestion
    if (!$this->validateImage($file) || empty($image_styles)) {
```

docroot/modules/custom/mass_content/src/ImageStylesWarmer.php
```diff
@@ -0,0 +1,98 @@
...
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+   */
+  public function __construct(EntityTypeManagerInterface $file_storage, ImageFactory $image_factory, EntityTypeManagerInterface $image_style_storage) {
+    $this->file = $file_storage->getStorage('file');
+    $this->image = $image_factory;
+    $this->imageStyles = $image_style_storage->getStorage('image_style');
+  }
+
+  /**
+   * {@inheritdoc}
```
deviantintegral: This and a few further methods say they inherit docs, but this class doesn't extend or implement anything.

docroot/modules/custom/mass_content/src/ImageStylesWarmer.php
```diff
@@ -0,0 +1,98 @@
...
+      if (!file_exists($derivative_uri)) {
+        $style->createDerivative($image_uri, $derivative_uri);
+      }
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function validateImage(FileInterface $file) {
```
deviantintegral:
```suggestion
  private function validateImage(FileInterface $file): bool {
```

docroot/modules/custom/mass_content/mass_content.install
```diff
@@ -1009,3 +1009,19 @@ function mass_content_update_8016(&$sandbox) {
     }
   }
 }
+
+/**
+ * Delete field_bg_narrow from Organization content type.
```
deviantintegral: 🤔 is this actually needed? On my local, if I disable this update hook, config import still works to delete the field.

It's not actually _breaking_ anything to have this, this is more a curiosity on my part.

deviantintegral (approved):
A few minor suggestions. IMO the only thing that actually blocks this from merging is the incorrect docblocks (unless the warmer class is actually supposed to extend or implement something). Feel free to review and take them as you see fit. Nice work!

docroot/modules/custom/mass_content/mass_content.services.yml
arthurbaghdas: Agree, will take a look, thanks

docroot/modules/custom/mass_content/src/ImageStylesWarmer.php
arthurbaghdas: What's the difference?

docroot/modules/custom/mass_content/mass_content.install
arthurbaghdas: Right, removed the update hook

docroot/modules/custom/mass_content/src/ImageStylesWarmer.php
arthurbaghdas: Added some info.

docroot/modules/custom/mass_content/src/ImageStylesWarmer.php
deviantintegral: I was mainly thinking about this from a "reading code" perspective. The method is "file, image styles", and it's generally logical to validate parameters in the same order they are in the method signature.

_however_ there are times it's good to break from that, such as if `validateImage()` is much slower, so by checking other conditions you can be a tad faster. My guess is it doesn't matter much in this case performance wise.

docroot/modules/custom/mass_content/mass_content.module
```diff
@@ -1083,3 +1090,30 @@ function mass_content_crop_update(EntityInterface $entity) {
     }
   }
 }
+
+/**
+ * Sets the sort_bef_combine URL parameter to sort Collection views.
+ */
+function mass_content_set_default_sort_from_collection(Term $collection) {
+  if (!$collection) {
```
deviantintegral, line 1098: Is this possible, given the `Term` typehint above? It looks like `mass_content_get_collection_from_current_page` can return `null`, so we either need to move this check to  `mass_content_views_pre_view` or add `= NULL` to the function signature.

deviantintegral (approved):
One quick note, but feel free to merge after fixing it. Nice!

docroot/modules/custom/mass_content/mass_content.module
arthurbaghdas: This is not related to my ticket, this came from the develop branch, when I merged today. Do you think I should change this in the scope of this ticket? @deviantintegral

arthurbaghdas: Leaving this as is, since there are several places where this function is used.

PR #1428, opened 2022-03-29
DP-23508 Remove CloudFlare module as we use Akamai

deviantintegral:
**Description:**
Cleaning up a module that is no longer installed on the site.

**Jira:** (Skip unless you are MA staff)
DP-23508

**To Test:**
- [ ] Static tests passing

---

[Peer Review Checklist](https://github.com/massgov/openmass/blob/develop/docs/peer_review_checklist.md)

PR #1450, opened 2022-04-08
Update entity_hierarchy and its patch to allow enabling of webprofiler.

weitzman:
- The deeper problem is unsolved in Drupal Views so lets just fix the obstacle at hand (cant enable webprofiler module).
- This PR also adds https://github.com/marketplace/actions/composer-diff so we can more easily review composer.lock changes in PRs, and spot bad rerolls. Commenting is not running yet but lets merge this PR while that gets worked out at https://github.com/IonBazan/composer-diff-action/issues/7
- Incorporated a reroll of #1428 so we dont get double conflicts

DP-24363

**To Test:**
- [ ] Try to enable webprofiler. If you can see its toolbar at bottom of the page, its good.

---

### DIFF
--- a/.github/workflows/composer-diff.yml
+++ b/.github/workflows/composer-diff.yml
@@ -18,7 +18,8 @@ jobs:
         id: composer_diff # To reference the output in comment
         uses: IonBazan/composer-diff-action@v1
 
-      - uses: marocchino/sticky-pull-request-comment@v2
+      - name: Post comment
+        uses: marocchino/sticky-pull-request-comment@v2
         # An empty diff result will break this action.
         if: ${{ steps.composer_diff.outputs.composer_diff_exit_code != 0 }}
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/composer-diff.yml
```diff
@@ -0,0 +1,34 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0 # Required to make it possible to compare with PR base branch
+
+      - name: Generate composer diff
+        id: composer_diff # To reference the output in comment
+        # Use master branch until next release.
+        uses: IonBazan/composer-diff-action@master
```
deviantintegral, line 20: 🤔 should this be pinned to a version?

deviantintegral (approved):
Tested locally and webprofiler enables without issues. Nice! My one comment isn't a merge blocker unless you think it is.

.github/workflows/composer-diff.yml
weitzman: I think we should proceed as is. master has fixes to code and docs that we are using here.

PR #1451, opened 2022-04-11
DP-24527: Problem with link in Service Page banner not formating correctly

arthurbaghdas:
**Description:**

DP-24527

**To Test:**

---

### DIFF
--- a/changelogs/DP-24527.yml
+++ b/changelogs/DP-24527.yml
@@ -1,0 +1,41 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
+# #############################################################################
+# You can use one of the following types:
+#  - Added: For new features.
+#  - Changed: For changes to existing functionality.
+#  - Deprecated: For soon-to-be removed features.
+#  - Removed: For removed features.
+#  - Fixed: For any bug fixes.
+#  - Security: In case of vulnerabilities.
+#
+# Format
+# #############################################################################
+# The format is crucial. Please follow the examples below. For reference, the requirements are:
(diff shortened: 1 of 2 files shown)


changelogs/DP-24527.yml
```diff
@@ -0,0 +1,41 @@
...
+# Changed:
+#  - description: Automating the release branch.
+#    issue: DP-10166
+#  - description: Second change item that needs a description.
+#    issue: DP-19875
+#  - description: Third change item that needs a description along with an issue.
+#    issue: DP-19843
+#
+Fixed:
+  - description: Describe the change. If you need multiple lines, start the first line with the following "|-" characters.
```
deviantintegral: Looks like the description is missing.

docroot/themes/custom/mass_theme/templates/content/node--service-page.html.twig
```diff
@@ -70,7 +70,7 @@
   {% if node.field_log_in_links is not empty %}
     {% for item in node.field_log_in_links %}
       {% set button = {
-          "href": item.uri,
+          "href": item.url,
```
deviantintegral, line 73: One character makes all the difference!

deviantintegral (commented):
For sure the changelog text needs fixing, but the code change looks good.

Given this bug made it to production, should we write a functional test too?

changelogs/DP-24527.yml
arthurbaghdas: Thank you, missed that info. I don't think we need a specific test for this case.

deviantintegral (approved):

PR #1455, opened 2022-04-12
DP-24557 Fix null method call when entities are deleted

deviantintegral:
**Description:**

This PR fixes an error on locals and one that could also affect production if a media or file entity is deleted. I suggest reviewing like this:

* The first few commits [apply some basic cleanups](https://github.com/massgov/openmass/pull/1455/files/9616bec5c473bbd22d67e1f0d83e64720c3dfce1..0f2a411e74673f9415dc193d4908510366d0d76e) for incorrect return types, equality checks, new PHP syntax and missing required extensions.
* [This commit](https://github.com/massgov/openmass/pull/1455/commits/a3caa5ced524f0f93eb7e0169eb479c329138d1f) actually fixes the bug, though later on I flip the conditional around.
* [Then I refactored to a class so I could write a test](https://github.com/massgov/openmass/pull/1455/files/a3caa5ced524f0f93eb7e0169eb479c329138d1f..303704ff67bfa33cebb83d4241729d3e7c576a8c). I'm glad I did because I discovered that file entities were subject to the same bug, and I also found an incorrect comment.
  * I know, it's a lot of mocking, but I only wrote a test for the _failure_ state so I went for performance over an existing site test.

DP-24557

**To Test:**
- [ ] Static review and tests passing

---


docroot/modules/custom/mass_schema_metatag/mass_schema_web_page/mass_schema_web_page.module
```diff
@@ -31,62 +32,12 @@ function mass_schema_web_page_tokens($type, $tokens, array $data, array $options
...
-                // Get the uploaded file values.
-                $files = $media_entity->get('field_upload_file')->getValue();
-                foreach ($files as $file) {
-                  if (!empty($file['target_id'])) {
-                    $file_id = $file['target_id'];
-                    // Load the file with the fid.
-                    $file = File::load($file_id);
-                    // Get the file uri and the url path to the file.
-                    $uri = $file->getFileUri();
-                    $filepaths[] = file_create_url($uri);
```
deviantintegral, line 81: `file_create_url` is gone in D10 so I converted this over.

docroot/modules/custom/mass_schema_metatag/mass_schema_web_page/src/DecisionTokenRenderer.php
```diff
@@ -0,0 +1,167 @@
...
+   * @param \Drupal\Core\Field\EntityReferenceFieldItemListInterface $field
+   *   The field referencing documents.
+   * @param array $replacements
+   *   The current array of token replacements.
+   * @param string $original
+   *   The original token name.
+   */
+  private function getEntityReferenceReplacements(EntityReferenceFieldItemListInterface $field, array &$replacements, string $original): void {
+    $filepaths = $this->getFilepaths($field);
+    $replacements[$original] = json_encode($filepaths, JSON_THROW_ON_ERROR);
```
deviantintegral, line 96: While this should never fail since we control the data, as is it would _silently_ fail so I added this flag.

docroot/modules/custom/mass_schema_metatag/mass_schema_web_page/src/DecisionTokenRenderer.php
```diff
@@ -0,0 +1,167 @@
...
+   */
+  private function getUploadedFileValues(\Drupal\media\MediaInterface $media_entity, array &$filepaths): void {
+    foreach ($media_entity->get('field_upload_file')->getValue() as $file) {
+      if (!empty($file['target_id'])) {
+        $file_id = $file['target_id'];
+        /** @var \Drupal\file\FileInterface $file_entity */
+        $file_entity = $this->entityTypeManager->getStorage('file')
+          ->load($file_id);
+
+        // Skip file entities that have been deleted underneath their media
```
deviantintegral, line 156: need to fix this

deviantintegral: nevermind, I misread this does correctly mention file entities.

tom-fleming:
LGTM!

tom-fleming (approved):

PR #1459, opened 2022-04-14
Dp 24485/header hamburger menu allow google translation

clairesunstudio:
Fixed:
  - description: Make header hamburger menu translatable by Google.
    issue: DP-24485

Link to MF https://github.com/massgov/mayflower/pull/1618


docroot/themes/custom/mass_theme/templates/block/block--mainnavigation-3.html.twig
```diff
@@ -43,7 +43,15 @@
               class="ma__header__hamburger__menu-button js-header-menu-button">
         <span class="ma__header__hamburger__menu-icon"></span>
```
clairesunstudio, line 44: why do we duplicate the main menu twig template in openmass instead of using the component from mayflower?

ygannett: @clairesunstudio As far as I know, Drupal cannot pull some MF templates directly based on where components locate in the Drupal structure.  This causes a necessary of setting up duplicate templates to Mayflower's in Drupal.

ygannett (approved):
LGTM.

PR #1462, opened 2022-04-18
DP-24545 add option for temporary unpublished access links to last for 6 weeks

beto-aveiga:
**Description:**

DP-24545

**To Test:**

---

### DIFF
--- a/changelogs/DP-24545.yml
+++ b/changelogs/DP-24545.yml
@@ -1,0 +1,41 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
+# #############################################################################
+# You can use one of the following types:
+#  - Added: For new features.
+#  - Changed: For changes to existing functionality.
+#  - Deprecated: For soon-to-be removed features.
+#  - Removed: For removed features.
+#  - Fixed: For any bug fixes.
+#  - Security: In case of vulnerabilities.
+#
+# Format
+# #############################################################################
+# The format is crucial. Please follow the examples below. For reference, the requirements are:
+#  - All 3 parts are required and you must include "Type", "description" and "issue".
+#  - "Type" must be left aligned and followed by a colon.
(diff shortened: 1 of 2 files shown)


deviantintegral (approved):
Looks good!

I don't see an exact version of this upstream at https://www.drupal.org/project/issues/access_unpublished?text=lifetime&status=All&priorities=All&categories=All&version=All&component=All. The closest is https://www.drupal.org/project/access_unpublished/issues/2942249 but that could also be problematic given a user could effectively make an unlimited token.

It seems to me a good solution would be to make it so that the duration is configurable in the module settings. If you agree, how about we file a feature request upstream?

As well, this alter function is getting pretty unwieldy! If I pull out and add in a few comments, this one function is:

```php
  // Update access unpublished generate token button label.
  // Remove 'Unlimited' option for Temporary Unpublished Access.
  // Add help text link for Scheduling options.
  // Add help text link for Temporary Unpublished Access.
  // Add 6 weeks unpublished access duration.
  // Changed the help text for the campaign landing page for the unpublish date only.
  // Add Organization Stakeholders field to 'advanced' vertical tabs group.
  // Attach the fields to group.
  // Hide the preview button when a node is unsaved.
  // Add link for contacting a user in the publish area.
  // Add Node ID to meta details.
  // Hide the publish_state field on the node/edit form.
  // Hide the unpublish_state field on the node/edit form.
  // For Intended Audience Field update the default value to 'Not Set'.
  // Invalidate node tags when submitting.
```

In particular, we can see that related functionality (like temporary unpublished access) isn't grouped together.

It's not a blocker for merging this, but I'm going to file a followup ticket to clean this up.

beto-aveiga:
@deviantintegral, I can't agree more. The code needs cleaning, and also, making this configurable at the module seems the best option.
Thanks for the review!

deviantintegral:
I filed https://massgov.atlassian.net/browse/DP-24673 for the followup.

PR #1472, opened 2022-04-25
DP-9273 [a11y] Unnecessary alt value for decorative images

ygannett:
**Description:**
Remove the alt text value for node.field_org_page_thumbnail as it's a duplicate to the link text.

DP-9273

**To Test:**
- [ ] Go to an org page with the Our Organizations section. ex)/orgs/office-of-the-governor
- [ ] Check the image link markup at the Our Organizations section.
- [ ] Find the alt values for the images are empty, which used to be duplicate link text.

Before
before

After
after

---


docroot/themes/custom/mass_theme/templates/content/node--org-page--organization-grid.html.twig
```diff
@@ -19,4 +18,5 @@
     'image': image
   }
 }%}
+{# {{ content.imageLink.image }} #}
```
clairesunstudio:
```suggestion
```
cleanup

docroot/themes/custom/mass_theme/templates/content/node--org-page--organization-grid.html.twig
```diff
@@ -7,8 +7,7 @@
 {% set image = false %}
 {% if node.field_org_page_thumbnail is not empty %}
   {% set image = {
-    'src': file_url(node.field_org_page_thumbnail.entity.fileuri ?? ''),
-    'alt': node.field_org_page_thumbnail.alt
```
clairesunstudio, line 11: should we also remove this field in the cms? we shouldn't allow people to enter the information if we are not planning to use it right?

clairesunstudio:
There are two places this is added 
- openmass/docroot/themes/custom/mass_theme/templates/content/node--org-page--logo-link.html.twig 
- openmass/docroot/themes/custom/mass_theme/templates/content/node--org-page--organization-grid.html.twig


Should the alt text be removed in both places?

docroot/themes/custom/mass_theme/templates/content/node--org-page--organization-grid.html.twig
ygannett: I like the idea, but this field just pulls a content item that contains its image and alt text data.  In this edit page, authors don't see the alt text field.  So, we simply don't pass the alt value to the template.
edit-page

docroot/themes/custom/mass_theme/templates/content/node--org-page--organization-grid.html.twig
ygannett: Cleaned up.  Good catch!

clairesunstudio (approved):

PR #1481, opened 2022-04-28
Feature/dp 24690 config ignore

weitzman:
DP-24690

**To Test:**
- [ ] If its green its good.

---


arthurbaghdas (approved):
LGTM, The only thing that I'm thinking about is how this can be realted to the seckit module. But I guess something was changed in the major version and this is just exporting some empty configs

PR #1482, opened 2022-04-28
Fix ahoy pull (devs getting latest DB)

weitzman:
We never had a volume before DDEV so I don't think the reduced persistence will be an issue. Folks can use `ddev pause​​​​​​​` to shut down containers without losing WIP.

DP-24691

**To Test:**
- [ ] Assure that `ahoy pull` gets latest content

---

### DIFF
--- a/.ddev/config.yaml
+++ b/.ddev/config.yaml
@@ -37,3 +37,5 @@ web_environment:
 hooks:
   pre-start:
     - exec-host: echo "No output is shown during the first container build. It may take a minute to install Tugboat, CircleCI, and GitHub CLI tools."
+  post-stop:
+    - exec-host: echo "The database has been reset to a pristine state. If you didn't mean this, next time use ddev pause instead."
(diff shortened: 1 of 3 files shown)


deviantintegral:
> Folks can use ddev pause​​​​​​​ to shut down containers without losing WIP.

Sort of? If you quit Docker for Mac or shutdown colima, or reboot your computer, won't that also reset the database?

For those who work on multiple projects at a time, it seems like this change in behaviour is likely to cause lost data by mistake. Also, wouldn't you lose data if you upgrade ddev, since that triggers new container builds?

Honestly, I'd love for this to work, because it will reduce disk space and time taken on first start. I just wonder if there's a way to do this that keeps `ddev stop` working as expected.

weitzman:
True. Those scenarios will lose the existing DB. In my experience I rarely have valuable data in the DB but that may be because I don't do much site building. Curious what @tom-fleming thinks. I have no problem keeping the volume and then deleting it at beginning of `ahoy pull`.

I dont see a way to ditch the volume AND have persistence after `ddev stop`.

deviantintegral:
It's not part of this PR, but why are we running `~/.ddev/bin/docker-compose -f .ddev/.ddev-docker-compose-full.yaml down --rmi all` in `ahoy pull`?  Even with a local registry cache it takes several minutes for things to download and extract. Without that, I'd expect this to take a while given it triggers downloading 3-5 GB of data.

Also, how often does `massgov/mysql-sanitized:latest` get updated? I pulled a few hours ago, and I'm surprised to see it redownloading from Docker Hub again.

Anyways, I may not get to finishing reviewing this until tomorrow.

deviantintegral:
My comment about pull above may not be relevant. ddev uses `:super`, not `:latest`, so I'm guessing it's not the same.

I still see a volume being created: `local     mass-mariadb` - however it's empty, and isn't mounted, so I think it's OK.

Overall things look good! I tested both ddev pause, and restarting colima. In both cases I didn't lose any data 🎉 .

I added a post-stop reminder about resetting the database. If you're happy with that I think we're good to merge!

deviantintegral (approved):

PR #1488, opened 2022-05-03
DP-24369: Migrate content to integration branch for landing page features on service page PART 2

arthurbaghdas:
**Description:**
Drush command created with offset and limit arguments.

DP-24369

**To Test:**
- [ ]  drush mass-content:migrate-service 0 500
- [ ] drush mass-content:migrate-service 500 500
- [ ] drush mass-content:migrate-service 1000 500
- [ ] drush mass-content:migrate-service 1500 500
- [ ] drush mass-content:migrate-service 2000 500

Note: the first number is the offset where to start, the second number is how many nodes to process
---


docroot/modules/custom/mass_content/src/Commands/MassContentCommands.php
```diff
@@ -126,4 +126,95 @@ public function migrateDateFields(string $type = '', int $limit = 0) {
...
+    $fields = [
+      'field_service_ref_actions_2',
+      'field_link_group',
+      'field_service_ref_actions',
+      'field_service_key_info_links_6',
+      'field_template',
+      'field_service_ref_locations'
+    ];
+    // 1. Log the start of the script.
+    $this->loggerChannelFactory->get('mass_content')->info('Update nodes batch operations start');
```
weitzman: The Drush way is `$this->logger()->info()`

docroot/modules/custom/mass_content/src/Commands/MassContentCommands.php
```diff
@@ -126,4 +126,95 @@ public function migrateDateFields(string $type = '', int $limit = 0) {
...
+      $query->sort('nid');
+      if ($limit !== 0) {
+        $query->range(0, $limit);
+      }
+
+      $nids = $query->execute();
+    }
+    catch (\Exception $e) {
+      $this->output()->writeln($e);
+      $this->loggerChannelFactory->get('mass_content')->error('Error found @e', ['@e' => $e]);
```
weitzman: Are you intending to log an eror and keep going? Thats what this does. ALso, I think you want e->getMessage() at the end. Otherwise you are not passing a string but rather an object of type Exception.

docroot/modules/custom/mass_content/src/Commands/MassContentCommands.php
```diff
@@ -126,4 +126,95 @@ public function migrateDateFields(string $type = '', int $limit = 0) {
...
+    batch_set($batch);
+    // 6. Process the batch sets.
+    drush_backend_batch_process();
+    // 6. Show some information.
+    $this->logger()->notice("Batch operations end.");
+    // 7. Log some information.
+    $this->loggerChannelFactory->get('mass_content')->info('Update batch operations end.');
+    \Drupal::getContainer()->get('plugin.cache_clearer')->clearCachedDefinitions();
+
+    // Turn on entity_hierarchy writes after processing the item.
```
weitzman, line 218: This comment is inconsistent with the code. The code is disabling, not enabling. Also, does this need to be done for every item? Note that there is a finished callback for batches where you can do cleanup work like enabling writes.

weitzman (changes requested):

changelogs/DP-24369.yml
```diff
@@ -0,0 +1,41 @@
...
+# Changed:
+#  - description: Automating the release branch.
+#    issue: DP-10166
+#  - description: Second change item that needs a description.
+#    issue: DP-19875
+#  - description: Third change item that needs a description along with an issue.
+#    issue: DP-19843
+#
+Changed:
+  - description: Migrate content to integration branch for landing page features on service page part 2.
```
weitzman, line 40: We have 5 changelog files, some with multiple items. The point of these files is to provide release notes. Can we provide just 1 brief release note instead of many?

docroot/modules/custom/mass_content/src/Commands/MassContentCommands.php
arthurbaghdas: Fixed, thank you

docroot/modules/custom/mass_content/src/Commands/MassContentCommands.php
arthurbaghdas: Fixed.

docroot/modules/custom/mass_content/src/Commands/MassContentCommands.php
arthurbaghdas: Code is correct, it is enabling entity hierarchy. I will look into the other part and refactor accordingly. Thank you

changelogs/DP-24369.yml
arthurbaghdas: We can, but This can be done on the integration branch itself before merging to develop.

PR #1495, opened 2022-05-05
Dp 24326 promo page permissions updates

beto-aveiga:
**Description:**
      - Not require unpublished date for promotional pages
      - KPI fields not required on promo pages
      - Delete role campaign_landing_page_publisher

DP-24326

---


docroot/modules/custom/mass_scheduled_transitions/mass_scheduled_transitions.module
```diff
@@ -24,7 +24,6 @@ const MASS_SCHEDULED_TRANSITIONS_OPTIONS = [
 ];
 define('MASS_SCHEDULED_TRANSITIONS_CAMPAIGN_LANDING_MAX_DURATION', '14 months');
```
weitzman: I dont think this needs its own max duration anymore. Treated like other content?

weitzman (changes requested):

docroot/modules/custom/mass_scheduled_transitions/mass_scheduled_transitions.module
beto-aveiga: I don't understand the question or what to do. 

I see 2 references to that constant on `docroot/modules/custom/mass_scheduled_transitions/mass_scheduled_transitions.module`.

weitzman: OK, we dont have to change this now as the Jira ticket doesnt mention this policy changing.

weitzman (approved):

PR #1496, opened 2022-05-05
DP-24306 Upgrade to the latest Chrome image for tests

deviantintegral:
**Description:**
- Switches to Selenium images so we can easily support ARM. I found no pure up-to-date chromedriver only images.
- Sets up ddev for it.
- Switches one test to prove it works (it passes on my local!)

DP-24306

**To Test:**
- [ ] Remove `.ddev/docker-compose.chromedriver.yaml` if it exists
- [ ] Run `ddev service enable selenium-chrome`
- [ ] Run `ddev start`
- [ ] Run `ddev drush -y updatedb && ddev drush -y cim`
- [ ] In your browser, connect to https://mass.local:7900 and log in with VNC using the default password of `secret`.
- [ ] Run `ddev ssh vendor/bin/phpunit --testsuite existing-site-javascript`
- [ ] See Chrome start and go through tests.
- [ ] All tests should pass 🎉 

---


.ddev/.disabled-services/docker-compose.chromedriver.yaml
```diff
@@ -6,13 +6,24 @@
 version: '3.6'
 services:
   chromedriver:
-    image:  previousnext/chrome-headless:65
+    image: seleniarm/standalone-chromium:4.1.4-20220429
+    shm_size: 2gb
```
deviantintegral: This is needed to prevent ERR_INSUFFICIENT_RESOURCES in Chrome, but I'm not sure why we don't have it specified in https://github.com/Lullabot/drainpipe-dev/blob/main/scaffold/nightwatch/docker-compose.selenium.yaml. I want to followup on that.

.ddev/.disabled-services/docker-compose.chromedriver.yaml
```diff
@@ -6,13 +6,24 @@
...
     container_name: ddev-${DDEV_SITENAME}-chromedriver
+    expose:
+      - 7900
```
deviantintegral: This isn't working yet (novnc is a browser vnc client).

.ddev/.disabled-services/docker-compose.chromedriver.yaml
```diff
@@ -6,13 +6,24 @@
...
+    ports:
+      - 5900:5900
```
deviantintegral: I was sure ddev had dynamic port exports, but it's not here or in our database container. Perhaps port conflicts are only a problem for people who commonly run multiple sites concurrently.

.ddev/.disabled-services/docker-compose.chromedriver.yaml
```diff
@@ -24,7 +35,6 @@ services:
       # - SIMPLETEST_DB=sqlite://tmp/test.sqlite
       #
       # Note: Do not modify the base URL value.
-      - DTT_API_URL=http://chromedriver:9222
       - BROWSERTEST_OUTPUT_DIRECTORY=/tmp
       - BROWSERTEST_OUTPUT_BASE_URL=${DDEV_PRIMARY_URL}
-#      - MINK_DRIVER_ARGS_WEBDRIVER=["chrome", {"browserName":"chrome","chromeOptions":{"args":["--disable-gpu","--headless", "--no-sandbox"]}}, "http://chromedriver:9515"]
+      - DTT_MINK_DRIVER_ARGS=["chrome", null, "http://chromedriver:4444"]
```
deviantintegral: I think we may want to rename chromedriver to just chrome or selenium.

docroot/modules/custom/mass_alerts/tests/src/ExistingSiteJavascript/AlertsPlacementTest.php
```diff
@@ -5,13 +5,13 @@
...
 use Drupal\paragraphs\Entity\Paragraph;
-use weitzman\DrupalTestTraits\ExistingSiteWebDriverTestBase;
+use weitzman\DrupalTestTraits\ExistingSiteSelenium2DriverTestBase;
 use weitzman\LoginTrait\LoginTrait;
 
 /**
  * Test Alerts Placement.
  */
-class AlertsPlacementTest extends ExistingSiteWebDriverTestBase {
+class AlertsPlacementTest extends ExistingSiteSelenium2DriverTestBase {
```
deviantintegral, line 14: Fixing this both explained so much and left me confused.

`chromedriver` is suppose to speak the webdriver protocol. `selenium` is also the webdriver protocol. So why do we need separate base classes? Does chromedriver speak webdriver, but also extend it with additional non-webdriver APIs like the call to `http://chromedriver:9515/json/version`?

deviantintegral:
There is one remaining failure:

```
1) Drupal\Tests\mass_content\ExistingSiteJavascript\ExpandCollapseElementsTest::testMultipleAccordions
WebDriver\Exception\UnknownError: element not interactable
  (Session info: chrome=101.0.4951.41)
Build info: version: '4.1.4', revision: '535d840ee2'
System info: host: '41adaa747c88', ip: '172.20.0.5', os.name: 'Linux', os.arch: 'amd64', os.version: '5.15.32-0-virt', java.version: '11.0.14'
Driver info: driver.version: unknown

/var/www/html/vendor/instaclick/php-webdriver/lib/WebDriver/Exception.php:156
/var/www/html/vendor/instaclick/php-webdriver/lib/WebDriver/AbstractWebDriver.php:172
/var/www/html/vendor/instaclick/php-webdriver/lib/WebDriver/AbstractWebDriver.php:271
/var/www/html/vendor/instaclick/php-webdriver/lib/WebDriver/Container.php:241
/var/www/html/vendor/behat/mink-selenium2-driver/src/Selenium2Driver.php:805
/var/www/html/vendor/behat/mink-selenium2-driver/src/Selenium2Driver.php:791
/var/www/html/vendor/behat/mink/src/Element/NodeElement.php:153
/var/www/html/docroot/modules/custom/mass_content/tests/src/ExistingSiteJavascript/ExpandCollapseElementsTest.php:32
/var/www/html/docroot/modules/custom/mass_content/tests/src/ExistingSiteJavascript/ExpandCollapseElementsTest.php:50
/var/www/html/docroot/modules/custom/mass_content/tests/src/ExistingSiteJavascript/ExpandCollapseElementsTest.php:59
/var/www/html/vendor/phpunit/phpunit/src/Framework/TestResult.php:726
/var/www/html/vendor/phpunit/phpunit/src/Framework/TestSuite.php:677
/var/www/html/vendor/phpunit/phpunit/src/Framework/TestSuite.php:677
/var/www/html/vendor/phpunit/phpunit/src/Framework/TestSuite.php:677
/var/www/html/vendor/phpunit/phpunit/src/TextUI/TestRunner.php:673
/var/www/html/vendor/phpunit/phpunit/src/TextUI/Command.php:143
/var/www/html/vendor/phpunit/phpunit/src/TextUI/Command.php:96
phpvfscomposer:///var/www/html/vendor/phpunit/phpunit/phpunit:97
```

This is because on https://mass.local/how-to/qag-request-help-with-a-computer-problem we have an accordion inside the "This is a part of" table of contents, but it's not visible when we try to click it.

As well, I haven't figured out how the Behat tests are connecting to Chrome, so we can make sure anything there is upgraded too.

docroot/modules/custom/mass_alerts/tests/src/ExistingSiteJavascript/AlertsPlacementTest.php
weitzman: I'm actually not that clear on this part of DTT either. Last we spoke @jibran wanted to remove `ExistingSiteWebDriverTestBase` and just keep ExistingSiteSelenium2DriverTestBase  so perhaps we dont need to puzzle over this.

.ddev/.disabled-services/docker-compose.chromedriver.yaml
deviantintegral: Ah-ha! This is now not recommended. In drainpipe, we disable shared memory via this flag: https://developers.google.com/web/tools/puppeteer/troubleshooting#tips

I'll find out about getting this flipped. I'm curious why it's not by default in Selenium's containers.

deviantintegral: According to https://github.com/puppeteer/puppeteer/issues/1834 it's not default because it's not a "solve everywhere" problem. Since you're storing to disk, it's entirely possible to run out of disk space, or have much slower tests.

It looks like Cypress has landed on disabling this by default, so I'm guessing in practice it's better to "assume Docker": https://github.com/cypress-io/cypress/blob/develop/packages/server/lib/browsers/chrome.ts#L118-L120

However, locally I'm seeing around a 30% performance hit by disabling the flag. Perhaps we're happier to just set a limit? I'm going to do a little more benchmarking, because adding minutes to test runs seems worth it.

deviantintegral: I tested with both on and off with:

`time (for i in {1..3}; do vendor/bin/phpunit --testsuite existing-site-javascript --filter testWideAlertPlacement; done)`

Test times were not noticeably different, taking around 90s each. I'll push up a commit taking upstream's recommendation.

.ddev/.disabled-services/docker-compose.chromedriver.yaml
deviantintegral: This is now working!

.ddev/.disabled-services/docker-compose.chromedriver.yaml
deviantintegral: I've done this!

deviantintegral:
We're getting there! Three tests are failing still which I need to investigate. I'm updating my local DB so I'm pausing for now.

- Behat doesn't need updates since all of it's tests are with goutte. I added comment to note that, while still updating the configuration.
- There's three tests failing I need to investigate.
- I think we probably want to flip around the configuration where phpunit.xml.dist contains ddev configs, and then we override it via circleci variables. Or perhaps we make circle respond at `selenium-chrome`.

docroot/modules/custom/mass_alerts/tests/src/ExistingSiteJavascript/AlertsPlacementTest.php
deviantintegral: I think I figured it out, but totally unintentionally. I wanted to see why a test wasn't failing on develop, and I wanted to see the browser live. `chromedriver` wasn't working, throwing 400 responses. After inspecting the existing container, it's using _headless chrome_ with the [Chrome DevTools Protocol](https://chromedevtools.github.io/devtools-protocol/). At a basic level, that means running Chrome directly with `/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome --remote-debugging-address=0.0.0.0 --remote-debugging-port=9222` (macOS firewall for me blocks the port externally, and I was lazy so I just set `0.0.0.0`).

I think the confusion comes from various naming:

- `\DMore\ChromeDriver\ChromeDriver` is older than the `chromedriver` project. The former class uses the DevTools protocol, while `chromedriver` speaks WebDriver.
- `\weitzman\DrupalTestTraits\WebDriverTrait` and `\weitzman\DrupalTestTraits\ExistingSiteWebDriverTestBase` both use `\DMore\ChromeDriver\ChromeDriver`, so they use the DevTools protocol.

If this all makes sense, I think we should file a few pull requests updating documentation and class headers to make this all clearer.

weitzman: It doesnt really make sense to me, but thats not your fault. I'm not really looking to become a Chrome expert. I would be more than happy to merge those PRs :)

deviantintegral:
I've been fixing up more issues with state crossing between tests for accordions. Mostly, I will fix one test case, but then that leaves an accordion closed that was originally expected to be open. However, I've got those done.

The one part left is tricky:

Screen Shot 2022-05-10 at 8 23 09 PM

Chrome throws an error that the button isn't clickable, because it has a span over top of it. That's true from a spec standpoint, even if user events are allowed to break through. I'm not quite sure how to fix this, and if this is something we can fix in `contact-us.twig` or if the HTML really is OK.

I suppose a really horrible hack would be to use jQuery to click the link instead of WebDriver. Let's aftertalk this on standup tomorrow.

deviantintegral:
Tests are finally passing! Once https://github.com/massgov/openmass/pull/1506 is reviewed and merged, I will merge here, and then I will likely create a new PR that omits all of my testing commits here.

deviantintegral:
Closed in favour of https://github.com/massgov/openmass/pull/1508

docroot/modules/custom/mass_alerts/tests/src/ExistingSiteJavascript/AlertsPlacementTest.php
deviantintegral: Filed https://gitlab.com/weitzman/drupal-test-traits/-/merge_requests/109

PR #1500, opened 2022-05-09
DP-24808 Add list tag for news list on org pages

weitzman:
**Description:**
Add cache tag for the list. When a new news item is posted, all org pages will be invalidated. 

DP-24808

**To Test:**
- [x] Enable `http.response.debug_cacheability_headers: true` in `services.local.yml`
- [x] Run `drush cr`
- [x] Run `curl -v https://mass.local/orgs/executive-office-for-administration-and-finance > /dev/null`
- [x] Verify that `node_list:news` shows up in the `x-drupal-cache-tags` header.

---

### DIFF
--- a/changelogs/DP-24808.yml
+++ b/changelogs/DP-24808.yml
@@ -1,0 +1,41 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
+# #############################################################################
+# You can use one of the following types:
+#  - Added: For new features.
+#  - Changed: For changes to existing functionality.
+#  - Deprecated: For soon-to-be removed features.
+#  - Removed: For removed features.
+#  - Fixed: For any bug fixes.
+#  - Security: In case of vulnerabilities.
+#
+# Format
+# #############################################################################
+# The format is crucial. Please follow the examples below. For reference, the requirements are:
+#  - All 3 parts are required and you must include "Type", "description" and "issue".
+#  - "Type" must be left aligned and followed by a colon.
+#  - "description" must be indented with 2 spaces followed by a colon
+#  - "issue" must be indented with 4 spaces followed by a colon.
+#  - "issue" is for the Jira ticket number only e.g. DP-1234
+#  - No extra spaces, indents, or blank lines are allowed.
+#
+# Example:
+# #############################################################################
+# Fixed:
+#  - description: Fixes scrolling on edit pages in Safari.
+#    issue: DP-13314
+#
+# You may add more than 1 description & issue for each type using the following format:
+# Changed:
+#  - description: Automating the release branch.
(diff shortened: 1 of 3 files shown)


deviantintegral (approved):
If I had to guess, this is hard to replicate in production because other tags like `node_list:event` are on the pages as well, and they get cleared often enough to make the missing news tag not noticeable.

This adds the new cache tag as advertised. It also makes sense to me why this is happening; as soon as we call `referencedEntities()` above, we need to manually remember to combine the appropriate cache tags.

Is there a reason we're calling in directly to the renderer service? Normally I would expect something like:

```php
      (new \Drupal\Core\Cache\CacheableMetadata())->addCacheTags(['node_list:news'])
        ->applyTo($variables);
```

I'm also confused as to why we have a `$cache_tags` variable that appears to be written to, but never read or passed to the renderer. That's not touched by this PR, but feels like we've got some code cleanups to do? Part of me thinks we should push all cache tag code into methods like `\Drupal\mayflower\Prepare\Organisms::preparePressListing`, especially as `mass_theme.theme` is over 7000 lines long.

Anyways, these can all be followups or further discussion.

weitzman:
> Is there a reason we're calling in directly to the renderer service? Normally I would expect something like:

$variables is just a regular array of name/value pairs, and not a render array. So applyto() would have no effect as $variables never gets rendered. Maybe I am missing something? Or are you saying to do this in addition to the render?

Yeah the unused $cache_tags is odd. I'd like to fix that in this PR. I think its OK to hold this for a week to get it right.

deviantintegral:
It does work! I actually tested locally before commenting. Core does this itself in `\template_preprocess_pager()`.

Higher level, when it comes to this type of code, I tend to fall back to https://www.drupal.org/project/drupal/issues/2848158#comment-11904744 noting that complex rendering in preprocess functions is difficult at best. But for a site like this, that's a ton of refactoring.

weitzman:
Pushed a commit which propagates $cache_tags using applyTo() as you suggest.

weitzman:
>  complex rendering in preprocess functions is difficult at best. But for a site like this, that's a ton of refactoring.

Agreed! The rendering approach of this site comes from the early days of D8, before we really knew what we were doing. Further, we had to integrate with Patternlab and that best practice was super unclear. I'd love for you to research this for us, if time permits.

weitzman:
@deviantintegral can you take one more look at this PR? I pushed a commit with your suggestions. Thanks.

deviantintegral (approved):
This PR looks good to me!

Do we want to file followup tickets on how cache tags are managed at a higher level? I think it depends on how often we're encountering production bugs due to caching issues.

weitzman:
I dont think we need that ticket as this time. These bugs are rare.

PR #1505, opened 2022-05-11
Install Imagick at Tugboat

weitzman:
DP-24370

Please review these three upstream patches created to solve some of our issues:

- https://www.drupal.org/project/stage_file_proxy/issues/3283162
- https://www.drupal.org/project/stage_file_proxy/issues/3282524
- https://www.drupal.org/project/stage_file_proxy/issues/3282542

Since there is a conflict patch can't solve (but git merge can!) I've taken the result of merging all three branches and committed that as a patch, along with the separate files for posterity.

**To Test:**
- [ ] Enable backstop in ddev, and then run `find docroot/sites/default/files -type f -name '*jpg' -exec rm -v {} \; && ddev exec --service backstop backstop --config=/src/backstop.js reference --target=local --list=all` while watching `ddev logs -f`. There should be no errors related to images or stage file proxy. There may be others! Note this will **delete all of your local cached assets** so if you want those, move them out of the way instead of running the `find` command.

**Comparing Acquia and Tugboat**:
- [ ] Edit `all.json` to only contain:
```json
[
  {"label": "OrgGovernorElectedOffical", "url": "/orgs/office-of-the-governor", "screens": ["desktop", "mobile"]},
  {"label": "NewsNews", "url": "/news/qag-newsnews", "screens": ["desktop", "tablet"]},
  {"label": "LocationPark3", "url": "/locations/qag-locationpark3"}
]
```
- [ ] `ddev exec --service backstop backstop --config=/src/backstop.js reference --target=feature3 --list=all`
- [ ] `ddev exec --service backstop backstop --config=/src/backstop.js test --target=tugboat --tugboat=https://pr1505-l7ugynjrbbjlgtqlvm8rjnru6oj9d7dv.tugboat.qa --list=all`
- [ ] open `backstop/report/index.html`.
- [ ] No differences should be detected.

---


.tugboat/config.yml
```diff
@@ -47,7 +47,12 @@ services:
     commands:
       # Commands that set up the basic preview infrastructure
       init:
-        # Install opcache and imagemagick.
+        - sudo apt update
+        # Install imagick
+        - apt-get install -y libmagickwand-dev
+        - echo '' | pecl install imagick
```
deviantintegral, line 54: What's this echo for? Add a comment explaining why?

weitzman: Added comment.

backstop/scripts/cookies.js
```diff
@@ -0,0 +1,13 @@
+[
+  {
+    "expirationDate": 1798790400,
```
weitzman, line 3: Is this a far future date?

deviantintegral: It's `Friday, January 1, 2027 8:00:00 AM` and is what Backstop uses by default in their examples.

deviantintegral:
I ran the above test steps and got six failures:

1. It appears we have lazy loading images, and aren't ensuring they are loaded before taking the screenshot. Left is `cd`, right is `tugboat` in this example:2. 
Screen Shot 2022-05-24 at 3 53 56 PM

2. We are still getting invalid image errors (which I've seen locally too). This only happens either on first page load (stage file proxy?) _or_ is transient. I will investigate.
Screen Shot 2022-05-24 at 3 55 15 PM

3. There is an `ariaHidden` notice. While we should fix this, if we want to use tugboat for backstop tests we should consider not rendering errors _or_ rendering them on `cd` too?

https://massgov.atlassian.net/browse/DP-24801?jql=text%20~%20%22ariahidden%22

weitzman:
Code looks good. Thanks for all the d.o. patches. I'll test this locally and then merge.

PR #1510, opened 2022-05-12
Add new mailchimp text format and use to formail mails

weitzman:
**Description:**
- The Pathologic text filter is living up to its name. It is making relative links out of our absolute links which is helpful in a browser but unhelpful in an email. Create a new mailchimp text filter that doesnt include it.
- The mailchimp_transactional mailsystem is adding unwanted encoded characters at end of lines. Override its format() method in a new mail plugin.

DP-24906

**To Test:**
- [ ] Reenable the secret key on stage and send a test New User email. Also a password reset email since that was how the ticket came into us.

---


tom-fleming (approved):
LGTM

PR #1515, opened 2022-05-13
DP-24915 Only load development services when in development mode

deviantintegral:
**Description:**
Disables development services, notably twig debugging, on Tugboat.

DP-24915

**To Test:**
- [ ] [/orgs/qag-executive-office-of-technology-services-and-security/locations?_page=1&cachebuster=4cti8n](https://pr1515-uba7xoqbpyqysbmldjz7e5irnkrvwl5w.tugboatqa.com/orgs/qag-executive-office-of-technology-services-and-security/locations?_page=1&cachebuster=4cti8n) should no longer show Twig debugging comments.

---


deviantintegral:
The failures are because there are tests requiring the `X-Drupal-Cache-Tags` header, and that's no longer being emitted.

What's odd is that development.services.yml has that option set to `false`. I need to investigate further. Core sets this for it's own functional tests at `\Drupal\Core\Test\FunctionalTestSetupTrait::initSettings()`, and we probably need to do something similar for both Behat and PHPUnit.

PR #1532, opened 2022-05-20
DP-24950 expose collections on /admin/ma-dash/documents

beto-aveiga:
**Description:**
- Exposing collections on /admin/ma-dash/documents
- Adding collections filter to /admin/ma-dash/documents and its "export to CSV" display.
- Adding the collections field on the CSV file.

DP-24950

**To Test:**
- [ ] Visit /admin/ma-dash/documents
- [ ] Select a collection.
- [ ] TIP: Narrow your search if needed using the title filter.
- [ ] Filter the view.
- [ ] Open in a new tab the link for CSV on the bottom of the page.
- [ ] Compare the results, must be the same results.

---


conf/drupal/config/views.view.all_documents.yml
```diff
@@ -1126,6 +1128,65 @@ display:
...
+          group_type: group
+          admin_label: ''
+          plugin_id: taxonomy_index_tid
+          operator: or
+          value: {  }
+          group: 1
+          exposed: true
+          expose:
+            operator_id: field_collections_target_id_op
+            label: 'Collections (field_collections)'
```
joeg8612: let's just get rid of the field name in parentheses here. We don't do this for any other field. Then

conf/drupal/config/views.view.all_documents.yml
```diff
@@ -3170,6 +3297,65 @@ display:
...
```
joeg8612: Remove field name that is in parentheses.

arthurbaghdas: All set

conf/drupal/config/views.view.all_documents.yml
arthurbaghdas: All set

arthurbaghdas (approved):

PR #1545, opened 2022-05-25
DP-24800 fix order changing in curated list links

beto-aveiga:
**Description:**
- Automatic list "sort by date" fixes.
- Tests added.

DP-24800

**To Test:**


docroot/modules/custom/mass_content/src/EntitySorter.php
```diff
@@ -49,6 +42,18 @@ public function sortEntities(array &$entities, $sort) {
...
 
+  /**
+   * Sets a time format that avoid ambiguos date strings.
+   *
+   * To avoid potential ambiguity, it's best to use ISO 8601 (YYYY-MM-DD) dates
+   * or DateTime::createFromFormat() when possible.
+   *
+   * @see https://www.php.net/manual/en/function.strtotime.php
+   */
+  protected function formatDateValue($date, $convert_to_time = 1) {
```
deviantintegral: Should we add typehints to the method parameters, and a return type hint of `string`?

Why is `$convert_to_time` an integer and not a boolean?

docroot/modules/custom/mass_content/tests/src/ExistingSite/EntitySorterTest.php
```diff
@@ -0,0 +1,202 @@
+<?php
+
+namespace Drupal\Tests\mass_content\ExistingSite;
+
+use Drupal\mass_content\EntitySorter;
+use Drupal\node\Entity\Node;
+use stdClass;
```
deviantintegral: I think the convention is to not import it, and instead reference with `\stdClass`. Did PHPStorm suggest this?

docroot/modules/custom/mass_content/tests/src/ExistingSite/EntitySorterTest.php
```diff
@@ -0,0 +1,202 @@
...
+use weitzman\DrupalTestTraits\Entity\MediaCreationTrait;
+use weitzman\DrupalTestTraits\ExistingSiteBase;
+
+/**
+ * Tests EntitySorter.
```
deviantintegral, line 11:
```suggestion
 * Tests EntitySorter.
 *
 * @group existing-site
```

docroot/modules/custom/mass_content/tests/src/ExistingSite/EntitySorterTest.php
```diff
@@ -0,0 +1,202 @@
...
+ */
+class EntitySorterTest extends ExistingSiteBase {
+
+  use MediaCreationTrait;
+
+  private $entitiesForSorting;
+  private $expectedIndexesAsc;
+  private $expectedIndexesDesc;
```
deviantintegral: Let's typehint these at least?

deviantintegral (commented):
Is there a reason why we can't use Drupal's built in date formatter? https://www.drupal.org/docs/8/core/modules/datetime/datetime-overview#s-formatting-dates-programmatically

Otherwise, I love there's a test here, since if we do decide to change the formatter the test should prove we did it right. Nice!

docroot/modules/custom/mass_content/src/EntitySorter.php
beto-aveiga: 🤷🏼‍♂️ I don't know haha... Maybe I was tired and unconsciously decide that 1 and 0 were shorter compared to booleans. I will add the type hints for both parameters. Thanks!

docroot/modules/custom/mass_content/tests/src/ExistingSite/EntitySorterTest.php
beto-aveiga: It was added automatically from VSCode. Maybe it was a good decision from VSCode, the problem is that I didn't add the \ at the beginning to avoid the USE statement.

beto-aveiga:
@deviantintegral 
> Is there a reason why we can't use Drupal's built in date formatted?
I tend to forget these built-in functions when the code I'm fixing is not using them.

Thanks for the valuable review.

This is ready for review.

deviantintegral (approved):
Looks great! Nice work.

PR #1554, opened 2022-05-31
Fix for errors caused by merging cache tags - DP-25028

tom-fleming:
**Description:**

DP-25028

**To Test:**

---

### DIFF
--- a/docroot/modules/custom/mayflower/src/Prepare/Organisms.php
+++ b/docroot/modules/custom/mayflower/src/Prepare/Organisms.php
@@ -1905,7 +1905,8 @@ class Organisms {
           $params = $url->getRouteParameters();
           $entity_type = key($params);
           $entity = \Drupal::entityTypeManager()->getStorage($entity_type)->load($params[$entity_type]);
-          $cache_tags = Cache::mergeTags($entity->getCacheTags(), $cache_tags);
+          $entity_cache_tags = $entity->getCacheTags() ?? [];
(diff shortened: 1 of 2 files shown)


docroot/modules/custom/mayflower/src/Prepare/Organisms.php
```diff
@@ -1905,7 +1905,8 @@ public static function prepareExpandableContent($entities, array $options = [],
           $params = $url->getRouteParameters();
           $entity_type = key($params);
           $entity = \Drupal::entityTypeManager()->getStorage($entity_type)->load($params[$entity_type]);
-          $cache_tags = Cache::mergeTags($entity->getCacheTags(), $cache_tags);
+          $entity_cache_tags = $entity ? $entity->getCacheTags() : [];
+          $cache_tags = Cache::mergeTags($entity_cache_tags, $cache_tags);
 
           // If the entity is a topic_page, include Link Groups and links.
           if (!empty($entity) && $entity->bundle() == 'topic_page') {
```
deviantintegral, line 1910: Instead of checking $entity multiple times, should we either add an early return or wrap the whole block?

tom-fleming: Ah, thanks for the catch! I moved the tag merge to the existing block where we're checking whether the entity is empty.

deviantintegral (approved):
👏 looks great!

PR #1566, opened 2022-06-06
DP-25103 Fix scaffold overwriting example settings file

deviantintegral:
**Description:**
Fixes `composer install` causing immediate git diff's that have to be ignored.

~~Since logically it makes sense to group container configuration together, I've replaced the entire file rather than appending to it: https://github.com/drupal/core-composer-scaffold#overwrite~~

I realized `false` doesn't mean "don't create this file at all and remove it", but "let the file exist as it is in the repo". That's simpler, so I've done that instead.

DP-25103

**To Test:**
- [ ] Check out the branch locally and run `ddev composer install`
- [ ] `docroot/sites/example.settings.local.php` should not show up in `git diff`

---


tom-fleming:
Tested and this looks good to me. However, only approving at the moment due to the phpunit failures.

tom-fleming (approved):

deviantintegral:
> #!/bin/bash -eo pipefail
> su -s /bin/bash root -c "phpunit docroot/modules/custom --verbose --log-junit /tmp/test-results/dtt/junit.xml"
> PHPUnit 9.5.20 #StandWithUkraine
> 
> Runtime:       PHP 7.4.20
> Configuration: /var/www/code/phpunit.xml.dist
> 
> ........S...........S..........[2022-06-06 16:52:52] file.NOTICE: The file public://test.jpg was not deleted because it does not exist.
> ..................
> Session terminated, killing shell...
> 
> Too long with no output (exceeded 10m0s): context deadline exceeded

This is what was brought up on standup, right? I wonder if they're about to post another outage notice: https://status.circleci.com/incidents/4jv5p8lkn124

PR #1567, opened 2022-06-06
Dp 25055 change collection field widget

beto-aveiga:
**Description:**
Use entity_reference_tree instead of term_reference_tree for collections.

DP-25055

**To Test:**
- [ ] Edit any node with a collections field, and notice the entity_reference_tree widget.

image

---


deviantintegral:
@beto-aveiga looks like config tests are failing on this one.

deviantintegral:
Since we're making changes, let's fix the period here?

image

I noticed that I can scroll behind the modal window. It looks like this affects all modals, so I filed https://massgov.atlassian.net/browse/DP-25197.

There's some odd UX issues here, combined with the above scrolling bug:

1. When expanding a tree, the modal grows down below the viewport. The save button is then missing.
2. When you scroll, everything jumps into view, but you also scroll the container.
3. When scrolling the interior  view, you can lose the header

Here's a recording: https://user-images.githubusercontent.com/255023/174820555-2dc2d280-fbb1-4f89-ba2f-8d32cb016b37.mov

There's some odd issues with visual focus:

image

image

Let's discuss on standup today if we should hold off on this until we get the scrolling issues fixed.

PR #1568, opened 2022-06-07
DP-24660 Improve CircleCI Backstop Times

deviantintegral:
**Description:**
- Switches from using a remote docker instance which is limited to 2 CPUs, to running on the job executor node itself.
- Backstop's docker images require us to run docker with `--init`, but we can't do that in Circle's configuration. Instead, we run inside a nodejs image. It's taking ~30 seconds to install backstop, which doesn't matter much given the performance improvemements here. In the future we could build our own docker container from circleci's base or backstop, but that may not actually be any faster depending on shared layer caching.
- Uploads an archive of backstop results instead of individual files, saving ~1m 30s on each build. I used zip because I know we have Windows users and it's the only format it can open out of the box.
- Updates to backstop 6.1.0 (since at first I wondered if that fixed anything)
- Fixes init on ddev for backstop
- Adds an environment variable to set the number of browsers to run. I tried automating this, but it was tricky to do so in a way that would work well on all classes of systems like CI, Intel MacBooks, and M1 Macbooks.
- Logs if load average goes over the number of CPUs, which is a sign that the node is oversubscribed. It's how I noticed the remote docker issue, since it was logging at a load average over 2 instead of the expected 8.

DP-24660

**To Test:**
- [ ] `ddev drush ma:backstop cd feature1 --ci-branch=DP-24660-backstop-perf`
  * Feel free to change the target environments to test.
  * I see an average of 25 minutes at https://app.circleci.com/insights/github/massgov/openmass/workflows/backstop_ad_hoc/overview?branch=develop, so if anyone knows of specific commands to run that are slower I'd be glad to hear them.

---


deviantintegral:
😭 I typo'ed the ticket number in the branch name. I'm going to rebase and open a new PR.

PR #1572, opened 2022-06-07
Dp 25105 fix entity usage tests

tom-fleming:
**Description:**
Adds an entity_usage option to sanitization commands.

DP-25105

**To Test:**
- [ ] Locally, run `ddev exec drush sql-sanitize -y --sanitize-entity-usage` (this might work best if you `ddev ssh` into the container first)
- [ ] Run phpunit tests: `ddev exec phpunit docroot/modules/custom/mass_entity_usage --debug --verbose --log-junit /tmp/test-results/dtt/junit.xml`

---

### DIFF
--- a/docroot/modules/custom/mass_entity_usage/tests/src/ExistingSite/EntityUsageTest.php
+++ b/docroot/modules/custom/mass_entity_usage/tests/src/ExistingSite/EntityUsageTest.php
@@ -36,7 +36,7 @@ class EntityUsageTest extends ExistingSiteBase {
 
     // Remove everything from the entity_usage table
     // to avoid long cleaning times that break this test.
-    $this->database->delete('entity_usage')->execute();
+    \Drupal::service('database')->delete('entity_usage')->execute();
 
     $this->emptyEntityUsageQueues();
     $user = User::create(['name' => $this->randomMachineName()]);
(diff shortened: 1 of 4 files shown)


docroot/modules/custom/mass_entity_usage/tests/src/ExistingSite/EntityUsageTest.php
```diff
@@ -33,6 +33,11 @@ class EntityUsageTest extends ExistingSiteBase {
    */
   protected function setUp() {
     parent::setUp();
+
+    // Remove everything from the entity_usage table
+    // to avoid long cleaning times that break this test.
```
deviantintegral, line 38: Let's leave a hint for us in case we later have trouble with these events needing to be fired for future tests to work.

```suggestion
    // to avoid long cleaning times that break this test.
    // Note this avoids triggering events on bulk deletes, such as at
    // \Drupal\entity_usage\EntityUsage::bulkDeleteTargets().
```

deviantintegral:
```
andrew@mass-web:/var/www/html$ drush sql:sanitize -y --sanitize-entity-usage


  The "--sanitize-entity-usage" option does not exist.
```

I don't see it in the help message. `sanitize-` has no results in the modules directory either.

I pushed a small comment addition to the PR.

Given the PR doesn't change any CI steps to use the above flag, I'm guessing it's not needed or is a typo. I'm going to mark this to merge if tests pass. Nice work!

deviantintegral (approved):

PR #1589, opened 2022-06-15
DP-25136 entity usage not showing all relationships

beto-aveiga:
### Description
Entity usage wasn't correctly working when entities were modified or deleted.

### Explanation
Originally when registering usage count on modified entities, the `original` property was present on the modified entity, and the process of updating the usage count needs the `original` property. 

However, because we are using a queue to calculate the usage count later, the `original` property is not on the entity anymore when we load the entity when processing a queue item to calculate usage. The missing `original` property caused the usage count process to skip, misleading results (for editing and deleting).

Existing tests were passing because they were related to the creation of entities, and the process for registering usage count on creation doesn't need the `original` property, hence the usage numbers were correct, and the existing tests were passing.

On this PR, everything is treated as a creation of an entity when registering usage records, avoiding the need for the `original` property on the entity. The entity usage table is cleaned before counting usage by deleting the usage records from older versions of the source entity.

I added tests for the edition and deletion of the source entities.

DP-25136

**To Test:**
Before testing, to speed up the usage count and ensure you are processing your recent usage records, **only for development environments**, run the following:

> drush sqlq "TRUNCATE entity_usage" # truncate entity usage table
> drush sqlq "DELETE FROM queue WHERE name='entity_usage_tracker'" # delete entity usage tracking queue

All nodes created for entity usage tests should be published.

---
- [x] Create a node: node1.
- [x] Create a node: node2, and link to node1.
- [x] Run `drush queue:run entity_usage_tracker --time-limit=30  && drush clean_usage_table`
- [x] Go to the second tab "Pages linking here" of node1, notice 1 usage record from node2.
---
- [x] Remove the link on node2.
- [x] Run `drush queue:run entity_usage_tracker --time-limit=30  && drush clean_usage_table`
- [x] Go to the second tab "Pages linking here" of node1, notice no usage records.
---
- [x] Restore the link to node1 on node2.
- [x] Run `drush queue:run entity_usage_tracker --time-limit=30  && drush clean_usage_table`
- [x] Go to the second tab "Pages linking here" of node1, notice 1 usage record from node2.
---
- [x] Delete node2.
- [x] Run `drush queue:run entity_usage_tracker --time-limit=30  && drush clean_usage_table`
- [x] Go to the second tab "Pages linking here" of node1, notice no usage records.

---


docroot/modules/custom/mass_entity_usage/mass_entity_usage.services.yml
```diff
@@ -4,6 +4,3 @@ services:
     arguments: ['@entity_type.manager', '@config.factory']
     tags:
       - { name: event_subscriber }
-  mass_entity_usage.clean_usage_table:
```
deviantintegral, line 7: I see this service has been moved to an upstream patch. It's now available as `entity_usage.clean_usage_table`

deviantintegral (commented):
I read through this while I was waiting for a fresh database to download. It all looks good! But I still actually need to test this.

deviantintegral:
@beto-aveiga I think I've found a bug, but I'm not sure if it's related to this PR or patch.

In my curated list, I have two references; one in the rich text overview field, and the other in the Organization entity reference:

Screen Shot 2022-06-16 at 4 11 37 PM

However, in the report, it shows up as if the organization is referencing itself. Instead, I think this second item should be the curated list again, with the title "Test" not "Testing org":

Screen Shot 2022-06-16 at 4 12 13 PM

beto-aveiga:
@deviantintegral, thanks, that is very interesting.

To clarify, this PR is not about how links are counted because that's an independent process for every type of field (rich text, entity reference, links, etc.). I haven't modified how to determine the source/target, and that seems to be the error.

I will try to replicate it locally and find out what is happening.

docroot/modules/custom/mass_entity_usage/tests/src/ExistingSite/EntityUsageTest.php
```diff
@@ -181,6 +181,96 @@ public function testEntityUsageTracking() {
...
+    $node_curated_list_1->delete();
+    $this->processEntityUsageQueues();
+    // Usage for node1 should be 0.
+    $this->assertUsageRows($node_org, 0);
+  }
+
+  /**
+   * Tests multiple references from one entity to another entity.
+   */
+  public function testCuratedList() {
```
beto-aveiga, line 246: @deviantintegral after trying to replicate the issues you had, I ended up writing a test that is hopefully very similar to what you did. The test is showing the expected results.

deviantintegral (approved):
Everything passes! Nice work. You're welcome to set this to automerge after resolving conflicts.

One question though. When a node has multiple references in different fields, the "Pages linking here" only shows the first reference. I'm assuming that's by design? If not, IMO that can be a followup ticket and doesn't block this PR or ticket.

beto-aveiga:
> Everything passes! Nice work. You're welcome to set this to automerge after resolving conflicts.
> 
> One question though. When a node has multiple references in different fields, the "Pages linking here" only shows the first reference. I'm assuming that's by design? If not, IMO that can be a followup ticket and doesn't block this PR or ticket.

That is by the design of the Entity Usage module. See `Drupal\entity_usage\EntityUsage::listUniqueSourcesCount`

PR #1593, opened 2022-06-16
Dp 25153 fix taking a screenshot before mobile menu has animated

beto-aveiga:
**Description:**
- Adds a rule to remove animations by setting the transition duration to 0s.

DP-25153

---

### DIFF
--- a/backstop/scripts/ready.js
+++ b/backstop/scripts/ready.js
@@ -52,6 +52,7 @@ module.exports = async function (page, scenario, vp) {
 
     // Zero delay on CSS transitions.
     jQuery("head").append('<style id="no-more-css-transitions" type="text/css"> * { transition-duration: 0s !important } </style>');
+    jQuery("head").append('<style id="new-animations" type="text/css">*:before, *:after { transition-duration: 0s !important; }</style>');
 
     // Immediately complete any in-progress animations.
     jQuery(':animated').finish();
--- a/backstop/scripts/ready.js
+++ b/backstop/scripts/ready.js
@@ -49,6 +49,10 @@ module.exports = async function (page, scenario, vp) {
   await page.evaluate(function (url) {
     // Disable jQuery animation for any future calls.
     jQuery.fx.off = true;
+
+    // Zero delay on CSS transitions.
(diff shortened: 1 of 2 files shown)


backstop/scripts/ready.js
```diff
@@ -49,6 +49,10 @@ module.exports = async function (page, scenario, vp) {
   await page.evaluate(function (url) {
     // Disable jQuery animation for any future calls.
     jQuery.fx.off = true;
+
+    // Zero delay on CSS transitions.
+    jQuery("head").append('<style type="text/css"> *, *:before, *:after { transition-duration: 0s !important } </style>');
```
deviantintegral, line 54: I like this because it would reduce all backstop test times, and we're not trying to test animations themselves.

However, do we still need to add a selector wait for the menu itself? It seems like there could still be a delay between calling click() and when the menu has _actually_ rendered.

beto-aveiga: @deviantintegral calling a click and rendering should be executed synchronously, in the same thread right? The same for taking the screenshot later. I don't think we need to wait, I can't understand why we would need to wait. If you think waiting is necessary, I can add a small `wait` right now to PR. Thanks!

deviantintegral: It's the same thread, but all events are async AFAIK: https://stackoverflow.com/a/69111229/7982687

Regardless, if this is a real-world problem we'll see it pretty quickly and know what to do. Let's get this in and we can always do a followup.

deviantintegral:
Running https://app.circleci.com/pipelines/github/massgov/openmass/19910/workflows/b2d46fb9-aff5-43b5-b329-fec9a4f2ba76 to make sure we don't see any menu-related issues.

deviantintegral:
Good news! I don't see any menu failures. I think this PR is good to merge.

However, it looks like we have a regression on develop. There appears to be some extra height on the breadcrumb. I'll follow up.

beto-aveiga:
@deviantintegral it seems that adding a bit of delay before taking the screenshots _just in case_ can fix these tests. I wonder why the delay is needed, and what is creating that gap.

PR #1596, opened 2022-06-17
DP-25190 Create revisions on media actions

tom-fleming:
**Description:**
This PR creates revisions, revision messages, and log messages for moderation state bulk actions executed on media items. 

DP-25190

**To Test:**
- [ ] Add 2-3 documents (media/add/document). 
- [ ] Go to Content > All Documents (admin/ma-dash/documents)
- [ ] Select the documents you created (other documents will not work on testing environments since the files do not exist in those environments)
- [ ] Test bulk actions on your documents for Publish, Unpublish, Trash, and Restrict. Check that the transitions work successfully
- [ ] Check logs for a message like "Bulk action: Moderation state set to published for media item 123456."
- [ ] Check revisions for the same message. At the time this PR is being created, there isn't a UI for this, but you can query for the messages in your environment's database with query like `select * from media_revision where mid=123456;`

---


docroot/modules/custom/mass_media/src/Traits/MediaModerationStateActionTrait.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+trait MediaModerationStateActionTrait {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function createRevision(MediaInterface $entity = NULL, $moderation_state) {
+    $revision_message = $this->t('Bulk action: Moderation state set to @moderation_state for media item @entity_id.', [
+      '@moderation_state' => $moderation_state,
+      '@entity_id' => $entity->id(),
```
deviantintegral, line 20: I think these should be passed as context variables in the `notice` call, avoiding the need to call `t()` here? That way we can ensure that if a future moderation state contains non-HTML printable characters that they will be escaped correctly.

deviantintegral (commented):
One minor note, but I'm curious; do we need this code at all? I see we have views bulk operations installed, and according to https://www.drupal.org/project/views_bulk_edit/issues/3038400 it should support media entities. I imagine they already handle revisions correctly.

Part of my thinking is that given this hit production, we should probably write a test around this to ensure media items get a revision. I'm imagining something like "loop through all moderation states and set them", so that if a future moderation state is added we'll be testing that automatically. But if VBO already setting media moderation states for us, then we could likely avoid writing that test.

tom-fleming:
> One minor note, but I'm curious; do we need this code at all? I see we have views bulk operations installed, and according to https://www.drupal.org/project/views_bulk_edit/issues/3038400 it should support media entities. I imagine they already handle revisions correctly.
> 
> Part of my thinking is that given this hit production, we should probably write a test around this to ensure media items get a revision. I'm imagining something like "loop through all moderation states and set them", so that if a future moderation state is added we'll be testing that automatically. But if VBO already setting media moderation states for us, then we could likely avoid writing that test.

@deviantintegral We still need some of the code to address the access rules and file moving behavior around the Restricted moderation state, but I think we could probably drop the revision creation code if we instead rely on VBO's ActionBase and set our own customAccess methods. At the time I didn't explore that, since there was a higher urgency to deploy a fix, but I think I'll take a look at that approach now.

deviantintegral (approved):

docroot/modules/custom/mass_media/src/Traits/MediaModerationStateActionTrait.php
tom-fleming: My primary motivation for using `t` here was to have the same message for both the revision log and the `notice` call, and `setRevisionLogMessage` doesn't support string translation as far as I can tell. With that said, I don't feel too strongly about having two different messages, especially if you think relying on the `notice` call's string translation is more beneficial.

deviantintegral: Eh, I'm happy either way. You know _about_ the context option, so good enough by me.

arthurbaghdas (approved):

PR #1598, opened 2022-06-17
DP-25156 wait for alerts to be processed

beto-aveiga:
**Description:**
Backstop - wait for all alerts on the page to be processed.

DP-25156

---


deviantintegral:
There significant content differences between Tugboat and CD right now. For example, tugboat's develop is missing social media links in the footer. My local was too, and updating my database fixed it. I'm refreshing develop now, and will test again after this branch PR is done.

deviantintegral:
The failures I see are:

1. The site-wide alert text changed due to the under-5 availability of the COVID vaccine 🎉 .
2. On QAG Decision Order, the image of the school bus has different compression artifacts.
3. The "show one more" bug that's fixed in another ticket.
4. A cut off footer in a screenshot (also a different ticket).
5. Other content differences.

Regardless, "BinderAudit" doesn't fail, which is what this fixed, so let's get it in. Nice!

PR #1608, opened 2022-06-21
Dp 25168 instrument cache hit rate for metatag

beto-aveiga:
**Description:**

DP-****

**To Test:**

---


beto-aveiga:
@deviantintegral: I just added cacheable dependencies for a few tokens in one file on this PR. Is that expected for:
> Make sure any custom tokens have in mass_* modules are setting appropriate cache tags. I know we’re removing the parent topics and services tokens, but they weren’t setting cache tags at all so we should check other tokens too.

If so, I can add similar code to the other custom modules.

weitzman:
I didn't see anywhere where we "instrument cache hit rate". Is that in the PR yet?

beto-aveiga:
@weitzman not yet. I talked to @alarrimore about this ticket a few days ago but I'm still not sure how to begin, so I just worked on one of the parts in the issue description.

weitzman:
OK. I've move it to Draft

PR #1612, opened 2022-06-22
DP-25201: Modify service page migration logic for integration branch

arthurbaghdas:
**Description:**

DP-25201

**To Test:**

---


docroot/modules/custom/mass_content/includes/mass_content.service_page.inc
```diff
@@ -11,7 +11,7 @@ use Drupal\mayflower\Helper;
 /**
  * Helper function to check if paragraph of given type is already populated.
  */
-function check_paragraph($node, $bundle) {
+function check_paragraph_bundle($node, $bundle, $heading = NULL) {
```
tom-fleming: @arthurbaghdas It doesn't look like the `$heading` argument is used here. Is it needed?

tom-fleming:
@arthurbaghdas Not sure why this wasn't called out or caught before, but the drush command is calling this currently non-static method as a static method, which triggered an error: https://github.com/massgov/openmass/blob/DP-25201_modify_service_page_migration/docroot/modules/custom/mass_content/src/MassContentBatchManager.php#L88

docroot/modules/custom/mass_content/includes/mass_content.service_page.inc
arthurbaghdas: Forgot to remove it, I was testing another approach

tom-fleming:
LGTM!

PR #1615, opened 2022-06-22
DP-25197 avoid body overflow when a modal is opened

beto-aveiga:
**Description:**
- A mutation observer is used to observe the new dialogs added to the body element. 
- Every time a mutation is observed, it checks if there is a modal opened.
- The mutation function is debounced, to avoid performance penalties.
- A data attribute is set to the body when mutations are observed, only true when a modal is added. If true, a CSS rule will avoid the body overflow.

DP-25197

**To Test:**

---


beto-aveiga:
Replaced with patch.
See new https://github.com/massgov/openmass/pull/1622

beto-aveiga:
Probably good for us.

deviantintegral:
This is looking good! The behaviour of the modal itself is much improved.

I'm testing with Safari, though I don't think this is a browser bug. When I try to insert a media entity download link in a rich text editor component, I get:

> TypeError: Drupal.debounce is not a function. (In 'Drupal.debounce(controlBodyOverflow, 250)', 'Drupal.debounce' is undefined)

The modal is left blank and open, without the link inserted:

image

Otherwise, image modals in rich text and views all look good.

beto-aveiga:
@deviantintegral I forgot to add core/drupal.debounce as a dependency. When I tested with user 1 it worked, probably because the debounce library was a dependency of another library. Anyhow, I just added core/drupal.debounce as a dependency, which will make it work for sure. Thanks!

deviantintegral:
Excellent, fixed! Clicking the merge button.

PR #1634, opened 2022-07-04
DP-25147: Remove locations-old

arthurbaghdas:
**Description:**

DP-25147

**To Test:**

---


.circleci/config.yml
```diff
@@ -255,7 +255,7 @@ jobs:
       - run: php scripts/directory-yaml-parse.php
       # https://architecture.lullabot.com/adr/20211201-use-strict-types-in-php-code/
       # New php files must use strict types.
-      - run: scripts/phpcs-new-files
+      # - run: scripts/phpcs-new-files
```
tom-fleming: @arthurbaghdas Is this meant to be commented out?

scripts/phpcs-new-files
```diff
@@ -10,4 +10,5 @@ git diff-tree \
             -z \
             --name-only \
             --no-commit-id \
+            --diff-filter=ACMRTUXB \
```
tom-fleming, line 13: @arthurbaghdas Is this a change that we still want in this file? Seems unrelated to the ticket.

arthurbaghdas: I tried to fix deleted file issue related to the strict types check without success and left here to revisit later

.circleci/config.yml
arthurbaghdas: Yes, this was communicated with Moshe, and we decided to revisit this in a separate ticket which is already in a spec state

tom-fleming:
LGTM

PR #1637, opened 2022-07-05
DP-25325 Fix 403 white screen

tom-fleming:
**Description:**
This PR updates monolog to address an issue 403 errors that only appeared in Acquia environments caused by a bug in Monolog that has since been fixed.

DP-25325

**To Test:**
- [ ] Go to a page that should trigger a 403 error (e.g. /admin, /qag-service3/edit). This change is deployed to feature5.

---


weitzman:
nice find. any idea why the fatal is not showing in the PHP Error log? I downloaded the log from acquia ui and dont see it https://docs.acquia.com/cloud-platform/monitor/logs/php-error/.

tom-fleming:
@weitzman I could only get the error to display after changing the setting on the "Logging and errors" configuration page (so, enabling `display_errors`). I think both the return of a 200 response and the lack of reporting to the PHP error log likely have something to do with Monolog's error handling. 

This is what I was seeing though:
```
TypeError: Cannot access offset of type string on string in MonologHandlerNewRelicHandler->write() (line 93 at vendor/monolog/monolog/src/Monolog/Handler/NewRelicHandler.php)
```

I also noticed that the default `error_reporting` setting changed in PHP 8.0, but I would expect to see more, not fewer, errors with this change: https://www.php.net/manual/en/errorfunc.configuration.php#ini.error-reporting

tom-fleming:
Closing in favor of a hotfix PR: https://github.com/massgov/openmass/pull/1639

PR #1646, opened 2022-07-12
DP-25204: Add confirmation warning when bulk transitioning documents

arthurbaghdas:
**Description:**

DP-25204

**To Test:**

---


composer.json
```diff
@@ -430,7 +430,8 @@
                 "Filter by redirect destination does not work with URL alias (https://www.drupal.org/project/redirect/issues/2981544)": "https://www.drupal.org/files/issues/2021-12-09/2981544-9.patch"
             },
             "drupal/views_bulk_operations": {
-                "Remove the select/deselect all results within VBO": "patches/DP-11179-hide-select-all-results-from-views.patch"
+                "Remove the select/deselect all results within VBO": "patches/DP-11179-hide-select-all-results-from-views.patch",
+                "Sometimes the selected items get lost -> all items in the view get processed (https://www.drupal.org/project/views_bulk_operations/issues/3277611)": "https://www.drupal.org/files/issues/2022-07-25/views_bulk_operations_prevent_all_items_processing_3277611_19.patch"
```
weitzman, line 439: I dont think we should support "quickly going back and forth several times" when running a batch. Therefore, we should remove this patch. Patches add maintenance burden, and the use case isnt strong enough hereIMO to justify it.

arthurbaghdas: We are just fixing an issue when going back to the main view data is not synced in the summary section and checkboxes, so we force uncheck all the checkboxes in that condition. See more info here https://www.drupal.org/project/views_bulk_operations/issues/3277611

weitzman: Thanks. I saw that issue. My opinion is still that real users will rarely/never encounter that bug. Or I am misunderstanding its scope.

PR #1660, opened 2022-07-19
Fix for autocomplete errors on non-English node selection and move children paths

tom-fleming:
**Description:**
Fix for autocomplete errors on non-English node selection and move children paths.

DP-25463

**To Test:**
- [ ] Go to a Move Children page (e.g. https://mass.local/code-of-massachusetts-regulations-cmr/move-children), using autocomplete and confirming that entity type information appears before the title in results.
- [ ] Go to a node with language selection (e.g. https://mass.local/info-details/covid-19-response-reporting/edit), using autocomplete and confirming that entity type information appears before the title in results.

---


docroot/modules/custom/mass_fields/src/EntityAutocompleteMatcher.php
```diff
@@ -66,9 +66,10 @@ public function getMatches($target_type, $selection_handler, $selection_settings
       foreach ($entity_labels as $bundle => $values) {
         foreach ($values as $entity_id => $label) {
           if (!isset($entity_type_labels[$bundle])) {
-            $bundle_info = $handler->entityTypeBundleInfo->getBundleInfo($target_type)[$bundle];
-            if (isset($bundle_info)) {
-              $entity_type_label = $bundle_info['label'];
+            $node_type = $this->entityTypeManager->getStorage('node_type')
```
weitzman: `node_type` looks odd. i would expect an entity type like 'node', 'media', etc.

tom-fleming: @weitzman This is what we had in place before a recent round of New Relic-related fixes, and it seems to be the most direct way to retrieve bundle information for nodes without loading each entity, since using entityTypeBundleInfo doesn't appear to be available via the handler.

docroot/modules/custom/mass_fields/src/EntityAutocompleteMatcher.php
```diff
@@ -66,10 +66,19 @@ public function getMatches($target_type, $selection_handler, $selection_settings
       foreach ($entity_labels as $bundle => $values) {
         foreach ($values as $entity_id => $label) {
           if (!isset($entity_type_labels[$bundle])) {
-            $bundle_info = $handler->entityTypeBundleInfo->getBundleInfo($target_type)[$bundle];
-            if (isset($bundle_info)) {
-              $entity_type_label = $bundle_info['label'];
-              $entity_type_labels[$bundle] = $entity_type_label;
+            if (isset($handler->entityTypeBundleInfo)) {
```
tom-fleming, line 69: This solution is less than ideal, but it does allow both Move Children page and the English version field to be functional again.

These two autocompletes are different in that they use two different views and two different base entities. The Move Children view, which uses taxonomy, has entityTypeBundleInfo available in the handler, but the English version field does not. The Move Children view is unable to retrieve bundle information using entityTypeManager.

arthurbaghdas: I think this approach is ok here. entityTypeBundleInfo is not available I think we can keep this approach

PR #1670, opened 2022-07-25
DP-25519: Search in some collections doesn't show expected results

arthurbaghdas:
**Description:**

DP-25519

**To Test:**

---


weitzman (approved):
Patch looks good. I've just granted @arthurbaghdas maintainership at drupal.org. Please commit the patch and update composer.lock so we use a newer views porter stemmer. Then its OK to merge this, as its approved.

PR #1674, opened 2022-08-01
DP-24803: Fix page ready event for leaflet map in Backstop

arthurbaghdas:
**Description:**

DP-24803

**To Test:**

---

### DIFF
--- a/backstop/scripts/ready.js
+++ b/backstop/scripts/ready.js
@@ -134,9 +134,10 @@ module.exports = async function (page, scenario, vp) {
       visible: true,
       timeout: 10000,
     })
+    await page.waitForTimeout(1000);
   }
   else if (leafletMapInitialized == true) {
-    await page.waitForTimeout(1000);
+    await page.waitForTimeout(2000);
   }
 
   if (!scenario.hideAlerts || scenario.hideAlerts === undefined) {
--- a/backstop/scripts/ready.js
+++ b/backstop/scripts/ready.js
@@ -115,7 +115,7 @@ module.exports = async function (page, scenario, vp) {
 
   let leafletMapInitialized = await page.evaluate(async function () {
     let initialized = undefined;
-    var containers = document.querySelectorAll(".js-leaflet-map");
+    const containers = document.querySelectorAll(".js-leaflet-map");
     if (containers.length) {
       containers.forEach(function (e) {
         var container = L.DomUtil.get(e);
--- a/backstop/scripts/ready.js
+++ b/backstop/scripts/ready.js
@@ -123,10 +123,12 @@ module.exports = async function (page, scenario, vp) {
           if (container._leaflet_id == null) {
             initialized = false;
           }
+          else {
+            initialized = true;
+          }
         }
       });
     }
-    initialized = true;
   })
 
   if (leafletMapInitialized == false) {
--- a/backstop/scripts/ready.js
+++ b/backstop/scripts/ready.js
@@ -118,7 +118,7 @@ module.exports = async function (page, scenario, vp) {
     const containers = document.querySelectorAll(".js-leaflet-map");
     if (containers.length) {
       containers.forEach(function (e) {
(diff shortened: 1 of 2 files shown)


backstop/scripts/ready.js
```diff
@@ -113,6 +113,33 @@ module.exports = async function (page, scenario, vp) {
     await page.waitForSelector('.cbFormErrorMarker', {visible: true})
   }
 
+  let leafletMapInitialized = await page.evaluate(async function () {
+    let initialized = undefined;
+    var containers = document.querySelectorAll(".js-leaflet-map");
```
clairesunstudio:
```suggestion
    const containers = document.querySelectorAll(".js-leaflet-map");
```

backstop/scripts/ready.js
```diff
@@ -113,6 +113,33 @@ module.exports = async function (page, scenario, vp) {
...
+    if (containers.length) {
+      containers.forEach(function (e) {
+        var container = L.DomUtil.get(e);
```
clairesunstudio:
```suggestion
        const container = L.DomUtil.get(e);
```

backstop/scripts/ready.js
```diff
@@ -113,6 +113,33 @@ module.exports = async function (page, scenario, vp) {
...
+        if (container != null) {
```
clairesunstudio, line 122: can you add some notes to these conditions?  

I feel like the condition should be reserved - currently, `if (containers.length)` is false, then `initialized` is set to true - that means for pages that don't have leaflet maps, we are adding a 5 second timeout ?

backstop/scripts/ready.js
```diff
@@ -113,6 +113,33 @@ module.exports = async function (page, scenario, vp) {
...
+        }
+      });
+    }
+    initialized = true;
+  })
+
+  if (leafletMapInitialized == false) {
+    await page.waitForSelector('.js-leaflet-map .leaflet-tile-container', {
+      visible: true,
+      timeout: 10000,
```
clairesunstudio, line 137:
```suggestion
      timeout: 10000,
```
we are adding a 10-second timeout here? is that too much?

backstop/scripts/ready.js
```diff
@@ -113,6 +113,33 @@ module.exports = async function (page, scenario, vp) {
...
 
+  let leafletMapInitialized = await page.evaluate(async function () {
+    let initialized = undefined;
+    var containers = document.querySelectorAll(".js-leaflet-map");
+    if (containers.length) {
+      containers.forEach(function (e) {
+        var container = L.DomUtil.get(e);
+        if (container != null) {
+          if (container._leaflet_id == null) {
+            initialized = false;
```
clairesunstudio, line 124:
```suggestion
            initialized = false;
```

I'm a little confused by the logic here - by chaining these conditions `containers.length` && `container != null` && `container._leaflet_id == null` means we are only setting `initialized` to false when all three conditions are met, and if any is unmet we are setting `initialized` to true?

backstop/scripts/ready.js
arthurbaghdas: That's the maximum time that we wait for an element to become visible, it is usually less then 2 second.

backstop/scripts/ready.js
arthurbaghdas: If the map is not initialized   container._leaflet_id returns null, that's the condition we need to establish a wait for the map to become visible

backstop/scripts/ready.js
arthurbaghdas: We are not running this on the pages that don't have a map.

clairesunstudio:
redeployed to feature1 and ran backstop --> https://output.circle-artifacts.com/output/job/373b0e50-ee80-4d04-af64-08bfe8b78b64/artifacts/0/home/circleci/code/backstop/report/index.html
none of the diffs are related to maps. Approved!

PR #1697, opened 2022-08-31
DP-25755: Fix behat test failure

arthurbaghdas:
**Description:**

DP-25755

**To Test:**

---


weitzman:
1. composer.lock is changing but I think thats a mistake?
2. Please add code comments about why we are doing unset() and when we can remove it.

arthurbaghdas:
> 1. composer.lock is changing but I think thats a mistake?
> 
>     2. Please add code comments about why we are doing unset() and when we can remove it.

Reverted composer.lock changes, added comments.

PR #1699, opened 2022-09-01
Fix updated_date render in event node

weitzman:
Also make event rendering code more readable via a few methods on the bundle class.

DP-25762

---


tom-fleming:
LGTM

PR #1700, opened 2022-09-05
DP-25692: Hide "Media" and remove "Moderated Content" from All Content view

arthurbaghdas:
**Description:**

DP-25692

**To Test:**

---


features/documents.feature
```diff
@@ -18,25 +18,6 @@ Feature: Media Documents
     And I attach the file "upload.txt" to "files[field_upload_file_0]"
     And I press "Save"
     Then I should see "Document upload.txt has been created."
-    Given I am on "/admin/content/media"
-    When I follow "upload.txt"
+    Given I am on "/doc/test-document-behat"
     Then I should see "Test document Behat"
     And I should see "(TXT 24 BYTES)"
-
-  Scenario: Create and view a new MP3
```
weitzman, line 26: Are we testing media creation elsewhere? If not lets just remove the last few lines and not the whole scenario

arthurbaghdas: That was requested in the ticket. and we are testing media creation in the same file. See here
https://github.com/massgov/openmass/blob/cf6303f8a3b56902ed5d946403d78b95a61cd1ee/features/documents.feature#L7-L23

PR #1718, opened 2022-09-20
Dp 25832 fix accordion tests in patternlab

clairesunstudio:


clairesunstudio:
pushed to wrong branch

PR #1721, opened 2022-09-25
DP-25897: Change unpublished preview links on to always display a "www.mass.gov" domain

arthurbaghdas:
**Description:**

DP-25897

**To Test:**

---


patches/access_unpublished/DP-25897_change_unpublish_preview_links_v2.patch
```diff
@@ -0,0 +1,78 @@
+diff --git a/access_unpublished.services.yml b/access_unpublished.services.yml
```
tom-fleming, line 1: @arthurbaghdas This seems like a service we should be overriding rather than something we should patch. Is there a reason for going this direction?

arthurbaghdas: I guess the only reason is to to keep track of the changes in the future with the new version updates.

PR #1724, opened 2022-09-28
DP-25830: Cache on location summary pages not clearing when updates made to individual location pages

arthurbaghdas:
**Description:**

DP-25830

**To Test:**

---


docroot/themes/custom/mass_theme/mass_theme.theme
```diff
@@ -1277,6 +1277,7 @@ function mass_theme_preprocess_node_location(&$variables) {
 
     if (!$node->field_ref_contact_info_1->isEmpty()) {
       $contact_information_entity = $node->field_ref_contact_info_1->entity;
+      $cache_tags = array_merge($cache_tags, $contact_information_entity->getCacheTags());
```
weitzman: This is a bit confusing. Can $cache_tags be populated at this point? If not, the the array_mergeI() is a null operation. Please just assign `$cache_tags = $contact_...`  direcly unless I'm missing something.

arthurbaghdas: I followed the pattern used in other implementations. Right now we are safe to assign directly, but in the future if someone adds a snippet related to the tags above, we should be careful, cause the values might be overridden.Refactored the code to assign directly. Added a comment for future reference.

PR #1728, opened 2022-09-29
DP-25357: Address entity usage performance, refactoring

tom-fleming:
**To reroll**
Accept changes from develop, then `composer require drupal/entity_usage:^2.0.0-beta7`

**Description:**
The goal of this ticket was to resolve performance issues seen with the entity_usage module. The fix to the specific performance issue was adding a source_type condition to delete queries in entity_usage, but more work was required to do this cleanly, since the delete queries only existed in a combined patch (combining three patches from drupal.org) that included custom code.

This PR accomplishes the following:
- Updates entity_usage from beta5 to beta7
- Removes [301287-entity_usage_queue_tracking.patch](https://github.com/massgov/openmass/pull/1728/files#diff-a4524ef88b62ef55f43b929bc1f200f7d97cbcf5228a1e319670932817c60489), which is no longer needed
- Removes [3101451-3.patch](https://github.com/massgov/openmass/pull/1728/files#diff-435c54519ba6e145bc600d5d30b33de8e9fa6462c949aa669d9de36e58abcd40), which was merged into entity_usage
- Removes [patches/2954973-3015287-3101451.patch](https://github.com/massgov/openmass/pull/1728/files#diff-6ae359a1adddaa4279b8cc79629465311693fc0099ba34fe3f6d15bef684e388), which was a combined patch of three different patches that included custom code
- Updates [entity_usage-usage_list_scale-3007752-25.patch](https://github.com/massgov/openmass/pull/1728/files#diff-6d9586c69d471774ee1a5756cc02c86cb9b33724433428aace40ff50ad734e76) to a version compatible with beta7
- Updates [2954973-68.patch](https://github.com/massgov/openmass/pull/1728/files#diff-4630875552b5457ab6f47638aa1e55b30f4181ca3303630195cfcb675925410a) to a version compatible with beta7 and moves it out of the combined patch
- Moves the queue_tracking and clean table Drush command functionality from the combined patch into a custom module called entity_usage_queue_tracking. The code for the queue_tracking functionality is based on the custom code in the combined patch and Luis's patch here: https://www.drupal.org/project/entity_usage/issues/3015287#comment-14410307 
- Addresses performance issues by adding a source_type conditional to entity_usage delete queries

**About entity_usage_queue_tracking:**
This module is based on [Luis's patch](https://www.drupal.org/project/entity_usage/issues/3015287#comment-14410307), which introduced an alternative to the background_tracking patch proposed in the same issue. The goals are similar, but the difference is that background_tracking is trying to move the work of tracking after a node operation to a background process, whereas this custom module performs that work via a queue. Additionally, the custom module includes some cleanup queries and a cleanup drush command that deletes entity_usage records for both non-current revisions and records pointing to themselves.

**To Do / Questions:**
- Would this be useful on drupal.org as a standalone module? 
- Is deleting non-current revisions really what we want? We should check that we can still see unpublished revisions that are more current than the published version
- Improve test performance in mass_entity_usage

DP-25357

**To Test:**
Based on test steps from https://github.com/massgov/openmass/pull/1589.

---
**Optional Test Setup**

Before testing, to speed up the usage count and ensure you are processing your recent usage records, **only for development environments**, run the following:

> drush sqlq "TRUNCATE entity_usage" # truncate entity usage table
> drush sqlq "DELETE FROM queue WHERE name='entity_usage_tracker'" # delete entity usage tracking queue

---
**Required Test Setup**

Ensure that the environment's settings file includes the following after its initial entity_usage_regenerate_queue run:
`$config['entity_usage_queue_tracking.settings']['queue_tracking'] = TRUE;`

All nodes created for entity usage tests should be published.

---
**Test Steps**

- [ ] Create a node: node1 of any content type that can be related to other content types.
- [ ] Create a node: node2, and link to node1 of any content type that can be related to other content types.
- [ ] Wait 5-10 minutes for the scheduled job to run or run `drush queue:run entity_usage_tracker --time-limit=30`
- [ ] Go to the second tab "Pages linking here" of node1, notice 1 usage record from node2.
---
- [ ] Remove the link on node2.
- [ ] Wait 5-10 minutes for the scheduled job to run or run `drush queue:run entity_usage_tracker --time-limit=30`
- [ ] Go to the second tab "Pages linking here" of node1, notice no usage records.
---
- [ ] Restore the link to node1 on node2.
- [ ] Wait 5-10 minutes for the scheduled job to run or run `drush queue:run entity_usage_tracker --time-limit=30`
- [ ] Go to the second tab "Pages linking here" of node1, notice 1 usage record from node2.
---
- [ ] Delete node2.
- [ ] Wait 5-10 minutes for the scheduled job to run or run `drush queue:run entity_usage_tracker --time-limit=30`
- [ ] Go to the second tab "Pages linking here" of node1, notice no usage records.

---


docroot/sites/default/settings.php
```diff
@@ -220,5 +220,4 @@
   ini_set('memory_limit', '2048M');
 }
 
-$config['entity_usage.settings']['queue_tracking'] = TRUE;
```
tom-fleming, line 223: Just a note on this change: I think this setting should not be defaulted and instead should be set per environment, once we've run the full entity_usage_regenerate_queue queue. This allows us to enable the module during testing and on production while only triggering the queueing of items on node operations when this configuration is set to true.

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
```diff
@@ -45,3 +46,86 @@ function mass_entity_usage_queue_info_alter(&$queues) {
...
+function mass_entity_usage_module_implements_alter(&$implementations, $hook) {
+  $overrides = [
+    'entity_insert',
+    'entity_update',
+    'entity_predelete',
+    'entity_translate_delete',
+    'entity_revision_delete',
+  ];
+
+  // Completely take over entity_usage hooks.
```
weitzman: I havent seen a module just delete another module's hooks before. Is this a common practice? It works reliably?

docroot/modules/custom/mass_entity_usage/tests/src/ExistingSite/EntityUsageTest.php
```diff
@@ -34,6 +34,12 @@ class EntityUsageTest extends ExistingSiteBase {
   protected function setUp() {
     parent::setUp();
 
+    $config_factory = \Drupal::configFactory();
+    $config = $config_factory->getEditable('mass_entity_usage.settings');
+    $this->tracking = $config->get('queue_tracking');
+    $config->set('queue_tracking', TRUE);
+    $config->save();
+
```
weitzman: Instead of changing config, you can override it. See https://github.com/massgov/openmass/blob/develop/docroot/modules/custom/mass_unpublish_reminders/tests/src/ExistingSite/UnpublishRemindersTest.php#L47-L50
. This doesn't mess up a developer's environment if test is stopped before teardown.

docroot/modules/custom/mass_entity_usage/src/Commands/CleanUsageTableCommand.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace Drupal\mass_entity_usage\Commands;
+
+use Drush\Commands\DrushCommands;
+
+/**
+ * Exposes the CleanUsageTable service to clean the usage table.
+ */
+class CleanUsageTableCommand extends DrushCommands {
```
weitzman: Usually commandfiles use the plural form so _CleanUsageTableCommands_.

docroot/modules/custom/mass_entity_usage/README.md
```diff
@@ -0,0 +1,10 @@
+Tracking via queue
```
weitzman: Would it make sense to move this module to Contrib and pull it in that way?

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
tom-fleming: Not a common practice, but it is working reliably. Previously, we had a patch that added conditional logic to the entity usage hooks to use the tracking queue instead of updating entity_usage immediately.

docroot/modules/custom/mass_entity_usage/README.md
tom-fleming: I've gone back and forth on it, but ultimately, I think now that we've some successful runs, yes. I can work on that

tom-fleming: @weitzman I split the module back out to a separate module and made a contrib project for it: https://www.drupal.org/project/entity_usage_queue_tracking/.

composer.json
```diff
@@ -128,6 +128,19 @@
...
+            "package": {
+                "name": "drupal/entity_usage_queue_tracking",
+                "type": "drupal-module",
+                "version": "9.1.x-dev",
+                "source": {
+                    "type": "git",
+                    "url": "https://git.drupalcode.org/project/entity_usage_queue_tracking.git",
+                    "reference": "9290c86e9c1ed10a3f976c9090c7108e8fcc487e"
+                }
+            }
```
tom-fleming, line 143: The contrib module wasn't pulling down, and I'm not entirely certain why. It should be meeting the criteria specified here: https://www.drupal.org/docs/develop/managing-a-drupalorg-theme-module-or-distribution-project/creating-a-new-project/general-projects

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
tom-fleming: This also only occurs if the queue_tracking config is set to true.

docroot/modules/custom/mass_entity_usage/tests/src/ExistingSite/EntityUsageTest.php
tom-fleming: I added this. I am seeing a tearDown step though in UnpublishRemindersTest that appears to be resetting config. Is that necessary here? https://github.com/massgov/openmass/blob/develop/docroot/modules/custom/mass_unpublish_reminders/tests/src/ExistingSite/UnpublishRemindersTest.php#L84

docroot/modules/custom/mass_entity_usage/src/Commands/CleanUsageTableCommand.php
tom-fleming: I fixed this, and moved it to the contrib module.

PR #1731, opened 2022-09-30
Hotifx DP-25875 Bubble max-age to response headers. Fixes event listing staleness

weitzman:
**Description:**
- I've tweaked the patch at https://www.drupal.org/project/drupal/issues/2352009#comment-14456719 so that it works via cache control headers and not expires since expires is ancient and not respected when cache-control is provided.
- When we are satisfied, the patch will be uploaded to drupal.org and this PR changed.
- The selenium-chrome change was needed for me to run client side tests locally

As mentioned in Jira, my hope is to deploy this solo in a hotfix so it can be rolled back easily if Varnish hit rate plummets. Its a bit hard to predict how many pages this will affect. When it does affect, the pages will become less stale which is great as long as the site can withstand the new load. I'm not too worried but it merits a close watch.

DP-25875

**To Test:**
- [ ] Browse the site and look at Cache-Control header using Dev Tools Inspector. You can filter for just HTML  responses. See the value of max-age being lower for pages with event listings.

---


arthurbaghdas (approved):

weitzman:
This was released via hotfix 0.337.1

PR #1738, opened 2022-10-06
DP-25704: Warn authors if they associate an event to a service or org page that doesn't have an events component

arthurbaghdas:
**Description:**

DP-25704

**To Test:**

---


docroot/modules/custom/mass_validation/mass_validation.module
```diff
@@ -925,13 +926,87 @@ function _mass_validation_form_node_regulation_form_validate(array &$form, FormS
...
+          }
+        }
+      }
+    }
+  }
+  return FALSE;
+}
+
+/**
+ * Submit hanlder for event nodes.
```
weitzman:
```suggestion
 * Submit handler for event nodes.
```

docroot/modules/custom/mass_validation/mass_validation.module
```diff
@@ -925,13 +926,87 @@ function _mass_validation_form_node_regulation_form_validate(array &$form, FormS
  */
 function mass_validation_form_node_event_edit_form_alter(&$form, FormStateInterface $form_state, $form_id) {
   $form['#validate'][] = '_mass_validation_form_node_event_form_validate';
+  foreach (array_keys($form['actions']) as $action) {
```
weitzman, line 929: I would think you can just use a form level submit handler instead of a button submit handler. That is, use `$form['#submit'][] = ` instead of `$form['actions'][$action]['#submit'][]`. That can be done without if the loop and the if() check

weitzman (changes requested):

docroot/modules/custom/mass_validation/mass_validation.module
arthurbaghdas: That was my initial thought as well, but it didn't work and here are some discussions on that topic.
https://drupal.stackexchange.com/questions/223342/add-a-custom-submission-handler-to-a-form
https://www.drupal.org/node/1901216
https://www.drupal.org/node/2068063
And here is an example implementation on Drupal.org: https://api.drupal.org/api/drupal/core!modules!menu_ui!menu_ui.module/function/menu_ui_form_node_form_alter/9.4.x

PR #1743, opened 2022-10-12
DP-24550: Display link to document in multiple languages

arthurbaghdas:
**Description:**

DP-24550

**To Test:**

---


weitzman:
This needs cacheability metadata so that when someone adds a translation the links are updated. I suggest moving this code into a computed field on the entity (no storage needed for the field). See https://www.drupal.org/docs/drupal-apis/entity-api/dynamicvirtual-field-values-using-computed-field-property-classes. If you grep our code for `setComputed(TRUE)` you will find a instances you can copy.

weitzman (changes requested):
This needs cacheability metadata so that when someone adds a translation the links are updated. I suggest moving this code into a computed field on the entity (no storage needed for the field). See https://www.drupal.org/docs/drupal-apis/entity-api/dynamicvirtual-field-values-using-computed-field-property-classes. If you grep our code for setComputed(TRUE) you will find a instances you can copy.

As usual for fields, you would get the data in one method and do the rendering/themeing in a formatter.

arthurbaghdas:
> 

To be honest, I'm not really sure which part of the code you want to change into a computed field implementation. We are rendering a view mode, but changing the implementation means we should change the configs for multiple paragraphs and nodes (Which can be a bit risky in the current stage of this task). I think we can keep the current implementation and refactor it in a separate ticket if that has a huge architectural advantage. Re architecturing this means starting the task from scratch I think.
Also I've added the caching mechanism for the component.

weitzman:
I am suggesting that the list of links be a new computed field.

Once that field is working, we have an array of links thats available when rendering media (preprocess, twig, etc). 

The change I'm suggesting should not be risky. The logic is not changing, just better code modularization.

I'm happy to take on the computed field change, once tests are passing. You can assign back to me if desired.

arthurbaghdas:
> I am suggesting that the list of links be a new computed field.
> 
> Once that field is working, we have an array of links thats available when rendering media (preprocess, twig, etc).
> 
> The change I'm suggesting should not be risky. The logic is not changing, just better code modularization.
> 
> I'm happy to take on the computed field change, once tests are passing. You can assign back to me if desired.

Same logic is for the node translation links as well. So ideally, we should change that here as well.
https://github.com/massgov/openmass/blob/570149a218fadcdd181190cead8b38255f6c4c0b/docroot/modules/custom/mass_translations/mass_translations.module#L23-L73

docroot/modules/custom/mass_translations/tests/ExistingSiteJavascript/DocumentMultilangTest.php
```diff
@@ -0,0 +1,125 @@
...
+use Drupal\file\Entity\File;
+use Drupal\user\Entity\User;
+use weitzman\DrupalTestTraits\Entity\MediaCreationTrait;
+use weitzman\DrupalTestTraits\ExistingSiteSelenium2DriverTestBase;
+use weitzman\LoginTrait\LoginTrait;
+
+/**
+ * Multilanguage documents tests.
+ */
+class DocumentMultilangTest extends ExistingSiteSelenium2DriverTestBase {
```
weitzman, line 14: Lets only use ExistingSiteSelenium2DriverTestBase when we have javascript interactions we need to test. Otherwise ExistingSiteBase is faster and simpler (no chrome service needed). You can read about the two types at https://gitlab.com/weitzman/drupal-test-traits#authoring-tests

docroot/modules/custom/mass_translations/tests/ExistingSiteJavascript/DocumentMultilangTest.php
```diff
@@ -0,0 +1,125 @@
...
+    // Create a "Llama" media item.
+    file_put_contents('public://llama-44.txt', 'Test');
+    $this->file = File::create([
+      'uri' => 'public://llama-44.txt',
+    ]);
+
+    // Generate media items.
+    $this->generateMedia();
+
+    // Create a node if type 'binder'.
```
weitzman:
```suggestion
    // Create a node of type 'binder'.
```

docroot/modules/custom/mass_translations/tests/ExistingSiteJavascript/DocumentMultilangTest.php
```diff
@@ -0,0 +1,125 @@
...
+    // Example: Español.
+    $this->translatedLangLabel = $predefined[$langcode][1];
+    $this->media = $media;
+    $this->translatedMedia = $translatedMedia;
+  }
+
+  /**
+   * Test Multilanguage document functionality.
+   */
+  public function testHasTranslationLinks() {
```
weitzman: Lets enhance the test so that we

1. Create a document without any translations initially. 
2. User views the document, assert that no translation download links are there.
3. Create translations that are affiliated with the document.
4. Assert that user now sees translation download links when viewing document. That shows that saving a translation clears the render cache of its parent entity

weitzman (changes requested):
Thanks for adding a test. A couple enhancement requests ... FYI, I am starting on the computed field implementation in another branch

docroot/modules/custom/mass_translations/tests/ExistingSiteJavascript/DocumentMultilangTest.php
arthurbaghdas: We need to test JS interaction as well, I've used ExistingSiteBase initially, but when reached to a point to test the Toggle translation labels functionality converted to be ExistingSiteSelenium2DriverTestBase.

docroot/modules/custom/mass_translations/tests/ExistingSiteJavascript/DocumentMultilangTest.php
arthurbaghdas: All set.

docroot/modules/custom/mass_translations/mass_translations.module
```diff
@@ -13,10 +14,99 @@ use Drupal\Core\Language\LanguageManager;
...
+              // Invalidate english version tags.
+              Cache::invalidateTags($en_mid->getCacheTags());
+            }
+          }
+        }
+      }
+    }
+  }
+}
+
```
weitzman: I think you can get rid of this whole function. Instead, make sure that the list of links has the `media-list:document` cache tag ([change record](https://www.drupal.org/node/3107058)) in addition to its existing cache tags. You can do that in mass_translations_preprocess_media(). This list cache tag is already cleared by core when a document is saved. See https://github.com/massgov/openmass/pull/1500 for an example from our site. The test instructions on that PR help check that the list tag was successfully added.

docroot/modules/custom/mass_translations/mass_translations.module
```diff
@@ -13,10 +14,99 @@ use Drupal\Core\Language\LanguageManager;
...
+            ];
+          }
+        }
+      }
+    }
+    if (count($languages) >= 1) {
+      $inlineLinks = Organisms::prepareInlineLinksForLanguages($languages, ["ariaLabel" => "Available Translations"]);
+      $variables += ["translationLinks" => $inlineLinks];
+    }
+    $variables['#cache']['tags'] = Cache::mergeTags($media->getCacheTags(), $cache_tags);
```
weitzman: I think the official way to do this is at https://github.com/massgov/openmass/pull/1500/files#diff-019e1a31e0240993bbdbb711689f78b2f7631c57e726d78ab81e5156b1266821R5122-R5126. We don't set `    $variables['#cache']['tags']` directly. @deviantintegral showed me this in #1500

weitzman (changes requested):

docroot/modules/custom/mass_translations/mass_translations.module
arthurbaghdas: Removed the function.

docroot/modules/custom/mass_translations/mass_translations.module
arthurbaghdas: Changed, as suggested

weitzman (approved):

tom-fleming:
LGTM

PR #1758, opened 2022-10-26
DP-26263: Do not add or change JS org variable tracking when on a top…

arthurbaghdas:
…ic page

**Description:**

DP-26263

**To Test:**

---

### DIFF
--- a/docroot/modules/custom/mass_content/mass_content.module
+++ b/docroot/modules/custom/mass_content/mass_content.module
@@ -537,14 +537,7 @@ function mass_content_theme() {
  */
 function mass_content_page_attachments(array &$attachments) {
   $attachments['#attached']['library'][] = 'mass_content/contextual-nav';
-  $route_name = \Drupal::routeMatch()->getRouteName();
-  if ($route_name === 'entity.node.canonical') {
-    $attachments['#cache']['contexts'][] = 'route.name:entity.node.canonical';
-    $entity = \Drupal::routeMatch()->getParameter('node');
-    if ($entity && $entity->bundle() !== 'topic_page') {
-      $attachments['#attached']['library'][] = 'mass_content/session_organizations';
-    }
-  }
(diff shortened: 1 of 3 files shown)


docroot/modules/custom/mass_content/mass_content.module
```diff
@@ -537,7 +537,13 @@ function mass_content_theme() {
  */
 function mass_content_page_attachments(array &$attachments) {
   $attachments['#attached']['library'][] = 'mass_content/contextual-nav';
-  $attachments['#attached']['library'][] = 'mass_content/session_organizations';
+  $route_name = \Drupal::routeMatch()->getRouteName();
+  if ($route_name === 'entity.node.canonical') {
+    $entity = \Drupal::routeMatch()->getParameter('node');
+    if ($entity && $entity->bundle() !== 'topic_page') {
+      $attachments['#attached']['library'][] = 'mass_content/session_organizations';
```
weitzman: I think you also have to add cacheability metadata or else the dynamic page cache for a topic_page could be used on a non-topic page and vice versa. See https://www.drupal.org/project/drupal/issues/2478393 for an example patch which adds this data.

arthurbaghdas: I've added cache context to the route entity.node.canonical.

weitzman: I'm not sure thats right. That context means that we will have different caches for node-canonical pages and non-node-canonical. But the code attaches the library based on topic page versus non-topic page. So node-canonical is not the right bucket.

However, I dont see a per-bundle cache context in drupal core. So maybe I am missing something. Can we always attach mass_content/session_organizations library but have the javascript not take action for topic pages? Javascript has access to entity type via Drupal.settings object, I think.

arthurbaghdas: Refactored the code according to the request.

weitzman (approved):

PR #1766, opened 2022-10-31
Add Drupal core patch to suppress error about deleting a table that no longer exists

weitzman:
See linked drupal.org issue. Using the first patch there since its the most tightly scoped to what we need.

DP-26287

**To Test:**
- [ ] None.

---


tom-fleming:
LGTM

PR #1767, opened 2022-11-01
DP-26299: Update page/org alert rendering to include Updated

arthurbaghdas:
**Description:**

DP-26299

**To Test:**

---


docroot/modules/custom/mass_alerts/src/Controller/AlertsController.php
```diff
@@ -340,7 +340,7 @@ public function handlePageRequest($nid, Request $request) {
           $timestamp = '';
         }
 
-        $alert['suffix'] = $timestamp;
+        $alert['suffix'] = $this->t('Updated ') . $timestamp;
```
weitzman:
```suggestion
        $alert['suffix'] = $this->t('Updated @timestamp', ['@timestamp' => $timestamp]);
```

weitzman (changes requested):
If we ever translate the site, it is helpful to give translator context about how a string is used.

docroot/modules/custom/mass_alerts/src/Controller/AlertsController.php
arthurbaghdas: Thanks for the note, all set.

weitzman (approved):

PR #1771, opened 2022-11-01
DP-26115-mainnav-keyboard-arrow-bug

clairesunstudio:
Fixed:
  - description: Fix mainNav submenu skipping the first item using arrow keys.
    issue: DP-26115


ygannett (approved):
LGTM

PR #1775, opened 2022-11-02
DP-26284: Un-install jsonapi_page_limit module to fix jsonapi page li…

sagar-lcm:
**Description:**
Fix for jsonapi_page_limit module to respect page offset and limit parameters.

https://massgov.atlassian.net/browse/DP-26284

**To Test:**
- Test one of the JSON API endpoints with page[offset] and page[limit] parameters.


tom-fleming (approved):
LGTM

weitzman:
The page limit module increases the maximum number of items in a page to 250 (IIRC). Perhaps you tried that and it was not working due to https://www.drupal.org/project/jsonapi_page_limit/issues/3318412#comment-14785343. I just merged that issue, so perhaps update the module to see if it works better ... If we just remove the module then all callers will be reduced to very few items in a call. Maybe that doesnt matter since alerts no longer use JSON API. I'm fine either way.

sagar-lcm:
@weitzman The module behaved the same way for the limit of 50 items as well, I think if we are not using the increased limit for any of the available APIs then it is better to uninstall it else we can go ahead with the patch.

weitzman:
Approved.

PR #1788, opened 2022-11-16
DP-26434: Disable search autocomplete tests

tom-fleming:
**Description:**
This PR removes search autocomplete tests.

DP-26434

**To Test:**
- [ ] Check that phpunit tests are passing

---


weitzman:
The way we have handled temporary disabling in the past is to put [this line](https://github.com/massgov/openmass/blob/d129a4b81cff102fee114f08f7763c946bc4fc0d/docroot/modules/custom/mass_alerts/tests/src/ExistingSiteJavascript/OrganizationAlertsClientSideTest.php#L24) at top of the test. That way the file stays in the repo and reminds us to re-enable when possible.

PR #1822, opened 2022-12-23
DP-26237 collapsable content labels

ygannett:
**Description:**

DP-26237

**To Test:**
- [ ] Test sample page:  /pandemic-unemployment-assistance-pua
See https://github.com/massgov/mayflower/pull/1703 for testing.

---


clairesunstudio:
Maybe we need to set the `state` prop on openmass side? I noticed that the expanded state is not announced in VO initially. I wonder if that's related.

ygannett:
> Maybe we need to set the `state` prop on openmass side? I noticed that the expanded state is not announced in VO initially. I wonder if that's related.

I tested the state with VO in [Tugboat](https://pr1822-hsbydecv0fsokwy6tw1ofjy8nymn4od8.tugboatqa.com/pandemic-unemployment-assistance-pua) after updating the MF version.  It announces the state with no change in openmass side.  It should be all set.
openmass

clairesunstudio (approved):

PR #1844, opened 2023-01-19
DP-27004: Security Update

arthurbaghdas:
**Description:**

DP-27004

**To Test:**

---


composer.json
```diff
@@ -171,9 +171,9 @@
         "drupal/composer_deploy": "^1.4",
         "drupal/conditional_fields": "^4.0@alpha",
         "drupal/config_ignore": "3.x-dev",
-        "drupal/core-composer-scaffold": "^9.4",
-        "drupal/core-project-message": "^9.4",
-        "drupal/core-recommended": "^9.4",
+        "drupal/core-composer-scaffold": "~9.4.10",
+        "drupal/core-project-message": "~9.4.10",
+        "drupal/core-recommended": "~9.4.10",
         "drupal/csv_field": "^1.0@beta",
```
weitzman, line 177: We don't use the `~` operator elsewhere in this file, and we have not needed it for years. So I'm curious why we would introduce it now? What I would expect is no changes at all for the ^9.4 lines in composer.json. You can run `composer update drupal/core-recommended drupal/core-dev etc.`. This will change composer.lock but not change composer.json. If you are fuzzy on the difference between these two files see https://www.engineyard.com/blog/composer-its-all-about-the-lock-file/

arthurbaghdas: `~` is temporary here. I had to add that to the composer.json in order to keep the minor version same. If we keep the `"^9.4"` in the composer.json it will try to update to `9.5.2` version of the drupal core. Adding `~` to the composer.json will let us keep the patch level updates for 9.4 version and we can switch back to `^` whenever we decide to the minor version update. 
I know the difference of the 2 files, thanks for the article.

weitzman:
That makes sense, thanks ... Is there a test failure or other problem with upgrading to 9.5? I think we should do that since thats what the Drupal 10 upgrade requires.

weitzman (approved):

arthurbaghdas:
> That makes sense, thanks ... Is there a test failure or other problem with upgrading to 9.5? I think we should do that since thats what the Drupal 10 upgrade requires.

Didn't try updating to 9.5

weitzman:
I merged this PR. Lets do 9.5 the next time we have to update core.

PR #1855, opened 2023-01-25
DP-26781: Gtranslate 2

arthurbaghdas:
**Description:**

DP-26781

**To Test:**

---


docroot/themes/custom/mass_theme/templates/layout/html.html.twig
```diff
@@ -58,6 +58,9 @@
       try {document.addEventListener("DOMContentLoaded", $buo_f,false)}
       catch(e){window.attachEvent("onload", $buo_f)}
     </script>
+      <div class="gtranslate_wrapper"></div>
+      <script>window.gtranslateSettings = {"default_language":"en","native_language_names":true,"languages":["en","fr","de","it","es"],"wrapper_selector":".gtranslate_wrapper"}</script>
```
clairesunstudio:
```suggestion
      <script>window.gtranslateSettings = {"default_language":"en","native_language_names":true,"languages":["en","fr","de","it","es","zh-CN"],"wrapper_selector":".gtranslate_wrapper"}</script>
```

clairesunstudio:
It seems with the widget implementation, we don't get access to the more advanced features like translation edits and URL translation. This implementation seems to be using exactly the same google translate widget that we are using now - but with a custom frontend. Maybe we should figure out how they do that, then we don't need a third party service. 

However, I think it's worth testing out the more advanced features they offer - I'm particularly interested in the edit translation feature and search engine indexing. I think we will need to setup subdomain to do those.

clairesunstudio:
I contacted them to point the domain to feature5 instead. However, I wasn't able to follow their setup instructions on feature 5 - getting this when I try to install the module, and not seeing the module appear under languages and regions under configuration
Screen Shot 2023-01-31 at 9 40 59 AM
https://my.gtranslate.io/setup-wizard

I see you had another PR that was able to add the module in tugboat going the module route, but the language picker doesn't work.

PR #1867, opened 2023-02-01
Remove and block non-Azure logins for Drupal

weitzman:
- Also ups the autologout duration to be closer to Azure AD duration
- Also disabled basic auth on Tugboat as not recommended anymore https://docs.tugboatqa.com/faq/common-questions/#can-i-password-protect-my-preview-urls
- Uninstalls TFA and password policy modules

DP-27097

**To Test:**
- [ ] Make sure login page no longer shows a form

**Followup**
- CX should update the email templates in production. No need to backport those - they are excluded from config sync.
- Possible updates needed to Gitbook

---


docroot/sites/default/settings.tugboat.php
```diff
@@ -1,21 +1,23 @@
...
-  $is_endpoint = strpos($_SERVER['REQUEST_URI'], '/api/v1/') !== FALSE;
-  if (!$is_oauth && !$is_endpoint && !(isset($_SERVER['PHP_AUTH_USER']) && ($_SERVER['PHP_AUTH_USER']==$username && $_SERVER['PHP_AUTH_PW']==$password))) {
-    header('WWW-Authenticate: Basic realm="This site is protected"');
-    header('HTTP/1.0 401 Unauthorized');
-    // Fallback message when the user presses cancel / escape
-    echo 'Access denied';
-    exit;
-  }
-}
+//$agent = $_SERVER['HTTP_USER_AGENT'];
```
arthurbaghdas, line 10: Should we revert this change? @weitzman

weitzman: No revert needed. Its deliberate as per https://docs.tugboatqa.com/faq/common-questions/#can-i-password-protect-my-preview-urls.. I decided to comment instead of removing the code just so we have a settings file for future customization.

PR #1880, opened 2023-02-08
DP-26251 a11y remove empty list from section links

ygannett:
**Description:**
[[All work is done in Mayflower.](https://github.com/massgov/mayflower/pull/1712)]
- Remove an empty list.
- Set up a button for accordion trigger separating from the link to prevent screen readers to announce aria content for the button with the link.
- Remove the "+" symbol from CSS for the accordion, set up as html and hide it from screen readers.

DP-26251

**To Test:**
- [ ] Go to /orgs/masshealth.
- [ ] Scroll down to What would you like to find? Open inspect.
- [ ] Find no empty `ul` when no list items.
before
after

- [ ] Check a section link heading desktop.  It's a link has no aria attributes for accordion button. So, screen readers don't announce the aria for accordion button.
link

- [ ] Change the view to mobile.  Instead of a link, a button is visible for accordion.
- [ ] "+" symbol is converted to html from css and hidden from screen readers. It's not announced as "plus" anymore.
button

---


ygannett:
The validate error is from composer.loc.  Will be resolved when Mayflower is set back to develop. 

https://app.circleci.com/pipelines/github/massgov/openmass/23577/workflows/27411105-6ec7-4704-ab8c-46fc181f7e12/jobs/100584

PR #1888, opened 2023-02-15
Upgrade Drush for more robust deployments

weitzman:
Also upgrades typhonius/acquia-php-sdk-v2 in pep for D10 dependencies

DP-27234

---

### DIFF
--- a/drush/Commands/DeployCommands.php
+++ b/drush/Commands/DeployCommands.php
@@ -2,8 +2,12 @@
 
 namespace Drush\Commands;
 
-use AcquiaCloudApi\CloudApi\Client;
-use AcquiaCloudApi\CloudApi\Connector;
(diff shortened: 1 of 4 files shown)


drush/Commands/DeployCommands.php
```diff
@@ -336,44 +293,8 @@ public function deploy($target, $git_ref, array $options = ['cache-rebuild' => T
     $process = Drush::drush($targetRecord, 'deploy', [], ['verbose' => TRUE]);
     $process->mustRun($process->showRealtime());
 
-    // Get a list of all an environment's domains.
-    // Note: This also returns load balancer URLs.
```
weitzman, line 340: The --varnish option has never been available so removing this dead code

justafish (approved):
lgtm, cron runs fine locally after the drush upgrade

PR #1892, opened 2023-02-17
DP-26079 Search suggestions

ygannett:
**Description:**
Adjust attributes on the search component for accessibility:

- Add `role="search"` to `form`
- Add info for autosugget for screen reader users.
- Change `type` value from `text` to `search` on `input.ma__header-search__input`
- Change `aria-own` to `aria-controls` on `input.ma__header-search__input`
- Add `inputmode="search"` to `input.ma__header-search__input`
- Add `aria-describedby="autosuggest-info"` for info in case no suggestion on `input.ma__header-search__input`
- Add `aria-expanded="false"`
- Set up status update for `aria-expanded` via js

DP-26079

**To Test:**

- https://pr1892-fzhdqw4xk6r6q3j0padv17c12xvuuobm.tugboatqa.com/
- https://pr1892-fzhdqw4xk6r6q3j0padv17c12xvuuobm.tugboatqa.com/topics/health-social-services

Test with desktop (the search component is in the header and the banner.) and mobile (under the menu flyout) versions.

In inpsect, 
- [ ] Find above markup changes.
- [ ] Find when the page is loaded, `aria-expanded="false"`

pageload

- [ ] Find when suggestions are presented, `aria-expanded="true"`

suggestions

- [ ] Find when suggestions are not presented, `aria-expanded="false"`

no-suggestions

[optional]
- [ ] Test in screen readers.
  - Voiceover doesn't announce the `aria-describedby` content (NVDA and JAWS do).
  - VO users know suggestions are available when they move onto the next element, which is the suggestion option list.
  - NVDA and JAWS announce how many suggestions are available (VO doesn't)
---


docroot/themes/custom/mass_theme/overrides/js/search.js
```diff
@@ -0,0 +1,40 @@
...
+        setTimeout(function () {
+          // div#suggestions-list remains after its child elements are removed.
+          // Check it has child elements or not.
+          var suggestionList = suggestionContainer.querySelector('#suggestions-list').hasChildNodes();
+
+          if (suggestionList) {
+            activeInput.setAttribute('aria-expanded', true);
+          }
+
+          if (!suggestionList) {
```
clairesunstudio:
```suggestion
          else {
```

clairesunstudio (approved):

docroot/themes/custom/mass_theme/templates/includes/molecules-banner-search.html.twig
```diff
@@ -6,21 +6,24 @@
...
-        aria-own="suggestions-list"
-        placeholder="Search Mass.gov"
-        data-suggest=""
-        type="text" />
+      id="banner-search"
+      name="q"
+      class="ma__header-search__input"
+      type="search"
+      inputmode="search"
+      autocomplete="list"
```
clairesunstudio:
```suggestion
      aria-autocomplete="list"
```

list is not a valid value for autocomplete, see https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

docroot/themes/custom/mass_theme/templates/includes/molecules-banner-search.html.twig
```diff
@@ -6,21 +6,24 @@
...
+      role="combobox"
+      aria-expanded="false"
+      aria-controls="suggestions-list"
+      placeholder="Search Mass.gov"
+      data-suggest=""
```
clairesunstudio, line 25: should we remove this empty attribute? Not sure why it's there

docroot/themes/custom/mass_theme/templates/includes/molecules-header-search.html.twig
```diff
@@ -8,21 +8,24 @@
...
+        name="q"
+        class="ma__header-search__input"
+        type="search"
+        inputmode="search"
+        autocomplete="list"
+        role="combobox"
+        aria-expanded="false"
+        aria-controls="suggestions-list"
+        placeholder="Search Mass.gov"
+        data-suggest=""
```
clairesunstudio, line 27: should we remove this empty attribute?

docroot/themes/custom/mass_theme/templates/includes/molecules-header-search.html.twig
```diff
@@ -8,21 +8,24 @@
...
+        aria-describedby="autosuggest-info" />
       <div class="ma__suggestions" data-suggestions=""></div>
```
clairesunstudio, line 29:
```suggestion
      <div class="ma__suggestions" data-suggestions=""></div>
```

curious if the `data-suggestions` attribute is being used in anyway. I don't see the value change when there's value in the listbox

docroot/themes/custom/mass_theme/templates/includes/molecules-header-search.html.twig
```diff
@@ -8,21 +8,24 @@
...
-          aria-own="suggestions-list"
-          placeholder="Search Mass.gov"
-          data-suggest=""
-          type="text" />
+        id="header-search"
+        name="q"
+        class="ma__header-search__input"
+        type="search"
+        inputmode="search"
+        autocomplete="list"
```
clairesunstudio: same comment here, this is not a valid attribute value pair
```suggestion
        aria-autocomplete="list"
```

clairesunstudio (commented):
please see the above comments for some minor changes. Otherwise looks good!

docroot/themes/custom/mass_theme/templates/includes/molecules-banner-search.html.twig
ygannett: I didn't touch ​`data-suggest​` because I thought that was a part of the Google search component and used to store data via JS or something.  As I typed a query into the input, it remains empty.  As long as the attribute is not required by Google search, we can remove it.

ygannett: With the conversation with Minghua, we keep the attribute for now since we cannot confirm if it's required by Google or not.

docroot/themes/custom/mass_theme/templates/includes/molecules-header-search.html.twig
clairesunstudio: removing this seems to affect the autocomplete functionality. tabling this for now

docroot/themes/custom/mass_theme/templates/includes/molecules-header-search.html.twig
clairesunstudio: I can't find docs from google about these attributes, but might be related to google autocomplete. tabling this for now until we have further evidence

clairesunstudio (approved):

PR #1894, opened 2023-02-20
DP-26787:  Bring back env indicator in Gin Theme

arthurbaghdas:
**Description:**

DP-26787

**To Test:**

---

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -168,6 +168,7 @@
         "drupal/cache_metrics": "^2",
         "drupal/ckeditor_liststyle": "^1.4",
         "drupal/coffee": "^1",
+        "drupal/color": "^1.0",
         "drupal/components": "^2.4",
         "drupal/composer_deploy": "^1.4",
         "drupal/conditional_fields": "^4.0@alpha",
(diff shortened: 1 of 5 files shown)


weitzman (approved):
LGTM with one minor comment.

composer.json
```diff
@@ -168,6 +168,7 @@
         "drupal/cache_metrics": "^2",
         "drupal/ckeditor_liststyle": "^1.4",
         "drupal/coffee": "^1",
+        "drupal/color": "^1.0",
```
weitzman, line 171: Is this deliberate? Its disabled in this PR in core.extensions.yml.

arthurbaghdas: Yes, I've updated environment indicator module, and it removed/disabled color module. So i guess we don't need it. Also it is deprecated and will be removed in Drupal 10, so I think this is a good step forward to the upgrade

weitzman: This PR is *adding* color module to composer.json

arthurbaghdas: Updating the env_indicator removes the module, I've re-added to avoid issues. We need to initially disable it before removing. That's why I've added it back to be able to safely disable the color module from the configs.

weitzman: Makes sense. Feel free to resolve conflict and merge.

PR #1896, opened 2023-02-21
DP-26791:  Move 'watch' from frontend theme to Local Tasks in toolbar in Gin theme

arthurbaghdas:
**Description:**

DP-26791

**To Test:**

---

### DIFF
--- a/changelogs/DP-26791.yml
+++ b/changelogs/DP-26791.yml
@@ -1,0 +1,3 @@
+Changed:
+  - description: Move 'watch' from frontend theme to Local Tasks in toolbar in Gin theme.
+    issue: DP-26791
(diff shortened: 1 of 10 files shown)


weitzman:
The similar links are added in mass_gin_toolbar(). I didnt yet look into how to make the link dynamic to watch/unwatch and respect caching of that.

arthurbaghdas:
That was my initial implementation, i was having problems with cache invalidation and adding max age 0 helped only, i can revert that implementation, i just thought it might be easier to render the block which already has everything

weitzman:
If you put your browser in mobile mode (i.e. narrow viewport) you will see that the toolbar icons are shown and text is not. So we need to be consistent with that behavior for this link.

arthurbaghdas:
Saw that, it remains in the same place, I was suggesting an approach that will be close to what we have now on develop, I'll do the requested changes, thanks for the feedback

PR #1910, opened 2023-03-03
Stop using memcache for entity and render caches.

weitzman:
**Description:**
Stop storing render and entity caches in Memcache per https://docs.acquia.com/cloud-platform/performance/memcached/enable/ and https://github.com/acquia/memcache-settings/blob/main/memcache.settings.php. Also helps address revision confusion in the Jira issue noted below.

Eventually we should probably use the exact settings recommended by Acquia. But lets start a bit slower than that and make sure DB is healthy with new load.

DP-27400

**To Test:**
- [ ] See rows in cache_entity DB table

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## [0.352.2] - March 3, 2023
+
+### Fixed
+- DP-27400: Stop storing entity and render caches in Memcache.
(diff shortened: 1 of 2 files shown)


justafish (approved):
lgtm :+1:

PR #1916, opened 2023-03-07
DP-27447: Document language bar not showing when document description…

arthurbaghdas:
… is shown

**Description:**

DP-27447

**To Test:**

---


weitzman:
Seems like we have a hole in our test coverage. Lets add testing for this and any related functionality. If thats too hard, is OK to skip.

weitzman (commented):
Code looks good. Ideally lets add tests.

docroot/modules/custom/mass_translations/mass_translations.module
```diff
@@ -24,7 +24,7 @@ use Drupal\Core\Entity\EntityInterface;
  */
 function mass_translations_preprocess_media(&$variables) {
   $media = $variables['media'];
-  if ($variables["view_mode"] == 'media_download_link' && $media->hasField('field_media_english_version')) {
+  if (($variables["view_mode"] == 'media_download_link' || $variables["view_mode"] == 'linked_desc' || $variables["view_mode"] == 'manual_desc') && $media->hasField('field_media_english_version')) {
```
weitzman, line 27: technically we need to add cacheability metadata because now this method has different output based on view mode. we didn't have it before this PR so i'm, ok with skipping it for now and adding if it becomes a problem. if you agree, feel free to merge.

weitzman (approved):
thanks for adding tests

PR #1923, opened 2023-03-13
DP-27487: Pages linking here fixes to show all pages for documents

arthurbaghdas:
**Description:**

DP-27487

**To Test:**

---


docroot/modules/custom/mass_media/mass_media.deploy.php
```diff
@@ -0,0 +1,79 @@
...
+ */
+function mass_media_deploy_update_documents_usage(&$sandbox) {
+  $_ENV['MASS_FLAGGING_BYPASS'] = TRUE;
+
+  $query = \Drupal::entityQuery('media')->accessCheck(FALSE);
+  $query->condition('bundle', 'document');
+  $query->condition('field_document_label', '', 'IS NOT NULL');
+
+  if (empty($sandbox)) {
+    // Get a list of all nodes of type event.
```
weitzman:
```suggestion
```

Suggest deleting this old line

weitzman (commented):
Could you say a bit more about the approach here? Specifically, if we rebuild all usage information there is no entity save, so the presave logic will not run. Does that mean we would be missing this usage information?

arthurbaghdas:
> Could you say a bit more about the approach here? Specifically, if we rebuild all usage information there is no entity save, so the presave logic will not run. Does that mean we would be missing this usage information?

When the media entity presave hook is called we are triggering an entity usage track if the label is used in the automatic lists. I'm not sure what will happen if we rebuild the usage. Do you have a better suggestion @weitzman for the approach to the issue?

weitzman:
I dont have a specific suggestion as I havent studied the problem. All other usage types can be calculated without needing to save an entity. I'm. not sure why this is different. @tom-fleming any thoughts?

tom-fleming:
@arthurbaghdas I would actually expect [this](https://github.com/massgov/openmass/pull/1923/files#diff-89f7b1ac92a1b3af5df15dece03aef0f9f8e7c489fd9a4fd16eeb440d90690c4R185-R217) to be its own Track plugin in mass_entity_usage for automatic curated lists, which I think should address Moshe's question. We haven't implemented a custom plugin before, but you can see an example in entity_usage here: https://git.drupalcode.org/project/entity_usage/-/blob/8.x-2.x/src/Plugin/EntityUsage/Track/EntityReference.php.

arthurbaghdas:
> @arthurbaghdas I would actually expect [this](https://github.com/massgov/openmass/pull/1923/files#diff-89f7b1ac92a1b3af5df15dece03aef0f9f8e7c489fd9a4fd16eeb440d90690c4R185-R217) to be its own Track plugin in mass_entity_usage for automatic curated lists, which I think should address Moshe's question. We haven't implemented a custom plugin before, but you can see an example in entity_usage here: https://git.drupalcode.org/project/entity_usage/-/blob/8.x-2.x/src/Plugin/EntityUsage/Track/EntityReference.php.

I've tried that, but it doesn't solve the issue. More details in the ticket.

PR #1928, opened 2023-03-16
DP-27538: Investigate 500 errors experienced by authors

arthurbaghdas:
**Description:**
Adds additional cache metadata to SVG processor class. Fixes issue seen in local development environments.

DP-27538

**To Test:**
- [ ] If you previously used a workaround in your local ddev environment by modifying core code, delete docroot/core and rerun composer install
- [ ] Check that the local environment loads. Run additional smoke tests (i.e. log in, visit different pages, save a page, etc)

---


weitzman:
Its not super clear to me what this PR is doing. The original code in SvgProcessor.php is a service that delegates most calls to $this->inner. Thats a common OO pattern that most devs would recognize. This PR switches so that SvgProcessor.php *extends* HtmlResponseAttachmentsProcessor  which is fine but it still delegates calls to HtmlResponseAttachmentsProcessor. If SvgProcessor *is* an HtmlResponseAttachmentsProcessor, it has no need to delegate to an inner service. It *is* that service. I hope this makes sense.

tom-fleming:
@weitzman After discussing with Arthur a little further and reviewing the code a little more, I think we're going to close this PR for now. We suspect the issue is only occurring on production, and we have not been able to replicate locally. So, the goal of the PR was add some additional cache metadata and make the decorator class more closely resemble core and BigPipe in a low effort approach to see if we could either fix or get more information on the 500 errors reported in this ticket. 

With that said, after reviewing, it doesn't look like any of the methods we were looking to take advantage of are called by just simply extending the `HtmlResponseAttachmentsProcessor` class, and we don't think it's worth the effort to add them to the `processAttachments` implementation here when we cannot replicate the issue.

tom-fleming:
@arthurbaghdas I reopened this PR because this issue is occurring in local environments, and this patch fixes the issue. I think it might be worth revisiting this comment: https://github.com/massgov/openmass/pull/1928#issuecomment-1476877449. I think you changed the class to use _extends_ because this is also what some other attachment processor classes are doing, such as BigPipeResponseAttachmentsProcessor. But, wasn't sure if there was anything else we should double check before committing to this approach.

arthurbaghdas:
Thats correct I e looked at the Bigpipe approach

docroot/modules/custom/mayflower/mayflower.services.yml
```diff
@@ -11,9 +11,9 @@ services:
...
     class: Drupal\mayflower\Render\SvgProcessor
     decorates: html_response.attachments_processor
-    arguments: ['@html_response.attachments_processor.mayflower_inline_svg.inner']
+    arguments: ['@html_response.attachments_processor.mayflower_inline_svg.inner', '@asset.resolver', '@config.factory', '@asset.css.collection_renderer', '@asset.js.collection_renderer', '@request_stack', '@renderer', '@module_handler']
   ajax_response.attachments_processor.mayflower_inline_svg:
     public: false
     class: Drupal\mayflower\Render\SvgProcessor
     decorates: ajax_response.attachments_processor
-    arguments: ['@ajax_response.attachments_processor.mayflower_inline_svg.inner']
+    arguments: ['@ajax_response.attachments_processor.mayflower_inline_svg.inner', '@asset.resolver', '@config.factory', '@asset.css.collection_renderer', '@asset.js.collection_renderer', '@request_stack', '@renderer', '@module_handler']
```
weitzman, line 19: These lines duplicate properties above them

docroot/modules/custom/mayflower/src/Render/SvgProcessor.php
```diff
@@ -25,27 +33,55 @@
...
+
+      // (Note this is copied verbatim from
+      // \Drupal\Core\Render\HtmlResponseAttachmentsProcessor::processAttachments)
+      try {
+        $response = $this->renderPlaceholders($response);
+      }
+      catch (EnforcedResponseException $e) {
+        return $e->getResponse();
+      }
+
```
weitzman, line 84: Have you tried just adding these lines and not the rest of the changes? Does this fix the bug? I'm not sure how to reproduce the bug locally. As far as I can tell this is the only change that matters. The rest just reduces clarity as per my comment above. I see that BigPipe does it this way but occasionally core misses the mark.

PR #1937, opened 2023-03-22
DP-26304 Remove empty aria-label from press teaser

ygannett:
**Description:**
When a press teaser image doesn't have alt value, its container doesn't have empty `aria-label`.

DP-26304

**To Test:**
- [ ] Go to the Recent news & announcements section in a page with press teaser component:  /orgs/massachusetts-department-of-children-families
- [ ] Check the markup for the image and find that `span` for a background image doesn't have `aria-label` and `role="img"`
 See https://github.com/massgov/mayflower/pull/1752 for the actual change in Mayflower.

---

### DIFF
--- a/changelogs/DP-26304.yml
+++ b/changelogs/DP-26304.yml
@@ -1,0 +1,41 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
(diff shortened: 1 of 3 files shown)


clairesunstudio (approved):
backstop failure is not related

PR #1943, opened 2023-03-24
Enable key auth

weitzman:
I also disabled writes via JSON API because noone uses that AFAIK so its an unneeded attack surface.

DP-24499

**To Test:**

---


tom-fleming:
@weitzman It looks like adding the `use key authentication` permission is resulting in some behat and phpunit failures. Could you take a look?

PR #2026, opened 2023-05-03
DP-26782 - Filtering and reporting of feedback flagged as inappropriate

tom-fleming:
**Description:**
- DP-27258 - Allow requests for both flagged and unflagged feedback
- DP-27262 - Create new report of flagged feedback and link from reports page

DP-26792

**To Test:**

---


docroot/modules/custom/mass_feedback_loop/src/Form/MassFeedbackLoopPerNodeForm.php
tom-fleming, line 1: This form class is largely based on the per-node controller it replaces. The biggest addition here (besides submitForm) is the getTitle callback, which fixes the broken links currently in place on the per-node feedback page.

arthurbaghdas (approved):

PR #2043, opened 2023-05-16
Update Geo modules and misc modules to Drupal 10

weitzman:
Note: drupal/geocoder skipped due to Guzzle incompat

DP-27345

**To Test:**
- [ ] Test the modules that are mentioned in the bot comment below. At time of this writing, thats: 

1. google_map_field
2. geofield
3. prepopulate
4. scheduled_transitions
5. maxlength
6. google_tag (aka google tag manager)

---


composer.json
```diff
@@ -418,9 +418,6 @@
                 "CKeditor Plugin - using a combined patch to avoid conflict (https://www.drupal.org/node/2928318)": "https://www.drupal.org/files/issues/2020-10-28/media_entity_download-ckeditor_plugin-2928318-20.patch",
                 "3106062 - Track the usage of Media entity through media_entity_download route with entity_usage": "patches/media_entity_download-track_media_entity_usage-3106062-2.patch"
             },
-            "drupal/maxlength": {
-                "Don't enforce maxlength stop, allow negative count down": "https://www.drupal.org/files/issues/2020-12-07/2916687-allow-negative-count-13.patch"
```
arthurbaghdas, line 422: Why we didn't replace this patch with the latest one compatible with the 2.x version of the module, we might break something by just removing the patch.

conf/drupal/config/google_tag.container.default.yml
```diff
@@ -8,33 +8,8 @@ dependencies:
 id: default
 label: Default
 weight: 0
-container_id: GTM-MPHNMQ
```
arthurbaghdas, line 11: I"m not sure about this change, can you confirm this was exported correctly?

composer.json
```diff
@@ -205,8 +205,8 @@
         "drupal/geofield": "^1",
         "drupal/gin": "^3",
         "drupal/gin_toolbar": "^1",
-        "drupal/google_map_field": "^1",
-        "drupal/google_tag": "^1.4",
+        "drupal/google_map_field": "^2",
```
arthurbaghdas, line 208: Do we know if we are using this module?

arthurbaghdas (changes requested):

composer.json
weitzman: I dont know. If anyone suspects that we are no longer using a module, I suggest opening a research ticket.

composer.json
weitzman: Its a bit hard to follow, but the second to last comment there points points to an issue which is now merged. I refer to [this comment](https://www.drupal.org/project/maxlength/issues/2916687#comment-14576442) which points to https://www.drupal.org/project/maxlength/issues/3259425. So this patch is no longer needed?

conf/drupal/config/google_tag.container.default.yml
weitzman: I can confirm that the update hooks ran (google_tag_update_8200, google_tag_update_8201) and the export happened after that. So technically we look OK. During our call we decided to send this PR back to @joeg8612 for deeper external review of Google Tag Manager.

composer.json
arthurbaghdas: After some investigation I think that we can actually remove this patch, since there is an implementation already in 2.x branch, however that requires additional changes from the UI.

arthurbaghdas (approved):

PR #2047, opened 2023-05-17
DP-28274: Add Analytics NEW tab for authors when editing

arthurbaghdas:
**Description:**

DP-28274

**To Test:**

---


docroot/modules/custom/mass_analytics/src/Controller/MassAnalyticsController.php
```diff
@@ -0,0 +1,105 @@
...
+   *
+   * @param \Drupal\node\NodeInterface $node
+   *   The upcasted node object.
+   *
+   * @return array
+   *   The iframe render array or a no match message render array.
+   */
+  public function build(NodeInterface $node) {
+    $looker_studio_url = $this->state->get('mass_analytics.looker_studio_url', '') . '?params=%7B"nodeId":[node:nid]%7D';
+    $config_url = $this->token->replace($looker_studio_url, ['node' => $node]);
```
weitzman: What about calling this $iframe_url? It gets value not from config but from state.

docroot/modules/custom/mass_analytics/src/Controller/MassAnalyticsController.php
```diff
@@ -0,0 +1,105 @@
...
+   * Access callback for the page.
+   *
+   * @param \Drupal\node\NodeInterface $node
+   *
+   * @return \Drupal\Core\Access\AccessResult
+   *   Return true if the node is one of the list.
+   */
+  public function access(NodeInterface $node) {
+    $access = FALSE;
+    $node->bundle();
```
weitzman: Remove

docroot/modules/custom/mass_analytics/src/Form/ConfigForm.php
```diff
@@ -0,0 +1,83 @@
...
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    $form = parent::buildForm($form, $form_state);
+    $default = $this->state->get('mass_analytics.looker_studio_url', '');
+
+    $form['looker_studio_url'] = [
+      '#type' => 'textfield',
+      '#title' => $this->t('Looker Studio URL'),
+      '#description' => $this->t('Add Looker Studio URL without query string, it will be added in the code.'),
+      '#default_value' => $default,
+    ];
```
weitzman: Why do we get the url from both state and config? If we want to use environment overrides, the config system can do that.

docroot/modules/custom/mass_analytics/src/Form/ConfigForm.php
```diff
@@ -0,0 +1,83 @@
...
+
+    return $form;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
+    $value = $form_state->getValue('looker_studio_url');
+    $this->state->set('mass_analytics.looker_studio_url', $value);
```
weitzman: Here we have a configForm that persists data via State. If we want to use state for some reason, just use a simple form, not a config form.

docroot/modules/custom/mass_analytics/src/Controller/MassAnalyticsController.php
arthurbaghdas: Why?

docroot/modules/custom/mass_analytics/src/Controller/MassAnalyticsController.php
arthurbaghdas: Done

docroot/modules/custom/mass_analytics/src/Form/ConfigForm.php
arthurbaghdas: Changed to use config

docroot/modules/custom/mass_analytics/src/Form/ConfigForm.php
arthurbaghdas: Changed everything to be config

docroot/modules/custom/mass_analytics/src/Controller/MassAnalyticsController.php
weitzman: This line gets the bundle and does nothing with it. Did you want to assign it to a variable? It looks redundant with the line below it.

arthurbaghdas: Right, removed.

docroot/modules/custom/mass_analytics/src/Controller/MassAnalyticsController.php
```diff
@@ -0,0 +1,99 @@
...
+use Drupal\token\Token;
+
+/**
+ * Class MassAnalyticsController.
+ *
+ * @package Drupal\mass_analytics\Controller
+ */
+class MassAnalyticsController extends ControllerBase {
+
+  const SCOPE = [
```
weitzman: The name SCOPE seems unusual to me. I would expect BUNDLES or TYPES.

docroot/modules/custom/mass_analytics/src/Controller/MassAnalyticsController.php
```diff
@@ -0,0 +1,99 @@
...
+   *
+   * @return \Drupal\Core\Access\AccessResult
+   *   Return true if the node is one of the list.
+   */
+  public function access(NodeInterface $node) {
+    $access = FALSE;
+    if (in_array($node->bundle(), self::SCOPE)) {
+      $access = TRUE;
+    }
+    return AccessResult::allowedIf($access);
```
weitzman:
```suggestion
    return AccessResult::allowedIf(in_array($node->bundle(), self::SCOPE));
```

docroot/modules/custom/mass_analytics/src/Controller/MassAnalyticsController.php
arthurbaghdas: Changed

conf/drupal/config/config_ignore.settings.yml
```diff
@@ -2,4 +2,5 @@ _core:
   default_config_hash: UVH1aJ4b44UM-VdPVN7hNNuuVqfReJxwfVeDQH1Hvsk
 ignored_config_entities:
   - mass_utility.settings
+  - mass_analytics.settings
```
weitzman: If your goal with ignoring is env specific overrides, then I suggest adding code to https://github.com/massgov/openmass/blob/develop/docroot/sites/default/settings.acquia.php#L100-L131 instead.

arthurbaghdas: I don't think it's environment specific.

arthurbaghdas: I don't think it's environment specific.

weitzman: OK, whats the motivation for the config ignore?

arthurbaghdas: As Joe mentioned in our discussion about this ticket, we don't want the URL to be publicly available, which will the case if we export it.

weitzman: Aha. What we have done for this in the past is just put a placeholder URL in the config yml file and then do a config override in settings.php like https://github.com/massgov/openmass/blob/develop/docroot/sites/default/settings.php#L137-L139

arthurbaghdas: With the current implementation of analytics we don't have anything like that. It is configured from the UI with route_iframes module, And I don't see a benefit for complicating this to that level as you mentioned was done in the past. We have mass_utility settings ignored with config_ignore for example. I don't see a benefit for changing approach for this requirement.

weitzman: I see my suggestion as simpler, not more complicated. Getting rid of config ignore is a simplification. And I dont see how the config ignore approach will work anywhere but Prod. The value will get reset on every --db-refresh. Thats inconvenient on local dev, for example. You will have to get an URL from Prod every time you want to work on this feature.

Feel free to move this to In Progress and assign to me and I am happy to complete the ticket.

docroot/modules/custom/mass_analytics/src/Controller/MassAnalyticsController.php
```diff
@@ -0,0 +1,95 @@
...
+   *   The upcasted node object.
+   *
+   * @return array
+   *   The iframe render array or a no match message render array.
+   */
+  public function build(NodeInterface $node) {
+    $config = $this->config('mass_analytics.settings');
+    if (!empty($config->get('looker_studio_url'))) {
+      $looker_studio_url = $config->get('looker_studio_url') . '?params=%7B"nodeId":[node:nid]%7D';
+      $iframe_url = $this->token->replace($looker_studio_url, ['node' => $node]);
```
weitzman: Is there a need to use the token system here? If not we already have a $node-id() and can put that in the querystring directly, thus removing the injection of token service to this Controller.

arthurbaghdas: Yes we can remove token and use Node id directly.

PR #2056, opened 2023-05-23
Dp 22494 upgrade mayflower php to v8

clairesunstudio:
Changed:
  - description: Upgrade Mayflower PHP version to v8.
    issue: DP-22494


clairesunstudio:
I believe the backstop failures are all false positives. I think this is good to go in. @joeg8612 do you want to take a look at the backstop report to confirm?

PR #2083, opened 2023-06-12
DP-28475: Add testing so that we validate that Google Tag Manager code is actually inserted onto the pages

arthurbaghdas:
**Description:**

DP-28475

**To Test:**

---


weitzman:
What is the benefit of testing on a random node? I would test on one QAG page (any node type). That shows coverage for an entity page and the home page. This also gets testing code out of our non-testing codebase (the new alter hook). Further, it prevents us from hitting nodes whose content is in a rabbit hole (see rabbit hole module)

Otherwise this looks good.

arthurbaghdas:
> What is the benefit of testing on a random node? I would test on one QAG page (any node type). That shows coverage for an entity page and the home page. This also gets testing code out of our non-testing codebase (the new alter hook). Further, it prevents us from hitting nodes whose content is in a rabbit hole (see rabbit hole module)
> 
> Otherwise this looks good.

Ticket says: `We could test against a random published node of a random type.`

weitzman:
Thats an unfortunate suggestion in the ticket description. IMO it is the job of the developer to highlight problems in issue description.

This PR approach contradicts our testing docs at https://github.com/massgov/openmass/blob/develop/docs/testing.md#tips which say "Robust tests create and delete their own content. Relying on production content is brittle.". This PR is even more brittle since it relies on *random* production content. If the test fails it might pass on a subsequent run. And vice versa.

I suggest creating own content, or if you must use existing content, use a QAG node. I will pass the PR when it doesnt test a random node.

arthurbaghdas:
Changed

docroot/modules/custom/mass_content/tests/src/ExistingSiteJavascript/GoogleTagManagerTest.php
```diff
@@ -0,0 +1,60 @@
...
+namespace Drupal\Tests\mass_content\ExistingSiteJavascript;
+
+use weitzman\DrupalTestTraits\ExistingSiteSelenium2DriverTestBase;
+
+/**
+ * Ensures GTM is generated properly.
+ */
+class GoogleTagManagerTest extends ExistingSiteSelenium2DriverTestBase {
+
+  const QAG_PATHS = [
```
weitzman: Does the GTM insertion code actually vary by path or content type? If yes, then this looks good. If no, we dont need to slow down tests by making 20 requests. One path suffices and is faster.

arthurbaghdas: We are just making one request to one random path from the list
https://github.com/massgov/openmass/pull/2083/files#diff-16846f87bed735aee0f5335810c2ec167b74798249adbd4c95388338d663e501R53

Insertion code should be the same across the site. I'm not sure about the actual differences. So I guess 1 check for Homepage and 1 check for another page as the ticket says should be good.

weitzman:
Sorry I misread. As you say performance is not an issue. 

We have updated our testing docs to discourage randomness. See https://github.com/massgov/openmass/pull/2085. It creates a situation where a test fails and then a retry with the same code passes. I would prefer if we follow the best practice as per the testing doc. That is, to create content and then test that new node.

arthurbaghdas:
Ok done

docroot/modules/custom/mass_content/tests/src/ExistingSiteJavascript/GoogleTagManagerTest.php
```diff
@@ -0,0 +1,42 @@
...
+   * Tests GTM on Info details published page.
+   */
+  public function testGoogleTagManagerOnRandomPage() {
+    // Adding one node, to show one result.
+    $node = $this->createNode([
+      'type' => 'info_details',
+      'title' => $this->randomMachineName(),
+      'moderation_state' => MassModeration::PUBLISHED,
+    ]);
+    var_dump($node->toUrl()->toString());
```
weitzman:
```suggestion
```

weitzman (changes requested):

weitzman (approved):

PR #2084, opened 2023-06-12
Move config overrides later in settings.php

weitzman:
**Description:**
- Moved to after the include of settings.acquia.php so that global secrets file has been included on acquia.
- Followup to #2047

DP-28460

---


tom-fleming (approved):
LGTM

PR #2086, opened 2023-06-13
DP-28177: Add a new module to get analytics data from BigQuery

roberttstephens:
**Description:**

Analytics data (from Google Analytics and other sources) is currently downloaded from Superset and saved into Drupal in a table named `mass_superset_data`. It's referenced from several views, such as views that show pages with high amounts of negative feedback or pages that need attention. Moving forward, analytics data is going to be stored in BigQuery. The nightly jobs that download data from Superset should be replaced or modified so the data is pulled from BigQuery.

This adds a new module that gets data from BigQuery in the same manner as the data is currently coming from Superset. Inside of `hook_cron()`, nodes get added to a queue. When the queue is processed within certain time parameters, Drupal uses a BigQuery client to download the data for those nodes.

At a later point, the views could be modified to use this data coming from BigQuery. In addition, the mass_superset module could be disabled and eventually
deleted.

As mentioned on https://github.com/googleapis/google-cloud-php/blob/main/AUTHENTICATION.md , the recommended method for authentication is to use a JSON keyfile for a service account. The file is added to the webserver and the path is specified in an environment variable named `GOOGLE_APPLICATION_CREDENTIALS`. This key will need to be added on the non-prod and prod servers before launch. The environment variable will also need to be set. 

DP-28177

**To Test:**
```
# Add the key file and set GOOGLE_APPLICATION_CREDENTIALS=/var/www/html/docroot/file/path.json in .ddev/.env
# Set the GOOGLE_CLOUD_PROJECT environment varaible
ddev restart
# Modify the time window and last ran so `hook_cron()` runs.
ddev drush cset mass_bigquery.config end '21:59'
ddev drush state:set mass_bigquery.last_run 0
# Queue the nodes.
ddev drush eval 'mass_bigquery_cron();'
ddev drush queue:list
# Process the queue.
time ddev drush queue:run mass_bigquery_node_queue                            
ddev drush cset mass_bigquery.config end '02:00'
# Verify data was saved to the table.
ddev drush sqlq "select * from mass_bigquery_data;"
```

---


tom-fleming:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against afe81ae7f2fe87c0251f7da945260c1801d3002b
</p>

tom-fleming (approved):
LGTM!

PR #2122, opened 2023-07-11
Render cache fixes - media download links and org-wide invalidations

weitzman:
1. Stop adding media-list:document cache tag on any page that has a media download link (i.e. all pages). This is a huge problem. Almost all pages have a media download link so we are essentially blowing away our full cache whenever a document is added/updated. Was introduced in #1743 
2. Avoid org-wide invalidations. Because of our reference invalidation, whenever a node is posted, all other nodes in the same org are invalidated. This is unnecessary.

FYI, the new render debug is useful for cache metadata debugging - https://www.drupal.org/node/3162480

DP-27217

**To Test:**
- [ ] IMO there is no external review needed here.

---


docroot/modules/custom/mass_translations/mass_translations.module
```diff
@@ -23,9 +23,9 @@ use Drupal\node\NodeInterface;
  * Implements hook_preprocess_media().
  */
 function mass_translations_preprocess_media(&$variables) {
+  $cache_tags = [];
   $media = $variables['media'];
   if (($variables["view_mode"] == 'media_download_link' || $variables["view_mode"] == 'linked_desc' || $variables["view_mode"] == 'manual_desc') && $media->hasField('field_media_english_version')) {
-    $cache_tags[] = 'media-list:document';
```
arthurbaghdas, line 28: What do you offer instead?

arthurbaghdas (approved):

PR #2128, opened 2023-07-14
DP-28890: Process BigQuery queue outside of cron

roberttstephens:
**Description:**
There was an error while processing the BigQuery queue that prevented the cron from finishing. It’s best practice to process queue items separately to prevent any errors from preventing the core cron from finishing.

DP-28890

**To Test:**
Queue items, run cron, and verify that the BigQuery queue isn't processed during cron.
```
# To make sure the queue worker metadata is updated
ddev drush cr
# Modify the time window and last ran so `mass_bigquery_cron()` runs.
ddev drush cset mass_bigquery.config end '21:59'
ddev drush state:set mass_bigquery.last_run 0
# Queue the nodes.
ddev drush eval 'mass_bigquery_cron();'
ddev drush queue:list
ddev drush cron
ddev drush queue:list
```

---

### DIFF
--- a/changelogs/DP-28890.yml
+++ b/changelogs/DP-28890.yml
@@ -1,0 +1,41 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
+# #############################################################################
+# You can use one of the following types:
+#  - Added: For new features.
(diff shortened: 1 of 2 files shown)


tom-fleming:
LGTM

PR #2136, opened 2023-07-19
DP-28825: D10 upgrade - Simple Sitemap

arthurbaghdas:
**Description:**

DP-28825

**To Test:**

---


conf/drupal/config/simple_sitemap.sitemap.index.yml
```diff
@@ -0,0 +1,11 @@
+uuid: c9f06e76-3461-4032-8f6b-a1bd79d89583
+langcode: en
+status: false
```
weitzman, line 3: This is the only sitemap.index.yml config entity and it is disabled. We could/should enable it now?

arthurbaghdas: Why? Don't see a reason for that, in fact this is he config that was exported after the update hooks, so I guess everything is correct, including the UI and the types defined which is available there.

PR #2137, opened 2023-07-20
DP-28962: Get all node types at single endpoint

roberttstephens:
**Description:**
Add module to get all node types at a single JSON API endpoint.

The JSON API doesn't allow a way to get all nodes at a single endpoint, only all nodes per node type.

DP-28962

**To Test:**

---


roberttstephens:
We’re going to attempt to use the entityField_organizations field in dataLayer to map the organizations instead. In addition, it's possible to iterate over the node types and hit one endpoint per node type. Closing this PR.

PR #2164, opened 2023-08-22
Release 0.372.0

tom-fleming:
### Changed
  - DP-28733: Remove aria-labelledby from key actions, title attribute from its comp heading and correct the heading level.


weitzman:
Skip this week, vacations

PR #2168, opened 2023-08-28
DP-29391: Add flagging for migrated service details pages

arthurbaghdas:
**Description:**

DP-29391

**To Test:**
- [ ] drush mass-migrate:queue-flagging
- [ ] drush queue:run mass_migrate_queue -v

---


docroot/modules/custom/mass_migrate/src/Plugin/QueueWorker/MassMigrateQueueWorker.php
```diff
@@ -0,0 +1,115 @@
...
+        if (!$flag->isFlagged($node, $user)) {
+          // Flag an entity with a specific flag.
+          $this->flagService->flag($flag, $node, $user);
+          $flag->save();
+        }
+      }
+
+    }
+    catch (\Exception $e) {
+      $this->logger->warning("An error occurred when assigning flag for entity with data: @data. Error message: {$e->getMessage()}", ['@data' => json_encode($data)]);
```
weitzman, line 110: For the future, the logger methods take a second param with dynamic strings that can be replaced into log message. Its OK to leave this one as is.

weitzman (approved):
I dont think we need to batch the enqueue of flaggings. Did you run into trouble witth soing a straight enqueue without batching. I would think this queue insert is very fast and consumes  little memory. Its the QueueWorker that uses a bit more resources because it does an entity save. I'm fomfortable passing this because this is one time use code.

PR #2174, opened 2023-08-29
DP-29493: Fix default domain URLs from migration

roberttstephens:
**Description:**
See ticket

DP-29493

**To Test:**
You can see all instances of 'http://default' by taking a mysqldump with `--skip-extended-insert` and parsing with the following, which will return all tables (excluding cache and revision) that contain the default domain, ordered by the number of matches per table.

```
# cat or zcat depending on if the file is gzipped.
zcat recent_prod.sql.gz | grep -v 'INSERT INTO `batch`' | grep -v 'INSERT INTO `cache_entity`' | grep -v 'INSERT INTO `entity_usage`' | grep -v 'INSERT INTO `file_managed`' | grep -v 'INSERT INTO `key_value`' |  grep -v 'INSERT INTO `path_alias`' | grep -v 'INSERT INTO `redirect`' | grep -v 'INSERT INTO `simple_sitemap`' | grep -Ev 'INSERT INTO `.*revision.*`' | grep 'http://default/' | grep -oE '`.*`' | sort | uniq -c | sed 's/`//g' | sort -h -r
```

It may be helpful to take a database export, parse it to see the number of times this domain occurs, run the script with `drush scr modules/custom/mass_migrate/src/fix_default_urls.php`, and take another back up.

You can also check the links on pages. For example, on `/renew-my-license`, the `Check my Application Status` link should reference the entity. You may want to check on the edit page, since GTM is currently rewriting these default URLs.

---


tom-fleming:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against fb3e2a49519db66eadbba279ee7874b3a16c84de
</p>

weitzman:
The code looks good. I do have a question though. When you re-save a paragraph, typically a new revision is created. In order to have this new revision be used by a node, you have to update the Entity reference revisions (ERR) field thats on the node. Otherwise that ERR is still pointing at an earlier revision and the nothing is fixed. Is a new revision being created and referenced - I do not see that.

roberttstephens:
@weitzman  no, I didn't see any new revisions. My testing involved grepping a mysqldump result file and verifying that there are fewer matches for the default domain. It appears to be changing the same paragraph revision instead of a new one.

PR #2185, opened 2023-09-08
DP-29614: Modify content that needs attention to use BigQuery data

roberttstephens:
**Description:**
See ticket

DP-29614

**To Test:**
This is a change to the `Content that needs attention` block on `/admin/home`. A lot of this block was based around the snooze functionality, which is being removed. This code was refactored to make it simpler. In addition, the query was modified quite a bit. Testing should likely focus on verifying that content is showing as expected. It should show pages with 1 or more broken links OR (no's per 1000 page views > 4 AND pageviews > 20 AND the number of negative responses >=4) sorted by pageviews descending. All of this data is from BigQuery for the past 30 days.

If the data doesn't match what's on the embed on `/node/$nid/analytics-new`, then the data in the `mass_bigquery_data` table might be out of date. It can be refreshed by running `drush php-eval "\Drupal::service('bigquery.storage')->queueAll();"` and `drush queue:run  mass_bigquery_node_queue`. You can also manually modify the data in that table to verify that nodes are showing in this component as expected.

---


tom-fleming:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Changelog is not valid the following item(s) needs to be fixed (data should NOT have additional properties)</td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 63cb48818387d7305a8f6204680ce6cfc4721bcc
</p>

PR #2217, opened 2023-09-28
D10 - upgrade core and several contrib modules

weitzman:
DP-27208

**To Test:**
- [ ] Use major version spreadsheet for checklist

---


.ddev/config.personal.yaml
weitzman, line 1: This file was committed in error. It should be removed and config.personal.yaml should be added to .gitignore.

docroot/modules/custom/mass_admin_toolbar/mass_admin_toolbar.module
```diff
@@ -30,10 +30,10 @@ function mass_admin_toolbar_css_alter(&$css, $assets) {
...
   unset($css[\Drupal::service('extension.list.module')->getPath('admin_toolbar') . '/css/admin.toolbar.css']);
   // Disable styles coming from Stable because somehow they are getting into the admin theme.
-  unset($css[\Drupal::service('extension.list.theme')->getPath('stable') . '/css/toolbar/toolbar.icons.theme.css']);
-  unset($css[\Drupal::service('extension.list.theme')->getPath('stable') . '/css/toolbar/toolbar.module.css']);
-  unset($css[\Drupal::service('extension.list.theme')->getPath('stable') . '/css/toolbar/toolbar.menu.css']);
-  unset($css[\Drupal::service('extension.list.theme')->getPath('stable') . '/css/toolbar/toolbar.theme.css']);
+  // unset($css[\Drupal::service('extension.list.theme')->getPath('stable') . '/css/toolbar/toolbar.icons.theme.css']);
+  // unset($css[\Drupal::service('extension.list.theme')->getPath('stable') . '/css/toolbar/toolbar.module.css']);
+  // unset($css[\Drupal::service('extension.list.theme')->getPath('stable') . '/css/toolbar/toolbar.menu.css']);
+  // unset($css[\Drupal::service('extension.list.theme')->getPath('stable') . '/css/toolbar/toolbar.theme.css']);
```
weitzman: Lets remove this code or add a code comment about why it should stay.

docroot/modules/custom/mass_entityreference/src/Plugin/EntityReferenceSelection/MassFilterEntitiesSelection.php
```diff
@@ -31,7 +31,7 @@ protected function buildEntityQuery($match = NULL, $match_operator = 'CONTAINS')
       $types = unserialize($cookie);
     }
 
-    $query = $this->entityTypeManager->getStorage($target_type)->getQuery();
+    $query = $this->entityTypeManager->getStorage($target_type)->getQuery()->accessCheck();
```
weitzman, line 34: Usually we want accessCheck(FALSE). Is that not the case here?

.circleci/config.yml
```diff
@@ -15,7 +15,7 @@ references:
     run: {name: 'Disable SSH timeout', command: 'echo "ServerAliveInterval 30" >> ~/.ssh/config'}
 
   vendor_bin_add_path: &vendor_bin_add_path
-    run: echo 'export PATH=vendor/bin:node_modules/.bin:$PATH' >> $BASH_ENV
+    run: echo 'export PATH=vendor/bin:node_modules/.bin:$PATH' >> $BASH_ENV && source $BASH_ENV
```
tom-fleming: This change should be reverted and the deployment rerun. This was added as part of debugging deployment issues. CircleCI already sources `$BASH_ENV` before every command: https://support.circleci.com/hc/en-us/articles/360011501714-Source-a-bash-file-before-every-command-

```suggestion
    run: echo 'export PATH=vendor/bin:node_modules/.bin:$PATH' >> $BASH_ENV
```

docroot/modules/custom/mass_decision_tree/js/mass_decision_tree.js
```diff
@@ -321,7 +321,8 @@
       }
 
       // Render the templates for the root node.
-      $('div.main-content', context).once('decision-tree-init').each(function () {
+      var $elements = $(once('decisionTreeInit', 'div.main-content', context));
+      $elements.each(function (index) {
```
tom-fleming, line 325: I checked for console errors on the other spots where .once was updated. This one is throwing some console errors, but it doesn't look like functionality is affected.

docroot/modules/custom/mass_entityreference/src/Plugin/EntityReferenceSelection/MassFilterEntitiesSelection.php
tom-fleming: I was aiming for least harm to address a test error. My understanding is that prior to not having to explicitly call accessCheck, this query would have behaved like accessCheck(TRUE).
 
In this case, I think the main concern would be whether authors can view unpublished content they're supposed to see in reference fields. I tested out a reference field again, and can confirm that users who can view unpublished content can still see unpublished content with this check set to true. But, I also noticed when reviewing that we add an entity access tag [here](https://github.com/massgov/openmass/pull/2217/files#diff-0c09f380855341827536faa0a0839d59137f730de4b5d00062b706a39d2519a2R55). We probably don't need both access checks, so happy to change this if FALSE is more likely to be expected.

docroot/modules/custom/mass_admin_toolbar/mass_admin_toolbar.module
tom-fleming: Removed

docroot/modules/custom/mass_decision_tree/js/mass_decision_tree.js
tom-fleming: @arthurbaghdas to address as part of DP-30388

arthurbaghdas: There is an issue discussion about the error in console.
https://www.drupal.org/project/drupal/issues/3160052

PR #2225, opened 2023-10-10
DP-29961: D10 upgrade - Parent node dynamic property warning

arthurbaghdas:
**Description:**

DP-****

**To Test:**

---


docroot/modules/custom/mass_hierarchy/src/MassHierarchyRouteMatch.php
```diff
@@ -0,0 +1,66 @@
...
+namespace Drupal\mass_hierarchy;
+
+use Drupal\Core\Routing\RouteMatchInterface;
+
+/**
+ * Custom RouteMatch class which includes setter method.
+ */
+class MassHierarchyRouteMatch implements RouteMatchInterface {
+
+  /**
```
weitzman, line 13: Would it make sense to only have setParameter() on this class? Other methods can inherit from parent. Some methods in this class look like copies of their parent (e.g. getParameters).

arthurbaghdas: Changed the implementation

docroot/modules/custom/mass_hierarchy/src/MassHierarchyRouteMatch.php
```diff
@@ -0,0 +1,37 @@
...
+   */
+  public function __construct(RequestStack $request_stack) {
+    parent::__construct($request_stack);
+  }
+
+  /**
+   * Getter method to get the parameter we want.
+   */
+  public function getParameter($parameter_name) {
+    return $this->parameters[$parameter_name] ?? NULL;
```
arthurbaghdas, line 27: Note: getter method overridden as well to get the results we want.

PR #2228, opened 2023-10-11
Stop using permalinks in documents View

weitzman:
Upon creating or editing a node, absolute urls are converted to an entity reference and thus a path alias is used when viewing the page.

Thsi PR has some temp DDEV changes that make your local available on the edit.mass.gov domain. Once QA passes, please assign ticket to me so that the DDEV changes can be reverted.

DP-29972

**Test**
Testing of this is non-trivial since it is domain dependant. Testing must be done on local, not Acquia

- After switching to this branch, do a `ddev restart`. Then, edit your /etc/hosts so that edit.mass.gov points to your local machine instead of the internet.
- Re-save a node like https://edit.mass.gov/cleaner-greener-healthier-schools. Confirm that the first top task no longer shows a media/n link but rather an aliased link.

Follow-up: we can script a re-save of affected nodes, thus fixing their link fields. The hard part of this wil be determining the affected nodes. We need to identify which liink fields to check.

---


weitzman:
This PR now only changes the Docs View to show aliased URLs as discussed a while back. The "fix absolute URLs module" does some helpful stuff but it doesn't fix the more fundamental problem which is that folks are adding document references using the Link paragraph. The should be using the docs paragraph. Thats a whole UX discussion which is out of scope here, I think.

tom-fleming:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 371eaba2034e1c03ef4407a6360387c64b83ba10
</p>

PR #2246, opened 2023-10-26
DP-29261: Remove parent organizations from the mg_organization metadata field

arthurbaghdas:
**Description:**

DP-29261

**To Test:**

---


docroot/modules/custom/mass_metatag/src/Service/MassMetatagUtilities.php
```diff
@@ -99,6 +100,32 @@ public function getAllOrgsFromNode(Node $node, bool $parent_only = FALSE, bool $
...
+      /** @var \Drupal\node\Entity\Node[] $org_pages */
+      $org_pages = $node->field_organizations->referencedEntities();
+      foreach ($org_pages as $org_page) {
+        if ($org_page instanceof OrgPageBundle) {
+          $result[] = $this->slugify(trim($org_page->label()));
+        }
+      }
+    }
+    return array_unique($result);
+  }
```
weitzman: I hope we can do a bit of code reorganization here. The most natural way to use this code is $node->getOrg() and not need a helper. We are already setup to allow this. 

1. Copy and rename this function to \Drupal\mass_content\Entity\Bundle\node\NodeBundle. Change $node to $this.
2. Go a step father and override the method in OrgPageBundle (it lives in the same dir)

```
function getOrg() {
    return \Drupal\mass_metatag\Service\MassMetatagUtilities::slugify(trim($this->label()));
}
Note - add 'static' to function definition of `slugify`.
```

This reorg is just a little more work and results in a lot better developer experience.

weitzman (changes requested):

docroot/modules/custom/mass_metatag/src/Service/MassMetatagUtilities.php
arthurbaghdas: Thanks. I've refactored the code.

docroot/modules/custom/mass_content/src/Entity/Bundle/node/NodeBundle.php
```diff
@@ -27,4 +27,28 @@ public function getSearchNoSnippet(): FieldItemListInterface {
...
+      $org_pages = $this->field_organizations->referencedEntities();
+      foreach ($org_pages as $org_page) {
+        if ($org_page instanceof OrgPageBundle) {
+          $result[] = $utilities->slugify(trim($org_page->label()));
+        }
+      }
+    }
+    return array_unique($result);
+  }
+
```
weitzman: Thanks for moving this code. In order to match existing methods like [getSearch()](https://github.com/massgov/openmass/blob/8533a197127d71204ee8aee6a66a925f7f7b8844/docroot/modules/custom/mass_fields/src/MassSearchTrait.php#L12), getOrgs() should simply return a FieldItemListInterface(). It should be up to the caller to manipulate further after that. I propose that the slugify() call happen in MassUrlGenerator. If you have time, please search for gets of field_organizations and convert them to getOrgs() calls. 

If you prefer, I can implement this and then send to you for review.

arthurbaghdas: Sounds good.

weitzman (approved):

PR #2291, opened 2023-11-29
DP-30894: Bug editing search checkbox in bulk on org page

arthurbaghdas:
**Description:**

DP-30894

**To Test:**

---


patches/DP-30894_views_bulk_edit_remove_paragraph_fields_from_bulk_form.patch
```diff
@@ -0,0 +1,15 @@
...
++++ b/src/Form/BulkEditFormTrait.php
+@@ -255,6 +255,10 @@ trait BulkEditFormTrait {
+     $definitions = $this->entityFieldManager->getFieldDefinitions($entity_type_id, $bundle);
+ 
+     foreach (Element::children($form) as $key) {
++      if (isset($form[$key]['widget'][0]['#paragraph_type'])) {
++        unset($form[$key]);
++        continue;
++      }
+       if (isset($form[$key]['#access']) && !$form[$key]['#access']) {
```
weitzman, line 13: This doesn't look typical to me. usually #access is evaluated by render API and items are not shows there. It seems like we are doing the renderer's job here.

weitzman (commented):
_Update: I see that we tried this and it was moving too slowly._

Would it make sense to open an issue and perhaps contibute a patch to views_bulk_edit module?

patches/DP-30894_views_bulk_edit_remove_paragraph_fields_from_bulk_form.patch
arthurbaghdas: This is a temporary solution to resolve our issue. Refer to the issue on Drupal.org at https://www.drupal.org/project/views_bulk_edit/issues/3024419. We will switch to the contributed version once the community has fixed the problem.

weitzman: Good to know that this is temporary code, but it still may last in our codebase for years. It doesn't pass [peer review guidelines](https://github.com/massgov/openmass/blob/develop/docs/peer_review_checklist.md) IMO, temporary or not. If anyone else thinks differently, then they can record an internal review approval. I suggest  adding code comments indicating why we are bypassing the render system and when the bypass can be removed.

PR #2309, opened 2023-12-12
DP-30894: Populate contextual search field values

arthurbaghdas:
**Description:**

DP-30894

**To Test:**
- [ ] Run `drush mass-content:populate-search-fields` or alias version of the command `drush mpsf`

---


docroot/modules/custom/mass_fields/mass_fields.deploy.php
```diff
@@ -53,3 +53,91 @@ function mass_fields_deploy_search_fields() {
...
+  $show_parent_nids = [705346, 408791, 790226, 342871, 514371, 415871, 685541, 468571, 22826, 51951, 418061, 64866, 571991, 646046, 790506, 14686, 683761, 13626, 47751, 528126, 15091, 74746, 629876, 609336, 71546, 144916, 71416, 74466, 6801, 511121, 13726, 409611, 507841, 74116, 13891, 205351, 27511, 488491];
+  $nids_to_load = array_unique(array_merge($do_not_include_search_nids, $show_parent_nids));
+
+  $query = \Drupal::entityQuery('node')->accessCheck(FALSE);
+  $query->condition('type', 'org_page');
+  $query->condition('nid', $nids_to_load, 'IN');
+
+  if (empty($sandbox)) {
+    // Get a list of all nodes of type org_page.
```
weitzman:
```suggestion
    // Get a list of specified nodes.
```

docroot/modules/custom/mass_fields/mass_fields.deploy.php
```diff
@@ -53,3 +53,91 @@ function mass_fields_deploy_search_fields() {
...
+        $latest_revision = $storage->loadRevision($rid);
+        if (isset($latest_revision)) {
+          try {
+            if (in_array($latest_revision->id(), $do_not_include_search_nids)) {
+              $latest_revision->set('field_org_no_search_filter', 1);
+            }
+            if (in_array($latest_revision->id(), $show_parent_nids)) {
+              $latest_revision->set('field_include_parent_org_search', 1);
+            }
+            $latest_revision->setSyncing(TRUE);
```
weitzman: IMO we should set a revision log message that says 'Deploy hook setting field_org_no_search_filter and  field_include_parent_org_search to 1'

weitzman (changes requested):

docroot/modules/custom/mass_fields/mass_fields.deploy.php
arthurbaghdas: Added

weitzman (approved):

weitzman (approved):
Approved. Can merge when tests are passing.

docroot/modules/custom/mass_content/src/Commands/MassContentCommands.php
```diff
@@ -366,6 +366,84 @@ public function healRawLinkInContent($options = ['format' => 'table']): RowsOfFi
...
+   *
+   * @aliases mpsf
+   */
+  public function populateContextualSearchFields() {
+    // 1. Log the start of the script.
+    $this->logger()->info('Update nodes batch operations start');
+
+    // 2. Retrieve the nodes.
+    $storage = $this->entityTypeManager->getStorage('node');
+    try {
```
weitzman: Why do we have a try here? There is just array operations below. Maybe this is refactoring cruft.

docroot/modules/custom/mass_content/src/MassContentBatchManager.php
```diff
@@ -142,6 +142,70 @@ public static function processServiceNode($id, ContentEntityBase $node, array $s
...
+    catch (\Exception $e) {
+      \Drupal::state()->set('entity_hierarchy_disable_writes', FALSE);
+    }
+    if (!$node->isLatestRevision()) {
+      $storage = \Drupal::entityTypeManager()->getStorage('node');
+      $query = $storage->getQuery()->accessCheck(FALSE);
+      $query->condition('nid', $node->id());
+      $query->latestRevision();
+      $rids = $query->execute();
+      foreach ($rids as $rid) {
```
weitzman: What revision ids are we iterating over here. All the non-published ones? That seems unnecessary to edit all of those. If we are iterating over just one revision, lets be more clear and not use a loop.

docroot/modules/custom/mass_content/src/MassContentBatchManager.php
```diff
@@ -142,6 +142,70 @@ public static function processServiceNode($id, ContentEntityBase $node, array $s
...
+
+    // Store some results for post-processing in the 'finished' callback.
+    // The contents of 'results' will be available as $results in the
+    // 'finished' function (in this example, batch_example_finished()).
+    $context['results'][] = $id;
+
+    // Optional message displayed under the progressbar.
+    $context['message'] = t('Running Batch "@id" @details',
+      ['@id' => $id, '@details' => $operation_details]
+    );
```
weitzman, line 207: I am not seeing where $id is defined.

docroot/modules/custom/mass_content/src/Commands/MassContentCommands.php
arthurbaghdas: Yes, removed.

docroot/modules/custom/mass_content/src/MassContentBatchManager.php
arthurbaghdas: the $id variable was passed form `processContextualSearchFields` method as the first argument of the method it is set to be batch id see this line https://github.com/massgov/openmass/blob/10cafdfecc705b6685915a92fc7a99b355010bed/docroot/modules/custom/mass_content/src/Commands/MassContentCommands.php#L408

docroot/modules/custom/mass_content/src/MassContentBatchManager.php
arthurbaghdas: We are always getting the latest non-publish version, this is the standard code we use in all places, I've refactored it a bit to exclude iteration, since we will always get only 1 result from the query execution

PR #2317, opened 2023-12-19
Release 0.385.0

joeg8612:
### Changed
  - DP-28433: Modify buttons on location page to use 3 across format.
  - DP-30296: Convert content page banner to use new Mayflower components.
  - DP-31191: On the external link for collections content type, data type is no longer required.
  - DP-31225: Max header length on info details section increased to 200.
  
### Added
  - DP-31098: Change help text for org fields.


weitzman:
Abandoned. Did not deploy successfully.

PR #2336, opened 2024-01-09
Respect directions link override when rendering a contact info node

weitzman:
Introduces a bundle class for address paragraphs as a way to centralize its code.

DP-31372


tom-fleming:
LGTM

PR #2343, opened 2024-01-16
DP-24989: Change sort of autocomplete to put organizations, topics, services first

arthurbaghdas:
**Description:**

DP-****

**To Test:**

---


weitzman:
I personally would have written a custom autocomplete handler rather than add a View and alteration of the view but this is a matter of taste so Approved and merged.

PR #2266, opened 2023-11-08
Dp 30296 consolidate page header component

clairesunstudio:
Link to Mayflower PR https://github.com/massgov/mayflower/pull/1845

PR #2380, opened 2024-02-01
Fix local dev for test running for selenium-chrome

weitzman:
DP-31671

---

### DIFF
--- a/.ddev/config.arm.yaml
+++ b/.ddev/config.arm.yaml
@@ -1,1 +1,0 @@
-dbimage: massgov/mysql-sanitized:latest.arm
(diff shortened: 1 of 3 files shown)


.ddev/config.arm.yaml
```diff
@@ -1 +0,0 @@
-dbimage: massgov/mysql-sanitized:latest.arm
```
arthurbaghdas, line 1: Was this an intended change?

weitzman: Yes - it is unused. We use an env variable instead.

arthurbaghdas:
@weitzman here is the change. this helped me to solve my issue https://github.com/massgov/openmass/pull/2266/files/7b1cb37d207bc79aa3aa82df7497e6429d0dc6b5#diff-97a20696a8f6a925104f5174167c8c6f73dd7606b1f7f263dea1f1d9e23969cbR128

PR #2401, opened 2024-02-16
DP-24572 - Set html lang tag to match content

tom-fleming:
**Description:**
This PR sets the html lang tag to match the language of the content, and sets lang tags on UI elements that are explicitly English. Specifically, this PR and the Mayflower branch associated with it, adds lang="en" and dir="ltr" attributes to the header, footer, brand banner, pre-content, and post-content sections on patternlab templates. This is currently implemented in a non-optional way for the purposes of offering a proof of concept of allowing the html lang tag match the language of the primary content when UI elements surrounding it are not translated.

DP-24572

**To Test:**
- [ ] See info-details/welcome-to-the-boston-re-employment-center-rec and change languages

---


clairesunstudio:
Several things I noticed on [this page](https://pr2401-epncvt9bu6vz4f3ikwilvteqmcfviezo.tugboatqa.com/info-details/huanyingguanglinboshidunzaijiuyezhongxin-rec-welcome-to-the-boston-re-employment-center-rec-chinese#)
1. Some of the `lang` value set on the HTML doesn't match what Google uses. For example simplified Chinese, "zh-hans" (we set in HTML tag) and "zh-CN" (Google uses)
2. Even if the `lang` value matches, Google still translate the content. I can see the content changes in multiple different languages, but not sure if it's because we have different values set in the content area, for example "FR" in the richtext and "fr" in the HTML tag
3. When you close out Google, Google resets the HTML lang attribute to "en". This will be confusing for AT users as the content is in a different language. 
4. The language bar in the pre content probably shouldn't have the "en" value as they are in different languages. We have a translate labels toggle, maybe set the value conditionally?
5. We now set the templating components to "en", but when user uses the google translate, those components also get translated. I'm not sure what that means to the users especially AT users, as they are still set as "en" 

If we end up going with the option, we can remove language attributes added by CKeditor. Not sure if it's the CKeditor feature that currently adds all the spans and breaks up the paragraphs, but it definitely handles lang attribute setting inconsistently:

Lots of nested tags, values set to "FR" not sure recognized differently from "fr" 
Screen Shot 2024-03-13 at 10 25 16 AM

Lots of nested tags, sentence gets broken up and punctuations are tagged in another language
Screen Shot 2024-03-13 at 10 25 29 AM

Lots of nested tags, language attribute not set 
Screen Shot 2024-03-13 at 10 28 59 AM

tom-fleming:
Favoring #2583

PR #2421, opened 2024-03-02
DP-31982: org type field and taxonomy

joeg8612:
**Description:**
Adding org type field and org type taxonomy that will be used with it. The taxonomy will be populated later. The field will be visible to authors but optional for now but will be configured later after the taxonomy is populated in production.

DP-31982

**To Test:**
- [ ] Verify that the org type field is visible for org content type. Should be no choices to pick since taxonomy will be empty.

---


tom-fleming:
LGTM

PR #2437, opened 2024-03-12
Update entity_usage module and remove unneeded patches

weitzman:
DP-29889

---


tom-fleming:
Still in progress, but this is working without patches currently. Need to double check that changes from both patches are in place and fix any validation errors that are likely to show.

tom-fleming:
@weitzman The patches that were in place heavily modified how results are returned, and also required changes in a custom module. This was far too complex to untangle, and the amount of effort required to untangle the changes we needed from the heavily refactored changes in entity_usage was not worth it. Additionally, since this wasn't the first time we've had a blocker updating entity_usage due to patches, I don't think it's worth putting ourselves in a position of needing to deal with a tangle of patches to keep entity_usage up to date while also keeping a separate merge request up to date that now needs to deal with a refactor on top of waiting for other unresolved issues to be resolved.

I decided that the changes openmass requires for its local task listing are divergent enough to warrant a clean break from entity_usage. My goal with these changes was to make future updates to entity_usage more maintainable by eliminating overly complex patches while also ensuring that we do not override anything beyond UI and queries used to display results. The changes in this PR do the following:

- Creates a ListUsageController in mass_entity_usage that is not dependent on entity_usage
- The new, custom ListUsageController mirrors the patched beta7 version of ListUsageController, includes the changes from the custom mass_entity_usage module's MassLocalTaskUsageController, and eliminates any unused functions
- Creates MassEntityUsage, which extends EntityUsage and implements an extended version of EntityUsageInterface to include the subquery changes that were previously patched into EntityUsage. These changes do not override anything in EntityUsage, but are required here to access the same connection from EntityUsage.
- Eliminates the need for the two remaining patches

docroot/modules/custom/mass_entity_usage/src/MassEntityUsage.php
```diff
@@ -0,0 +1,121 @@
...
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\entity_usage\EntityUsage;
+
+/**
+ * Extends the entity usage base class.
+ */
+class MassEntityUsage extends EntityUsage implements MassEntityUsageInterface {
+
+  /**
+   * {@inheritdoc}
```
weitzman, line 14: Instead of inheritdoc, how about we say why we are overriding the parent method.

weitzman (commented):
Looks great to me! Do you think we need any more test coverage? That can be a follow-up.

docroot/modules/custom/mass_entity_usage/src/MassEntityUsage.php
tom-fleming: Ended up not needing this method

tom-fleming:
> Looks great to me! Do you think we need any more test coverage? That can be a follow-up.

I think the tests that we currently have in mass_entity_usage are still adequate with these changes. Also verified that we are not skipping them and that we are getting actual successful results.

arthurbaghdas (approved):

PR #2504, opened 2024-05-07
DP-32285: BUG - overview appears below TOC when only one section with rich text including H2s

arthurbaghdas:
**Description:**

DP-****

**To Test:**

---


docroot/themes/custom/mass_theme/templates/content/node--info-details.html.twig
```diff
@@ -122,7 +111,6 @@
         {% endblock %}
       {% endembed %}
     {% endif %}
-  {% endif %}
```
weitzman, line 125: Since an outer if() statement if removed, I would have expected all these lines to have 2 spaces less of indent. But the lines did not move left. Maybe the diff is misleading.

arthurbaghdas: it's correct. I don't see any spacing issues.

weitzman: I pushed a commit which removed 2 spaces of indentation.

arthurbaghdas: Strange, it was not showing in my IDE.

PR #2510, opened 2024-05-09
Move render cache to the database

weitzman:
Helps with debugging a Prod issue - https://massgov.atlassian.net/browse/DP-33081

We had render cache in the DB in Prod briefly but it was sligtly slower so we reverted. Note that Acquia recommends DB for the render bin https://github.com/acquia/memcache-settings/blob/main/memcache.settings.php


tom-fleming (approved):
LGTM

PR #2515, opened 2024-05-14
Hotfix/0.397.4

tom-fleming:
**Description:**

DP-****

**To Test:**

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## [0.397.4] - May 14, 2024
+
+### Fixed
+- Fixed potential render cache issues where cache is applied to existing render arrays.
+
 ## [0.397.3] - May 13, 2024
 
 ### Changed
--- a/composer.json
+++ b/composer.json
@@ -408,6 +408,9 @@
             "drupal/diff": {
                 "500 error: Possible infinite loop": "https://www.drupal.org/files/issues/2019-02-15/diff-500-errors-possible-infinite-loop-3033455-2-d8.patch"
             },
+            "drupal/dropzonejs": {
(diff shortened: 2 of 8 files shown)


arthurbaghdas:
Hi @tom-fleming 
I've debugged this a little.
So with the the google_tag module version 2.0.2 this is what we've got as a cachable metadata, as you can see in the screenshot below the cache contexts array is empty
Screenshot 2024-05-16 at 11 36 47
but when we update to the latest 2.0.4 we get this change in the contrib module
`$cacheable_metadata->addCacheableDependency($config);` on the line docroot/modules/contrib/google_tag/google_tag.module:72
Which adds the cache contexts, see the image below
Screenshot 2024-05-16 at 11 39 50
Now we have url.path cache context for the GTM.

Based on this change when we test redirects, the URL of the node changes, since we are changing node state to unpublished (appending the ---unpublished at the end of the URL) and manipulating with the redirects, so i think in this case this is false positive and we can add the lines below which will make the test to pass.

`diff --git a/docroot/modules/custom/mass_redirects/tests/src/ExistingSite/RedirectsTest.php b/docroot/modules/custom/mass_redirects/tests/src/ExistingSite/RedirectsTest.php
index f7936a7e4..c9752fd9f 100644
--- a/docroot/modules/custom/mass_redirects/tests/src/ExistingSite/RedirectsTest.php
+++ b/docroot/modules/custom/mass_redirects/tests/src/ExistingSite/RedirectsTest.php
@@ -22,6 +22,13 @@ class RedirectsTest extends MassExistingSiteBase {
   private $orgNodeTarget;
   private $sourcePaths;
 
+  /**
+   * Un-cacheable dynamic page patterns.
+   */
+  protected static array $uncacheableDynamicPagePatterns = [
+    'orgs/*',
+  ];
+
   /**
    * Create an editor and a node with url redirects.
    */`

PR #2532, opened 2024-05-27
DP-33081: Getting redirected to a different page on mass.gov when trying to edit a page

arthurbaghdas:
**Description:**

DP-33081

**To Test:**

---


docroot/themes/custom/mass_theme/templates/misc/status-messages.html.twig
```diff
@@ -21,8 +21,9 @@
  *   - class: HTML classes.
  */
 #}
-
+{{ attach_library('seven/classy.messages') }}
```
weitzman, line 24: Should mass_theme now be dependant somehow on seven theme? Does Olivero do that?

weitzman (commented):
We can remove  docroot/modules/custom/mass_admin_toolbar/src/Plugin/Block/MessageTabs.php

docroot/themes/custom/mass_theme/templates/misc/status-messages.html.twig
arthurbaghdas: We had that dependency when we switched to drupal 10, that's why we brought seven theme as a contrib to our D10

weitzman: OK, I guess we can keep it as an undeclared dependency.

PR #2548, opened 2024-06-06
DP-32968: Search and replace /media and /files URLs with /doc URLs for entity reference/link fields

arthurbaghdas:
**Description:**

DP-32968

**To Test:**

---


docroot/modules/custom/mass_content/src/Drush/Commands/MassContentCommands.php
```diff
@@ -425,6 +428,27 @@ public function searchAndReplaceMedia() {
...
               }
             }
+            elseif ($fieldType === 'link') {
+              foreach ($field as $key => $item) {
+                if ($item->uri) {
+                  if (strpos($item->uri, 'internal:') === 0) {
+                    $url = substr($item->uri, strlen('internal:'));
+                    $processed = $urlReplacementService->processLink($url);
+                    if ($processed['changed']) {
+                      $item->value = 'internal:' . $processed['link'];
```
tom-fleming, line 438: This is set to `$item->value` instead of `$item->uri` so that it works with `AutowireTrait`.

PR #2556, opened 2024-06-17
Removed add and remove tags buttons, converted Tags values to list

tom-fleming:
**Description:**
Removes tag modal from feedback manager.

DP-33637

**To Test:**
- [ ] Go to admin/ma-dash/feedback
- [ ] Check feedback filtering, including by tag
- [ ] Confirm that "Add tag" button does not show in table
- [ ] Confirm that tags appear as a comma-separated list of tags

---


docroot/modules/custom/mass_feedback_loop/src/Form/MassFeedbackLoopPerNodeForm.php
```diff
@@ -32,7 +32,6 @@ class MassFeedbackLoopPerNodeForm extends FormBase {
    */
   public function __construct(MassFeedbackLoopContentFetcher $content_fetcher) {
     $this->contentFetcher = $content_fetcher;
-    $this->request = $this->getRequest();
```
tom-fleming, line 35: Not related to the tag removal, but this was not being used and it was causing deprecation messages on the per-node feedback page.

PR #2568, opened 2024-06-26
DP-33767: Media contacts missing on press releases

arthurbaghdas:
**Description:**

DP-33767

**To Test:**

---

### DIFF
--- a/docroot/modules/custom/mayflower/src/Prepare/Organisms.php
+++ b/docroot/modules/custom/mayflower/src/Prepare/Organisms.php
@@ -9,6 +9,7 @@ use Drupal\Core\Entity\ContentEntityInterface;
 use Drupal\mass_content_moderation\MassModeration;
 use Drupal\mayflower\Helper;
 use Drupal\node\Entity\Node;
+use Drupal\node\NodeInterface;
 
 /**
  * Provides variable structure for mayflower organisms using prepare functions.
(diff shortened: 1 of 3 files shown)


docroot/modules/custom/mayflower/src/Prepare/Organisms.php
```diff
@@ -198,13 +199,19 @@ public static function prepareContactList($entity, array $options = [], array &$
...
+          if ($contact->moderation_state[0]) {
+            if ($contact->moderation_state[0]->value === MassModeration::PUBLISHED) {
+              $contacts[] = Molecules::prepareContactUs($contact, $options['groups']);
+            }
           }
         }
       }
+      else {
+        $contacts[] = Molecules::prepareContactUs($contact, $options['groups']);
+      }
```
tom-fleming, line 213:
```suggestion
      $contacts[] = Molecules::prepareContactUs($contact, $options['groups']);
```
@arthurbaghdas It looks like we should be able to remove the conditionals around this.

arthurbaghdas: Screenshot 2024-06-27 at 09 35 58

No that will actually cause a duplicate content. The else statement is for the cases when the referenced entity is a paragraph and doesn't have a moderation state.

PR #2577, opened 2024-07-04
DP-33880: 500 Error when comparing revisions

arthurbaghdas:
**Description:**

DP-33880

**To Test:**

---

### DIFF
--- a/docroot/modules/custom/mass_media/mass_media.module
+++ b/docroot/modules/custom/mass_media/mass_media.module
@@ -29,6 +29,8 @@ use Drupal\media\MediaInterface;
  * @throws \Drupal\Core\Entity\EntityStorageException
  */
 function mass_media_media_update(EntityInterface $entity) {
+
+  // getUploadFile method and field_upload_file field are only available to documents.
   if (!($entity instanceof DocumentBundle)) {
     return;
   }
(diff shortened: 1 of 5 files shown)


docroot/modules/custom/mass_media/mass_media.module
```diff
@@ -28,6 +29,10 @@ use Drupal\media\MediaInterface;
  * @throws \Drupal\Core\Entity\EntityStorageException
  */
 function mass_media_media_update(EntityInterface $entity) {
+  if (!($entity instanceof DocumentBundle)) {
+    return;
+  }
+
```
weitzman, line 37: What is the purpose of this? Maybe add a code comment. It doesnt look like the subsequent code is specific to document bundle.

arthurbaghdas: It is specific to document bundle only.
// getUploadFile method and field_upload_file field are only available to documents.
So the condition added to prevent logs thrown for other bundles.

weitzman: That implies that video media remain available at public urls even after they are unpublished. Is this a problem, @joeg8612?

weitzman: Regardless of that answer, the issue is fixed so merging.

PR #2583, opened 2024-07-12
DP-24572 - Enabled locale

tom-fleming:
**Description:**
Enabled locale for demonstration purposes.

DP-24572

**To Test:**

---


tom-fleming:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against edad1cfaa8a19959e79365899a1ab643d7f95044
</p>

conf/drupal/config/views.view.user_admin_people.yml
dstorozhuk, line 1: @tom-fleming  this file is unrelated to the task, right?

dstorozhuk:
@tom-fleming as we discussed yesterday, with @clairesunstudio — some elements on the page will not be translatable. 
For example, "OFFERED BY" referenced node, Breadcrumps. We can't translate them since if we do so, then we have to translate the entire reference node/ page to not confuse users who clicked. Let’s say a user clicked on some of the Spanish titles on the Spanish page and landed on the English page. 
In that case, we have to make sure untranslatable elements on the page are wrapped in elements with `lang="en"` attribute so screen readers will correctly recognize them. Probably, need to use the `dir="ltr"`, but I am not sure how it works in combination with RTL languages.

dstorozhuk:
One consideration is the deployment of the new translation system and interface translations simultaneously. Allowing editors to translate interfaces on live environments may not be ideal as changes will be reflected immediately, leading to partially translated pages. To prevent this, it is better to avoid live translations. 
Additionally, if new elements are added in the future, editors must be notified to provide or update translations for each language. 
To address this, we can establish a table that can be exported as a CSV file. This table would contain all languages and strings for translation, extracted from modules/themes as discussed in the meeting. 
Access to this table would be given to language editors, allowing them to update translations as needed. 
Before each release, we would export existing strings and their translations to this table using CSV. Editors would be requested to ensure all translations are up to date. 
Afterward, the table would be exported back to CSV format. 
During the deployment process, we would upload translations to Drupal using a method such as CSV to .po or a custom mechanism for importing translations from CSV.

dstorozhuk:
Also, if talking about untranslatable elements on the page (Main Menu, Help Us improve this page, Footer, Header links, etc.) - should the main menu stay untranslated? If not - we should also make it translatable, same as nodes, so it can be translated. 
I think it is worth translating the menu, even if the destination page stays untranslated.

dstorozhuk:
In my opinion, opting for the language negotiator plugin and locale modules is the best approach for this task. It's a wise decision that aligns with my own thinking.

conf/drupal/config/views.view.user_admin_people.yml
tom-fleming: Yes, it looks like the configuration export switched 0 values to false. It wouldn't need to be included in the PR. Will leave this comment open for me to fix.

PR #2627, opened 2024-08-13
DP-34207: Alter the ImageStyle class to substitute original URI if image is animated GIF.

dstorozhuk:
**Description:**
Altered the ImageStyle class so we can substitute the GIF image URI in certain conditions.

DP-34207

**To Test:**
- Open any page with giant GIF animated, for example /masstalent and observe that page is loading, image is visible (probably need to wait some time for the image to load).
Despite path is still point to image style - the original file loaded should be instead (so the size image must be equal to the original).

---

### DIFF
--- a/changelogs/DP-34207.yml
+++ b/changelogs/DP-34207.yml
@@ -1,0 +1,41 @@
+#
+# Write your changelog entry here. Every pull request must have a changelog yml file.
+#
+# Change types:
+# #############################################################################
+# You can use one of the following types:
+#  - Added: For new features.
+#  - Changed: For changes to existing functionality.
+#  - Deprecated: For soon-to-be removed features.
+#  - Removed: For removed features.
+#  - Fixed: For any bug fixes.
+#  - Security: In case of vulnerabilities.
+#
+# Format
+# #############################################################################
+# The format is crucial. Please follow the examples below. For reference, the requirements are:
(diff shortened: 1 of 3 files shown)


tom-fleming:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against eee9d55863334148471c44194e6d55c63690d0ba
</p>

weitzman:
Is there any benefit to using https://www.drupal.org/project/animated_gif instead of custom code? If not thats OK. The code here looks good.

dstorozhuk:
> Is there any benefit to using https://www.drupal.org/project/animated_gif instead of custom code? If not thats OK. The code here looks good.

@weitzman 
Thanks for pointing to that module. 
Yes, it is almost what we need except one cause - the referenced modules does not cover situation when direct call method `ImageStyle::buildUri()`, 'ImageStyle::buildUrl()`. 
And we can use that method directly across the site, for example on home page. 
Mass gov 2024-08-28 06-55-39

So our solution will cover more if not all cases.

PR #2638, opened 2024-08-19
DP-34517: Investigate /files URLs on Mass.gov

arthurbaghdas:
**Description:**

DP-****

**To Test:**

---


docroot/modules/custom/mass_content/mass_content.install
```diff
@@ -92,16 +92,6 @@ function mass_content_update_10001() {
   }
 }
 
-/**
```
dstorozhuk, line 95: @arthurbaghdas as far as I can see, `mass_content_update_10003()` was released to production already, in some of the previous versions. 
I don't think we remove the previously executed hook_update_N, since it can lead to an unclear upgrade path. 
What was the reason removing the hook?

docroot/modules/custom/mass_content/mass_content.install
```diff
@@ -119,3 +109,13 @@ function mass_content_update_10004() {
     \Drupal::keyValue("system.schema")->delete($module_name);
   }
 }
+
+/**
+ * Reset state variables used by search and replace media URLs drush command.
+ */
+function mass_content_update_10005() {
```
dstorozhuk, line 133: @arthurbaghdas  if you just need execute the same code second time and don't want have code duplication - please create `_modules_name_[helper]()` method in .install and call that code in `mass_content_update_10005` and `mass_content_update_10003`.

docroot/modules/custom/mass_fields/src/MassUrlReplacementService.php
dstorozhuk, line 1: Optional recommendation:

Code in this file has a lot of duplicates. 
It is hard to read the logic of the class. 
I would suggest at the least moving:
```
             // Replace the href with a media URL.
               // Note: We always want to concat download string to the URL.
               $mediaUrl = $mediaEntity->toUrl()->toString() . '/download';
               if (isset($url['query'])) {
                 $mediaUrl .= '?' . $url['query'];
               }
```
to a separate method. 
Ideally, the entire file should be refactored, and logic should be separated to a method, so that the file becomes maintainable. 
@arthurbaghdas try to avoid more than two nested statements, in a single method, so that code can be maintainable and understandable by other developers who might need to read and develop it later.

docroot/modules/custom/mass_fields/src/MassUrlReplacementService.php
```diff
@@ -265,17 +289,42 @@ public function processLink($link) {
...
           $mediaUrl = $mediaEntity->toUrl()->toString() . '/download';
+          if (isset($url['query'])) {
+            $mediaUrl .= '?' . $url['query'];
+          }
           $link = $mediaUrl;
           $changed = TRUE;
         }
+        else {
+          // We cover the case when the file was used in revisions.
+          $mediaQuery = \Drupal::entityQuery('media')
```
dstorozhuk: @arthurbaghdas  why don't use DI here? 

You already have `$this->entityTypeManager` injected, so use it. 

```
$entity_storage = $this->entityTypeManager->getStorage('media');
$entity_ids = $entity_storage->getQuery()
```

dstorozhuk (changes requested):

tom-fleming:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against b294583b80ea566d9e75197cf9a2f07d1b10400b
</p>

docroot/modules/custom/mass_content/mass_content.install
arthurbaghdas: I think you don't know that sometimes we do remove entire install files content as part of our code cleanup process. Don't understand what upgrade path do you refer to in this specific condition.

docroot/modules/custom/mass_content/mass_content.install
arthurbaghdas: I've done it the way you want. But FYI there is no need for a helper function and over complicating the code, see my previous comment. Sometimes as part of a cleanup work we do remove the old code form .install, deploy and post_update files.

docroot/modules/custom/mass_fields/src/MassUrlReplacementService.php
arthurbaghdas: Create a separate ticket if you want to handle it. I don't want to block merging urgent ticket for this reason.

docroot/modules/custom/mass_fields/src/MassUrlReplacementService.php
arthurbaghdas: Good catch, fixed, thank you

docroot/modules/custom/mass_fields/src/MassUrlReplacementService.php
dstorozhuk: I proposed that as optional change. If it is urgent - definitely not for now.

docroot/modules/custom/mass_content/mass_content.install
dstorozhuk: Ok, that is new to me. It is contradicts with recommendation to not remove the `hook_update_N` if it was already executed on the destination database, to not mislead the other developers. 
But if it a common process on massgov then we should follow that rule.

docroot/modules/custom/mass_content/mass_content.install
dstorozhuk: Thanks.

dstorozhuk (approved):

PR #2641, opened 2024-08-20
DP-34424: Fixed the issue with styles in rich text related to images.

dstorozhuk:
**Description:**
I added the CSS overrides for H4 in reach text paragraphs and some margin left for UL element

DP-34424

Related PR:
https://github.com/massgov/mayflower/pull/1930

**To Test:**

---


docroot/themes/custom/mass_theme/overrides/css/basic-html.css
```diff
@@ -115,3 +115,12 @@ form.fsForm.rtl {
 .page-content--intro .ma__video__skip-link {
   background-color: #fff;
 }
+
+.ma__rich-text h4 {
```
clairesunstudio: Can you move this into Mayflower? Also we should clear float for all heading levels, e.g. H2, H3, etc.
This is an example taken from Mayflower:
Screenshot 2024-08-22 at 5 32 24 PM


In mayflower packages/assets/scss/03-organisms/_rich-text.scss
```
// clear float for all headings after left of right floating figures 
.ma__rich-text {
  h1, h2, h3, h4, h5, h6 {
     clear: both;
}
}
```

docroot/themes/custom/mass_theme/overrides/css/basic-html.css
```diff
@@ -115,3 +115,12 @@ form.fsForm.rtl {
...
+  clear: both;
+}
+
+.ma__rich-text div.align-left.half-width.embedded-entity + ul {
+  margin-left: 83px;
```
clairesunstudio: This is too specific, the image could be at any size, float right or left, also we need to consider different screen sizes
For instance, after applying this margin, at a smaller screen size, the image takes up the full row however the bulleted list still has a margin-left of 83px
Screenshot 2024-08-22 at 11 28 00 AM

TBH, I think there are two acceptable fixes (please check with Joe for business requirement):
1. Clear float for a bulleted list (same as the headings). I don't think its common to have a bulleted list side by side with an image like this - The example in openmass is an edge case, also a misuse of a bulleted list. We should tell them not to use bulleted list this way. The simplest is to not have bulleted list wrap around an image like this. 
2. If we anticipate there's a need to have a bulleted list side by side with a floating image, we could add `list-style-position: inside` instead of adding a margin-left at a specific value. This moves the bullet inside the content. This is not a perfect fix but will avoid the overlap. I think this fix should also exist in Mayflower and take into consideration the media queries. 

Mayflower test: http://localhost:3000/?p=organisms-rich-text
Before:
Screenshot 2024-08-22 at 5 57 58 PM
After:
Screenshot 2024-08-22 at 5 58 03 PM

docroot/themes/custom/mass_theme/overrides/css/basic-html.css
dstorozhuk: Added to the PR https://github.com/massgov/mayflower/pull/1930 .

docroot/themes/custom/mass_theme/overrides/css/basic-html.css
dstorozhuk: @clairesunstudio thanks for the review. This is an edge case from my point of view, that is why I was so specific in my CSS selector. 
Agree that we need to test it on different screen sizes and probably try `style-position: inside`. (I like that solution more). 
If the image floats right - there is no issue with bullet points, because bullet points are on the left:
Mass gov dev — (L) SSC Service Catalog Offerings TEST | Mass gov 2024-08-23 09-25-56
Also, it is not an issue if the image is full width:
Mass gov dev — (L) SSC Service Catalog Offerings TEST | Mass gov 2024-08-23 09-28-43

So I would stay with the selector `a__rich-text div.align-left.half-width.embedded-entity + ul` and try `style-position: inside`.

@joeg8612, what do you think this is an edge case and not a widely used pattern, or we want to make this available for all Mayflower dependent sites?

weitzman (approved):

arthurbaghdas (approved):

PR #2659, opened 2024-09-03
Bulk update of help text and minor changes

arthurbaghdas:
**Description:**

DP-****

**To Test:**

---


docroot/modules/custom/mass_validation/mass_validation.module
```diff
@@ -1773,6 +1773,25 @@ function _mass_validation_form_node_info_details_form_validate(array &$form, For
       // If at least 1 section is populated, validation passed.
       $section_populated = isset($section['subform']);
     }
+    // External links should have a title.
+    if (isset($section['subform']['field_section_long_form_addition'])) {
```
dstorozhuk, line 1777: @arthurbaghdas it is optional, but my recommendation - don't do more that 2-3 nested statements in a single method. This type of code is very hard to read, understand and maintain. And very easy to break, especially if it not covered by the tests. I recommend in future splitting the code to a classes, methods, use Drupal 8 feature called ClassResolver which aimed to organize and encapsulate the code with Classes, and do the dependency injection if need, without need to declare a service.

If you don't think it is reasonable to write a test and do a refactor right now - it is ok, I would just recommend create a ticket about test coverage for the validation added.

PR #2666, opened 2024-09-09
DP-34850: Some test are missing by phpunit-finder because of incorrect test file location.

arthurbaghdas:
**Description:**

DP-34850

**To Test:**

---


docroot/modules/custom/mass_validation/tests/src/ExistingSite/PublishChildWithUnpublishedParentConstraintTest.php
```diff
@@ -29,6 +29,13 @@ protected function setUp(): void {
     $this->user = $user;
   }
 
+  /**
+   * Un-cacheable dynamic page patterns.
+   */
+  protected static array $uncacheableDynamicPagePatterns = [
+    'orgs/*',
```
weitzman: The comment needs to be expanded to say why we are giving up this test coverage on the org content type. It doesn't quite communicate that today.

Its OK if we do use a follow-up to deal with this, just make sure its open and slated.

arthurbaghdas: Solved the issue without a need for this. BY just removing unnecessary call https://github.com/massgov/openmass/pull/2666/files#diff-f91c28e35feac26cec9496013952132ae82d35e5e5edb0052386adcdfeb74e97L54

PR #2675, opened 2024-09-16
DP-34671: Don't render heading when no published content in section

arthurbaghdas:
**Description:**

DP-34671

**To Test:**

---

### DIFF
--- a/docroot/themes/custom/mass_theme/templates/content/node--guide-page.html.twig
+++ b/docroot/themes/custom/mass_theme/templates/content/node--guide-page.html.twig
@@ -80,7 +80,7 @@
     {% if content.field_image_credit is not empty %}
         {{ content.field_image_credit }}
     {% endif %}
-    {% if node.field_guide_page_related_guides.value %}
+    {% if suggestedPages %}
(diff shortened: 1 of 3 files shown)


docroot/themes/custom/mass_theme/templates/content/node--guide-page.html.twig
```diff
@@ -80,7 +80,7 @@
     {% if content.field_image_credit is not empty %}
         {{ content.field_image_credit }}
     {% endif %}
-    {% if node.field_guide_page_related_guides.value %}
+    {% if suggestedPages %}
```
dstorozhuk, line 83: @arthurbaghdas do we have a test for suggestedPages on Guide page?  
If yes - please update the test to reflect that last change (if no item to show in suggestedPages section - do not display the block).
Please update the test if it exists for suggestedPages visibility.
If the test does not exists - please create a test for this content type edit form and add the test case for you ticket there.

dstorozhuk (changes requested):

dstorozhuk (approved):
LGTM 👍

PR #2710, opened 2024-10-08
Create Jira issue on Dependabot PR

weitzman:
DP-35245

---


tom-fleming:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against c492bd8c7b37cb9f3393f2959f373dff5220c63d
</p>

.github/workflows/jira-dependabot.yml
```diff
@@ -0,0 +1,58 @@
...
+      - id: fetch-jira-cli
+        # https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#homebrew-note
+        run: |
+          eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)" && \
+          brew tap ankitpokhrel/jira-cli && \
+          brew install jira-cli
+      - id: configure-jira-cli
+        run: |
+          eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)" && \
+          jira init --installation cloud --project DP --board "Mass.gov feature/SSR" --server "https://massgov.atlassian.net" --login "moshe.weitzman@mass.gov"
```
dstorozhuk, line 30: @weitzman  what do you think about moving board name and login email to GitHub variables? 
Sever probably will never change or very rarely. 

Also, since it use your account - I believe we need a create documentation how to manage and update this part of the pipeline if needed.

dstorozhuk (dismissed):

.github/workflows/jira-dependabot.yml
weitzman: I dont see why we would separate those two pieces of data from the code. Sure it makes the CI config more reusable, but that is a non-goal IMO. Any project that wants to start from our file should copy and adapt as needed. Or are you suggesting that our Board Name and email are secrets? I would not agree. All my commits already expose my email.

dstorozhuk: No, i don’t think it is a secrets. Ok, I was more concerned about Board Name and Email as changeable values that could change. 
Ok, if don’t think that this could be an issues in feature (i also don’t thinks that this is something we will not be able to change in few minutes) - i will go ahead and merge.

PR #2734, opened 2024-10-29
Update .htaccess for Drupal 10.3

weitzman:
**Description:**
Drops a few things we no longer use like .twig downloading and CDN token verification. Thats now enforced by our custom VCL.

As a follow-up we could consider maintaining a patch against .htaccess instead of a modified .htaccess. If we want that, see https://github.com/drupal-composer/drupal-scaffold/issues/45. Its slightly harder to make changes but slightly easier to keep up with Drupal's edits to .htaccess.

DP-35160

---

### DIFF
--- a/docroot/.htaccess
+++ b/docroot/.htaccess
@@ -2,21 +2,13 @@
 # Apache/PHP/Drupal settings:
 #
 
-# Protect files and directories from prying eyes.  Drupal standard, less twig, plus package.json & yarn.lock.
-<FilesMatch "\.(engine|inc|install|make|module|profile|po|sh|.*sql|theme|tpl(\.php)?|xtmpl|yml)(~|\.sw[op]|\.bak|\.orig|\.save)?$|^(\.(?!well-known).*|Entries.*|Repository|Root|Tag|Template|composer\.(json|lock))$|package.json|yarn.lock|^#.*#$|\.php(~|\.sw[op]|\.bak|\.orig|\.save)$">
+# Protect files and directories from prying eyes.
+<FilesMatch "\.(engine|inc|install|make|module|profile|po|sh|.*sql|theme|twig|tpl(\.php)?|xtmpl|yml)(~|\.sw[op]|\.bak|\.orig|\.save)?$|^(\.(?!well-known).*|Entries.*|Repository|Root|Tag|Template|composer\.(json|lock)|web\.config|yarn\.lock|package\.json)$|^#.*#$|\.php(~|\.sw[op]|\.bak|\.orig|\.save)$">
   <IfModule mod_authz_core.c>
     Require all denied
   </IfModule>
   <IfModule !mod_authz_core.c>
     Order allow,deny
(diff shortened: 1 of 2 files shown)


docroot/.htaccess
```diff
@@ -60,7 +52,7 @@ AddType application/octet-stream .rte
 # Drupal\Core\DrupalKernel::bootEnvironment() for settings that can be
 # changed at runtime.
 <IfModule mod_php.c>
-  php_value zend.assertions 0
```
dstorozhuk, line 63: @weitzman  I see that core version of .htaccess in D11 still use `assert.active` in .htaccess, but it's support has beed removed from php 8.0. 

I see the [issue](https://www.drupal.org/project/drupal/issues/3398033) which propose to remove `php_value assert.active                   0`  from .htaccess, and I assume will be merged soon (According to Andrey Postnikov) comment. 
Should we keep `php_value zend.assertions 0`? 

```
	PHP 7.0 and PHP 7.1: assert.active is available and can be used to enable or disable assertions. Setting assert.active to 0 disables the assert() function, while setting it to 1 enables it.
	•	PHP 7.2: The assert.active directive was deprecated in this version. Although it can still be set, its use triggers a deprecation warning.
	•	PHP 8.0 and later: The assert.active directive was removed entirely. From PHP 8.0 onwards, assert() behavior is controlled by the zend.assertions directive instead:
	•	zend.assertions = 1 (default) enables assertions.
	•	zend.assertions = 0 disables the evaluation of assert() calls but keeps them in the code.
	•	zend.assertions = -1 completely removes assert() calls at runtime, which can improve performance.

```

dstorozhuk:
@weitzman  
>As a follow-up we could consider maintaining a patch against .htaccess instead of a modified .htaccess. If we want that, see https://github.com/drupal-composer/drupal-scaffold/issues/45. It's slightly harder to make changes but slightly easier to keep up with Drupal's edits to .htaccess.

I propose to create a ticket about the possibility to patch the scaffold files instead of  modifying and ignoring them and let team choose if we wand to that, and when.

If you like I can create ticket.

docroot/.htaccess
weitzman: I dont have a preference. feel free to change or remove as you think is best.

weitzman:
Sure, feel free to create that followup. I have no preference

dstorozhuk:
> Sure, feel free to create that followup. I have no preference

Created a ticket here https://massgov.atlassian.net/browse/DP-35784

docroot/.htaccess
dstorozhuk: Since the current docroot/core/assets/scaffold/files/htaccess file has the old value, and it is not affecting the  actual application - I propose to keep it as is, and fix the issue once we complete the https://massgov.atlassian.net/browse/DP-35784 with patch if till that time the issue https://www.drupal.org/project/drupal/issues/3398033 will not be merged.

PR #2834, opened 2025-02-18
DP-37481: Add Tableau Connected Apps variation of Tableau component

arthurbaghdas:
**Description:**

DP-37481

**To Test:**

---


docroot/modules/custom/mass_content/src/Plugin/Field/FieldFormatter/TableauEmbedFormatter.php
```diff
@@ -37,22 +38,48 @@ public function settingsForm(array $form, FormStateInterface $form_state) {
...
+    $paragraph = $items->getEntity();
 
-    foreach ($items as $delta => $item) {
+    // Determine the embed type from the paragraph field.
+    $embed_type = $paragraph->get('field_tableau_embed_type')->value ?? 'default';
+    $token_url = $paragraph->get('field_tableau_url_token')->uri ?? NULL;
+    $token = '';
+
+    // Fetch token only if embed type is 'connected_apps' and a token URL is provided.
+    if ($embed_type === 'connected_apps' && !empty($token_url)) {
```
tom-fleming: @arthurbaghdas One thing that gives me some pause on this approach is that, in the sample code provided, a new token is retrieved per page view. In this version, a token will be retrieved on render and then cached. I am unsure how long this token will be good for and/or if Connected Apps is expecting a new token per page view in order to maintain its own session tracking. Have you looked into this at all?

docroot/themes/custom/mass_admin_theme/templates/fields/mass_content_tableau_embed.html.twig
```diff
@@ -12,7 +12,14 @@
  * @see template_preprocess_link_formatter_link_separate()
  */
 #}
-{{ attach_library('mass_theme/tableau') }}
-<div class="ma_tableau_container" style="max-width:100%">
-  <div class="ma_tableau_item" id="{{ randId }}" data-tableau-url="{{ url }}"></div>
-</div>
+{% if token %}
```
tom-fleming: @arthurbaghdas I know I initially put this check in, but let's pass in the embed type on both of the templates and use that value instead. The problem with token is that if a visualization is switched back to the default style, the token value will still be available even if the field is hidden.

docroot/modules/custom/mass_content/src/Plugin/Field/FieldFormatter/TableauEmbedFormatter.php
arthurbaghdas: You are absolutely, right changed the architecture

docroot/themes/custom/mass_admin_theme/templates/fields/mass_content_tableau_embed.html.twig
arthurbaghdas: Changed

PR #2869, opened 2025-03-10
Add glossary functionality

andy-blum:
**Description:**

This PR adds a new content type `glossary` which allows authors to create term-definition pairs. Glossary nodes can then be referenced from other nodes and be used to inject popover markup into the body of a page, helping users discover definitions of terms as they are encountered.

> [!NOTE]  
> This PR is still a work-in-progress, but is deprioritized by other work for now. Remaining tasks below.

- [x] Enable glossaries to be referenced from other node types (content types reviewed by @joeg8612 3/11)
  - [x] advisory
  - [X] binder
  - [x] curated_list
  - [x] decision
  - [x] executive_order
  - [x] form_page
  - [x] how_to_page
  - [X] info_details
  - [x] location
  - [x] news
  - [x] org_page
  - [x] regulation
  - [x] rules
  - [x] service_page
- [X] ~Theme Glossary pages~
  - Glossary pages are rabbitholed
- [ ] [a11y/UX review of mayflower popover usage](https://github.com/massgov/mayflower/pull/1971)

DP-37841

**To Test:**

---


dstorozhuk (changes requested):
@andy-blum I will continue reviewing the code, but here are a few things we discovered today on the call which we want to be fixed / improved: 
1. Use this module https://www.drupal.org/project/paragraphs_table or similar output the glossary key - description in a table format. So ideally, we want to have the possibility filter, sort the glossary items withing the glossary using tools like Datatable (should come with paragraphs_table. If not - can be added as an additional JS library attached to the paragraph widget). We want to use datatable to sort and filter the glossary items.
2. @joeg8612  mentioned that he wanted the glossary to be sorted alphabetically on save - but if we can implement Datatable - this can be done on the frontend after the page load. 
3. Hide the available formats for now for the glossary item description field. We can enable that later if Editors request this functionality. 

Also, we need a test coverage for Glossary functionality:
1. Check that Editors can create/edit the glossaries.
2. Check that Editors can apply a glossary item to the selected node.
3. Check that glossary items highlight the words on the page for Anonymous and Logged in users.
4. Check that glossary items highlights are clickable (or react on hover?) and display the expected description.

dstorozhuk (approved):
LGTM

arthurbaghdas (approved):

PR #2878, opened 2025-03-17
DP-0000 Expose weight sub-field of the field_primary_parent field when value is a microsite

m4olivei:
**Description:**
To allow editors to influence the ordering of microsite pages, this change will always expose the weight sub-field of the Parent (`field_primary_parent`) field in config. We then add some logic to only show the weight sub-field of the Parent field if the value places the page into a hierarchy of an existing microsite.

**To Test:**
- [ ] If one does not yet exist, create a Microsite
- [ ] Edit any content that has a Parent (`field_primary_parent`) field
- [ ] The parent field should only display a "Weight" sub-field when the value would place the content into the hierarchy of a microsite. Adjust its value and verify. Weight should show/hide as appropriate afer loading the form, and as you change the value via AJAX.
- [ ] Also perform same testing steps with Creating content.

---

### DIFF
--- a/conf/drupal/config/core.entity_form_display.node.advisory.default.yml
+++ b/conf/drupal/config/core.entity_form_display.node.advisory.default.yml
@@ -297,7 +297,7 @@ content:
       match_limit: 10
       size: 60
       placeholder: ''
-      hide_weight: true
+      hide_weight: false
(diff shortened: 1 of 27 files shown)


andy-blum:
Replaced by #2901

PR #2884, opened 2025-03-18
DP-38060: Add Image Option Card Group

hotsauce:
**Description:**
This updates the `Card` and `Card Group` paragraph type to do the following
- Adds a new boolean field `Use image cards with single call to action` to the `Card Group` Paragraph type
- When checked, new fields in `Card` appear - `Image`, `Eyebrow` and `Content` (a long plain text field)
- When checked the WYSIWYG `Content` field disappears
- When checked, the `Link` and `Image` field become required
- Cards now display vertically in this order: image, eyebrow (optional), title, body. The entire card is clickable with the link provided, and the title of the card is underlined on hover/focus

DP-38060

**To Test:**

- [ ] Visit the deployed site and login as an admin
- [ ] Create a new `Services` piece of content
- [ ] Add a new `Service Section` and within it, add a `Card Group`
- [ ] Within the `Card Group`, check the `Use image cards with single call to action` box. An image field, eyebrow field and long plain text box should appear and the WYSIWYG content box should disappear. 
- [ ] Add at least 3 `Card`s, filling in various fields other than what is required.
- [ ] Verify that if `Use image cards with single call to action` is checked, the image and link field become required before submitting/saving.
- [ ] Modify the style of the `Service Section` to verify that the vertical version of `Card` appears as expected.
- [ ] Add another `Card Group` with `Cards` without an image. Verify that they appear as expected with no regressions/changes.

---


composer.json
```diff
@@ -364,7 +364,8 @@
...
                 "An update to symfony/http-foundation plus a trailing space took down the views UI (https://www.drupal.org/project/drupal/issues/3486195)": "https://www.drupal.org/files/issues/2024-11-27/3486195-46.patch",
                 "Error: Cannot read properties of undefined (reading 'settings')\u2028 with dialog.position.js (https://www.drupal.org/project/drupal/issues/3356667)": "https://www.drupal.org/files/issues/2024-09-19/3356667-69-drupal-core-10.3.5-undefined.patch",
-                "CKEditor 5 balloons invisible when CKEditor 5 is used inside a modal (https://www.drupal.org/project/drupal/issues/3328425)": "https://www.drupal.org/files/issues/2024-09-19/3328425-50-drupal-core-10.3.5-ckeditor5-ballon.patch"
+                "CKEditor 5 balloons invisible when CKEditor 5 is used inside a modal (https://www.drupal.org/project/drupal/issues/3328425)": "https://www.drupal.org/files/issues/2024-09-19/3328425-50-drupal-core-10.3.5-ckeditor5-ballon.patch",
+                "Display Bug when using #states (Forms API) with Ajax Request": "patches/core-1091852-186-display-bug-when-using-states-with-ajax-request.patch"
```
tom-fleming: Currently for patches that we are using directly from drupal.org comments, we typically load directly from drupal.org. The following suggestion will do this, and then you can delete the patch file: 
```suggestion
                "Display Bug when using #states (Forms API) with Ajax Request (https://www.drupal.org/project/drupal/issues/1091852)": "https://www.drupal.org/files/issues/2023-12-18/1091852-186.patch"
```

docroot/themes/custom/mass_theme/overrides/css/sections.css
```diff
@@ -89,3 +89,26 @@
...
+}
+
+.service-section--enhanced .ma__sections-3up__container .ma__card__wrapper:hover,
+.service-section--enhanced-two-column .ma__sections-3up__container .ma__card__wrapper:hover,
+.org-section--enhanced .ma__sections-3up__container .ma__card__wrapper:hover,
+.org-section--enhanced-two-column .ma__sections-3up__container .ma__card__wrapper:hover {
+  border-block-end: 2px solid #DCDCDC;
+}
+
+.service-section--enhanced .ma__sections-3up__container .ma__card__theme--default,
```
clairesunstudio: Shouldn't the enhanced version use the white variant instead of default? 

Also please note that I updated the css classname in mayflower to `.ma__card--c-white` instead, following BEM, only an element gets specified, theme color is a modifier.

clairesunstudio:
```suggestion
.service-section--enhanced .ma__sections-3up__container .ma__card--c-white,
```

docroot/themes/custom/mass_theme/overrides/css/sections.css
```diff
@@ -89,3 +89,26 @@
...
+.service-section--enhanced-two-column .ma__sections-3up__container .ma__card__theme--default,
```
clairesunstudio:
```suggestion
.service-section--enhanced-two-column .ma__sections-3up__container .ma__card--c-white,
```

docroot/themes/custom/mass_theme/overrides/css/sections.css
```diff
@@ -89,3 +89,26 @@
...
+.org-section--enhanced .ma__sections-3up__container .ma__card__theme--default,
```
clairesunstudio:
```suggestion
.org-section--enhanced .ma__sections-3up__container .ma__card--c-white,
```

docroot/themes/custom/mass_theme/overrides/css/sections.css
```diff
@@ -89,3 +89,26 @@
...
+.org-section--enhanced-two-column .ma__sections-3up__container .ma__card__theme--default {
```
clairesunstudio:
```suggestion
.org-section--enhanced-two-column .ma__sections-3up__container .ma__card--c-white {
```

docroot/themes/custom/mass_theme/overrides/css/sections.css
hotsauce: @clairesunstudio 
In the openmass version, no. The new `card.theme` is something that is exclusive now to Mayflower, since the background color of a section in openmass is set via the choice of "Simple" or "Enhanced". So the background color/styling of the section is determined by that choice, and anything in the section inherits and adapts based the styling selection made. I sort of alluded to this in Teams, in that on the openmass side there is no equivalent to `card.theme` anywhere. 

And since we are only passing in `.ma__card--c-white` if that is set (via Mayflower) then the cards in an "Enhanced" section will just get the default white styling since the value of `card.theme` isn't passed through on openmass (and no styling based on that class otherwise). There may be a use for setting a toggle on the openmass side later for the Card color in the future but for now its just inheriting everything from Mayflower.

Let me know if you have any questions, I will update the things on the openmass side to account for this.

PR #2888, opened 2025-03-20
DP-00000: Update previousnext/phpunit-finder to tagged version to avoid errors with other package updates or lock file maintenance

m4olivei:
**Description:**
While working on #2885, I needed a core patch, and after adding it to `composer.json`, and running the typical `ddev composer update --lock` to update the lock file, I was getting the following error. I also get this error simply running `composer update --lock` while on the `develop` branch.

```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires previousnext/phpunit-finder == dev-main -> satisfiable by previousnext/phpunit-finder[dev-main].
    - previousnext/phpunit-finder dev-main requires phpunit/phpunit ^10.5 -> found phpunit/phpunit[10.5.0, ..., 10.5.x-dev] but these were not loaded, likely because it conflicts with another require.
```

The issue is that `previousnext/phpunit-finder:dev-main`  has recently been updated to support `phpunit/phpunit:^10.5` (see https://github.com/previousnext/phpunit-finder/commits/main/), which conflicts with the sites’ requirements. There is a tagged version available for `previousnext/phpunit-finder`, which we're currently effectively using. The `composer.lock` file uses https://github.com/previousnext/phpunit-finder/commit/9c68e83570a505b71e2df5a92e6e49d7547f1a54. Version 2.0.2 is the commit just behind that one, with no functional difference between the two (see https://github.com/previousnext/phpunit-finder/compare/1964f6975a9234f97dd099c6171fed6198784bc2...9c68e83570a505b71e2df5a92e6e49d7547f1a54).

**To Test:**
- [ ] Pull this branch down locally and run `composer update --lock`. You should no longer get the above error.

---

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -276,7 +276,7 @@
         "kint-php/kint": "^5",
         "palantirnet/drupal-rector": "^0.20.3",
         "phpspec/prophecy-phpunit": "^2",
-        "previousnext/phpunit-finder": "dev-main",
+        "previousnext/phpunit-finder": "^2.0.2",
         "vincentlanglet/twig-cs-fixer": "^2.5",
         "weitzman/drupal-test-traits": "^2"
     },
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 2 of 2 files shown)


tom-fleming:
@m4olivei This makes sense to me. Thank you for looking into this!

PR #2901, opened 2025-03-26
DP-38280 field primary parent weights

andy-blum:
**Description:**
To allow editors to influence the ordering of microsite pages, this change will always expose the weight sub-field of the Parent (`field_primary_parent`) field in config. We then add some logic to only show the weight sub-field of the Parent field if the value places the page into a hierarchy of an existing microsite.

DP-38280

**To Test:**
- [ ] If one does not yet exist, create a Microsite
- [ ] Edit any content that has a Parent (`field_primary_parent`) field
- [ ] The parent field should only display a "Weight" sub-field when the value would place the content into the hierarchy of a microsite. Adjust its value and verify. Weight should show/hide as appropriate afer loading the form, and as you change the value via AJAX.
- [ ] Also perform same testing steps with Creating content.

---


dstorozhuk:
@andy-blum I see you unhide the `weight` for the field `field_primary_parent`. And on the save time I know that weight arrangement is not working from the Hierarchy tab on the Microsite homepage node.
It looks like it there are two different weighting systems, right?

andy-blum:
No, these two systems should be in sync.

dstorozhuk:
@andy-blum it looks like there are nine tests failed. Could you please take a look?

docroot/modules/custom/mass_microsites/src/ChildOfMicrositeLookup.php
```diff
@@ -0,0 +1,99 @@
...
+   *   The field name that defines the entity hierarchy.
+   *
+   * @return bool
+   *   If the given child node, parent pair has a microsite in the entity
+   *   hierarchy defined by the given field.
+   *
+   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+   */
+  public function hasMicrositeForChildNodeParentNid(NodeInterface $child_node, int $parent_id, string $field_name): bool {
```
dstorozhuk: @andy-blum  how this (`hasMicrositeForChildNodeParentNid`) method is used in this PR?

docroot/modules/custom/mass_hierarchy/src/Form/HierarchyChildrenForm.php
```diff
@@ -299,9 +318,19 @@ public function save(array $form, FormStateInterface $form_state) {
   /**
    * Batch callback to rebuild the tree.
    */
-  public static function rebuildTree($fieldName, ContentEntityInterface $entity, $parent, $weight) {
-    if ($entity->{$fieldName}->target_id == $parent) {
-      return;
+  public static function rebuildTree($fieldName, ContentEntityInterface $entity, $parent, $weight, $microsite = NULL) {
```
dstorozhuk, line 321: @andy-blum  This method overrides the default code from `entity_hierarchy` module  `rebuildTree()` method. 
That means that it will affect all other pages, even that one which is not part of the microsite. 
If that expected? 
I see that code from the original method is significantly different from the one written here.
Did you have some agreement with @joeg8612 that this override will affect other pages that aren’t part of the microsite but still have hierarchy (or part of hierarchy)?

dstorozhuk: As we discovered on the call with Andy and Joe, code change added in this PR add (restore?) the possibility to save the hierarchy weights and also improve the performance of Hierarchy tab on some specific scenarios when save the hierarchy.

docroot/modules/custom/mass_microsites/src/ChildOfMicrositeLookup.php
andy-blum: It is not - in fact this entire service is unused and has been deleted.

dstorozhuk:
@andy-blum thanks, I see the tests are green now. Thanks.
 What was the major issue which you fixed?

andy-blum:
There was a bug where we were trying to get the microsite of the chosen parent node on pages that did not have a chosen parent. The logic now checks for the existence of the parent prior to trying to see if that parent is within a microsite.

dstorozhuk:
>There was a bug where we were trying to get the microsite of the chosen parent node on pages that did not have a chosen parent. The logic now checks for the existence of the parent prior to trying to see if that parent is within a microsite.

Amazing, thanks for fixing that!

PR #2917, opened 2025-04-02
DP-38400: Add Key Message Header to Microsite Homepage

hotsauce:
**Description:**
This PR adds the `Key Message` Paragraph type to the settings for a Microsite. The `Key Message` only appears on the home page of the microsite and replaces the standard banner at the top of the page (in this case, only for the `Service` content type) if it is created.

This also adds a new field, `Use two columns` that is added to `Key Message` that allows for the content to be split into two columns, inline with the new `Key Message as Two Column` variant [being added in Mayflower](https://github.com/massgov/mayflower/pull/1969) but only applicable to users with the "Use Prototype features" permission.

**Jira:**
DP-38400

**To Test:**
- [ ] Visit the deployed site
- [ ] Login as admin and go to `/admin/structure/entity-hierarchy-microsites`
- [ ] Edit an existing microsite
- [ ] Verify that the `Key Message` field is present
- [ ] Verify that within the `Key Message` Paragraph type there is a new field `Use two column`
- [ ] Test the various options for the `Key Message` and save, viewing the page you set as your home page for the Microsite
- [ ] Ensure checking the `Use two column` box turns the `Key Message` into a two column layout like the new variant in [Mayflower](https://mayflower.digital.mass.gov/patternlab/b/patternlab/DP-38400_add_key_message_header_microsite_homepage/?p=organisms-key-message-as-two-column).
- [ ] Ensure that not checking the `Use two column` box keeps the updated Mayflower `Key Message` layout
- [ ] Verify that the existing header is replaced by the `Key Message` and the options that you set
- [ ] Edit the Microsite again
- [ ] Remove the `Key Message`. View the Microsite home page and verify that the default header is back
- [ ] Go to a non-homepage part of the Microsite
- [ ] Verify that if the `Key Message` is set on the home page, that it does not appear elsewhere
- [ ] Go to a page such as `/try-fishing` and verify that there are no visual regressions (i.e. the new Mayflower styles have not bled outward)

---


docroot/themes/custom/mass_theme/templates/content/node--service-page--microsite-homepage.html.twig
```diff
@@ -0,0 +1,174 @@
...
+    <div class="ma__microsite-key-message">
+      {% for paragraph in microsite_key_message %}
+        {{ paragraph }}
+      {% endfor %}
+    </div>
+
+  {% else %}
+    {% include "@organisms/page-banner/page-banner.twig" with {
+    "pageBanner": {
+      "image": bannerImageWide,
```
dstorozhuk: @hotsauce can you please use a responsive image here? 
If need - you can create a new one, but you can also try to reuse the `full_width` responsive image. 
Please see how it was implemented in is this PR https://github.com/massgov/openmass/pull/2915/files
This file for example:
`docroot/themes/custom/mass_admin_theme/templates/paragraphs/paragraph--image.html.twig`

In short - we should pass the full render array from with the image field with the responsive image formatter selected and configured. 
As an alternative you can use twig  tweak filter for responsive image.

dstorozhuk (changes requested):

docroot/themes/custom/mass_theme/templates/content/node--service-page--microsite-homepage.html.twig
hotsauce: @dstorozhuk 
I fully agree that all images should be responsive, but the code you referenced is the fallback if there is no key message added, and was copied verbatim from `docroot/themes/custom/mass_theme/templates/node--service-page.html.twig`. If a Key Message is added then the image that is displayed uses the responsive image styling.

The default page banner itself is implemented as a background image via Mayflower, with no Drupal image styling added. So making this into a new template using a responsive image would be a rework of the entire component for its implementation on Openmass, that to me seems out of scope for this ticket since the template itself is designed just for microsites and one specific implementation.

dstorozhuk: @hotsauce  if this is a copy of the original template - it is worth to an extent and re-using the exiting template. 
I see very minor differences between two templates. 
openmass  – node--service-page html twi  2025-04-11 08-52-51
Could you think hot to re-use the node--service-page.html.twig?
It can be included in `node--service-page--microsite-homepage.html.twig`. It should be easy to pass `microsite_key_message` variable to node--service-page.html.twig and then do the check in node--service-page.html.twig.

dstorozhuk (changes requested):

docroot/modules/custom/mass_fields/mass_fields.module
```diff
@@ -968,7 +982,15 @@ function _mass_fields_conditions_for_paragraphs(&$element, FormStateInterface $f
   $field_definition = $context['items']->getFieldDefinition();
   $paragraph_entity_reference_field_name = $field_definition->getName();
 
-  if ($paragraph_entity_reference_field_name == 'field_service_sections' || $paragraph_entity_reference_field_name == 'field_header' || $paragraph_entity_reference_field_name == 'field_info_details_header_media' || $paragraph_entity_reference_field_name == 'field_organization_sections' || $paragraph_entity_reference_field_name == 'field_location_details_sections' || $paragraph_entity_reference_field_name == 'field_service_section_content' || $paragraph_entity_reference_field_name == 'field_sections' || $paragraph_entity_reference_field_name == 'field_section_long_form_content') {
+  if ($paragraph_entity_reference_field_name == 'field_service_sections' ||
```
dstorozhuk: @hotsauce  please use in_array here. 
$allowed_paragraph_entity_reference_field_names = [...];

`if (in_array($paragraph_entity_reference_field_name, allowed_paragraph_entity_reference_field_names)`

dstorozhuk (changes requested):

hotsauce:
@dstorozhuk 
The reason for the new template is to keep the location of the key message scoped to just the microsite, and more specifically the microsite home page. 

Most of the microsite is made up of Service pages, but the key message field is being set on the settings for the microsite itself at `/admin/structure/entity-hierarchy-microsites` so if we do not scope it to the home page specifically for a microsite, then the heading that is set on the microsite will bleed out to all of the other Service pages in the microsite. 

I recognize that the templates are near identical, but the repetition was intended to keep everything else we need for the Service page without causing any other disruption to the use of Paragraphs etc on the microsite or elsewhere. And since the entity_hierarchy_microsite module already provides for separate templates for microsites, this seemed to be the best option.

dstorozhuk:
> Most of the microsite is made up of Service pages, but the key message field is being set on the settings for the microsite itself at `/admin/structure/entity-hierarchy-microsites` so if we do not scope it to the home page specifically for a microsite, then the heading that is set on the microsite will bleed out to all of the other Service pages in the microsite.

You can add a switcher in twig as I described in my previous comment. 

>I recognize that the templates are near identical, but the repetition was intended to keep everything else we need for the Service page without causing any other disruption to the use of Paragraphs etc on the microsite or elsewhere. And since the entity_hierarchy_microsite module already provides for separate templates for microsites, this seemed to be the best option.

We should avoid code repetitions as much as possible, for the health of the projects. For the developer who will continue to maintain that project it will be much more convenient to reduce the number of code lines as much as possible. 
If need - we can always split the templates.

dstorozhuk:
@hotsauce  I will do a few fixes to the code to resolve the caching issue Joe demonstrated to me today.

hotsauce:
> > Most of the microsite is made up of Service pages, but the key message field is being set on the settings for the microsite itself at `/admin/structure/entity-hierarchy-microsites` so if we do not scope it to the home page specifically for a microsite, then the heading that is set on the microsite will bleed out to all of the other Service pages in the microsite.
> 
> You can add a switcher in twig as I described in my previous comment.
> 
> > I recognize that the templates are near identical, but the repetition was intended to keep everything else we need for the Service page without causing any other disruption to the use of Paragraphs etc on the microsite or elsewhere. And since the entity_hierarchy_microsite module already provides for separate templates for microsites, this seemed to be the best option.
> 
> We should avoid code reception as much as possible, for the health of the projects. For the developer who will continue to maintain that project with will be much more convenient to reduce the number of code lines as much as possible. If need - we can always split the templates.

Thank you Dima. I will just stick to the one template as suggested, don't want to make a mess for you/future developers  🙂

hotsauce:
> @hotsauce I will do a few fixes to the code to resolve the caching issue Joe demonstrated to me today.

Thanks. I am going to roll back the caching stuff I had previously attempted (it was close but I was stuck), so if you have a quicker/better solution for it then I am 100% ok with that! 🙂

dstorozhuk:
@hotsauce  if you need to pull the branch - don't use pull to your existing branch with the same name, since I did the rebase. Better remove or rename your local branch and then clone that one.

docroot/themes/custom/mass_theme/mass_theme.theme
```diff
@@ -4090,6 +4092,30 @@ function mass_theme_preprocess_node(&$variables) {
       }
     }
   }
+
+  // Check if we're on a node that's part of a microsite
+  if ($node instanceof \Drupal\node\NodeInterface) {
```
dstorozhuk: @hotsauce please remove unnecessary qualifier .

hotsauce: Removed, thanks!

hotsauce:
> @hotsauce if you need to pull the branch - don't use pull to your existing branch with the same name, since I did the rebase. Better remove or rename your local branch and then clone that one.

Thanks @dstorozhuk . I deleted my local branch and pulled this down fresh again and verified that your most recent force push was included, before pushing up the fix to remove the unneeded node check. Hopefully we're good  🙂

dstorozhuk (approved):
openmass looks fine for me, waiting @clairesunstudio  to review/fix the MF part.

PR #2921, opened 2025-04-03
DP-38411: Adjust width of single call to action cards

hotsauce:
**Description:**
This adjusts for cards with a single call to action, that are within a `.ma__rich-text` wrapper (Info Details pages) to allow for the cards to be equal width regardless of if there is an uploaded image or not.

DP-38411

**To Test:**
- [ ] Login as an admin
- [ ] Go to a page such as `/info-details/qag-information-details2`
- [ ] Edit the page to add two sets of 4 cards each - one set with images, one without - ensuring that both sets have the "Use single call to action" box checked
- [ ] Save the page and verify that the cards are all the same width.

---


dstorozhuk:
@hotsauce could you please resolve the conflicts?
UPD: Sorry I will do that by myself, wrong PR.

PR #2947, opened 2025-04-16
SPIKE investigate possible ways to label or add collection data to do…

arthurbaghdas:
…cuments in bulk

**Description:**

DP-****

**To Test:**

---

### DIFF
--- a/docroot/modules/custom/mass_media/mass_media.module
+++ b/docroot/modules/custom/mass_media/mass_media.module
@@ -18,6 +18,7 @@ use Drupal\mass_content\Entity\Bundle\media\DocumentBundle;
 use Drupal\mass_content\Entity\Bundle\media\MediaBundle;
 use Drupal\media\MediaInterface;
 use Drupal\paragraphs\Entity\Paragraph;
+use Drupal\paragraphs\ParagraphInterface;
 
 /**
  * Immediately delete any old file entity on this media entity.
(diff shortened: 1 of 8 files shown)


arthurbaghdas:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 3c711aa9fa98005987e4dc878773e92db03f47d0
</p>

docroot/modules/custom/mass_media/mass_media.module
```diff
@@ -279,6 +280,31 @@ function mass_media_media_presave(MediaInterface $entity) {
   }
 }
```
dstorozhuk, line 283: @arthurbaghdas  can you add a comment about hook implementation?

dstorozhuk: @arthurbaghdas  would it be better do `mass_media_paragraph_presave(ParagraphInterface $paragraph)`? 
And then use Guard clause patter like `if ($entity->bundle() !== 'list_item_document') {return;}`

Code with nested statements is tough to read and maintain. I always prefer to use guard clauses + wrap function code in Drupal ClassResolver, which was specifically designed for Drupal hooks.

dstorozhuk (changes requested):

docroot/modules/custom/mass_media/mass_media.module
arthurbaghdas: All set

dstorozhuk: Ok, thanks. The next step will be to use class resolver to group the logic which should be called in a specific code, related to a specific task. So other developers can easily understand code and also add a new logic to the same hook. So it looks good, so merging for now.

dstorozhuk (approved):

PR #2950, opened 2025-04-17
DP-38880: Second Glossary Term in Same Text Node Not Displaying Popover

hotsauce:
**Description:**
This fixes a JS issue where the 2nd tooltip in the same text node was not displaying properly, as well as a slight refactor.

DP-38880

**To Test:**
- [ ] Go to the deployed site
- [ ] Login as `admin`
- [ ] Go to `/node/add/glossary`
- [ ] Create a new glossary with at least 3 common, easy to remember words ("pumpkin", "watermelon", "squash", etc)
- [ ] Go to `/info-details/qag-servicedetails-2` or any relevant page
- [ ] Assign your new Glossary to the page
- [ ] Edit the text content to add your added glossary terms. Add two different terms in the same block of text, as well as at least one duplicate of it. Add the third term to a different block of text.
- [ ] Save the page
- [ ] View the page and verify that they all have the correct tooltips, and that multiple instances of the word don't get a repeated tooltip.

---


andy-blum (approved):
This code seems pretty sound. The match-finding functionality is unchanged, and the new logic in the highlighter is now:

- group matches in the same text node together
- iterate over each node we need to modify
  - iterate over each match in that node
    - check if we've already highlighted a matched term
    - if not, note the match's start/end positions in the string
    - replace matches with the popover component markup, starting with the last match in the node to avoid moving matches around.
  
I've tested locally and see no issue with this specific approach. I've altered the code slightly to boost readability by reversing the sorting algorithm and replacing a decrementing for-loop with a forEach.

PR #2964, opened 2025-04-30
DP-39143: Backstop false positives

arthurbaghdas:
**Description:**

DP-****

**To Test:**

---

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -633,7 +633,7 @@ jobs:
             - run:
                 name: Verify Artifacts
                 command: |
-                  wget -q -O - https://dl.google.com/linux/linux_signing_key.pub | sudo apt-key add -
+                  wget -q -O - https://dl.google.com/linux/linux_signing_key.pub | apt-key add -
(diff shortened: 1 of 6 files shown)


docroot/modules/custom/mayflower/src/Helper.php
```diff
@@ -1876,7 +1876,7 @@ public static function getParentNode(Paragraph $paragraph): ?NodeInterface {
     }
 
     // Not a node, return null to match declared return type
-    return null;
+    return NULL;
```
clairesunstudio: not sure what the difference is in php, is this intentional?

clairesunstudio (approved):

docroot/modules/custom/mayflower/src/Helper.php
dstorozhuk: yes, in Drupal according to Drupal standards we should use uppercased `NULL`, `TRUE`, `FALSE`.

dstorozhuk (approved):

PR #2997, opened 2025-05-28
DP-39472 fix iframe focus not in view

clairesunstudio:
MassForm PR https://github.com/massgov/massgov-gravityforms/pull/179

Before:
iframe focus before

After:
iframe focus

### DIFF
--- a/docroot/themes/custom/mass_theme/overrides/js/iframe_resizer_focus.js
+++ b/docroot/themes/custom/mass_theme/overrides/js/iframe_resizer_focus.js
@@ -1,21 +1,27 @@
-iFrameResize({
-  log: false,
-  checkOrigin: false,
-  messageCallback: function (messageData) {
-    const { message, iframe } = messageData;
+(function ($, Drupal) {
+  Drupal.behaviors.initializeIframeResizer = {
+    attach: function () {
+      iFrameResize({
+        log: false,
+        checkOrigin: false,
+        messageCallback: function (messageData) {
+          const { message, iframe } = messageData;
 
-    if (message.type === 'scrollToFocus') {
-      // iframe position in the parent page
-      const iframeTop = iframe.getBoundingClientRect().top + window.scrollY;
+          if (message.type === 'scrollToFocus') {
+            // iframe position in the parent page
+            const iframeTop = iframe.getBoundingClientRect().top + window.scrollY;
 
-      // message.offset is the offset inside the iframe
-      const scrollTarget = iframeTop + message.offset;
-      const adjustment = -250;
+            // message.offset is the offset inside the iframe
+            const scrollTarget = iframeTop + message.offset;
+            const adjustment = -250;
 
-      window.scrollTo({
-        top: scrollTarget + adjustment,
-        behavior: 'smooth'
-      });
+            window.scrollTo({
+              top: scrollTarget + adjustment,
+              behavior: 'smooth'
+            });
+          }
+        }
+      }, '.js-ma-responsive-iframe iframe');
     }
-  }
-}, '.js-ma-responsive-iframe iframe');
+  };
+})(jQuery, Drupal);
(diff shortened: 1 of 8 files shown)


docroot/themes/custom/mass_theme/overrides/js/iframe_resizer_parent.js
```diff
@@ -62,7 +63,7 @@ document.addEventListener('DOMContentLoaded', function () {
   if (window.MutationObserver) {
 
     // Crea una instancia de observer
-    var observer = new MutationObserver(function (mutations) {
+    var observer = new MutationObserver(function () {
```
clairesunstudio:
```suggestion
    var observer = new MutationObserver(function (mutations) {
```

docroot/themes/custom/mass_theme/overrides/js/iframe_resizer_focus.js
dstorozhuk, line 1: @clairesunstudio  Please do this fixes.
1. As far as I can see - this iFrameResize object is coming from iframe_resizer/iframe_resizer_content_window or iframe_resizer/iframe_resizer  library (please check which particularly). 
2. Define ``docroot/themes/custom/mass_theme/overrides/js/iframe_resizer_focus.js as a separate library  in `docroot/themes/custom/mass_theme/mass_theme.libraries.yml` with dependency to one of the iframe_resizer libraries which add `iFrameResize` object.
3. Include a new frame_resizer_focus library you defined in the same way/place as `mass_theme/iframe-resizer`. 
4. Wrap `iFrameResize({...` code to use Drupal.behevious where possible.

docroot/themes/custom/mass_theme/overrides/js/iframe_resizer_parent.js
dstorozhuk: Don't add any chnged to this file.

dstorozhuk (changes requested):

docroot/themes/custom/mass_theme/overrides/js/iframe_resizer_focus.js
clairesunstudio: Addressed 2,3,4. Re #1:  It comes from iframeResizer.min.js which is applied on the mass.gov side (parent frame)

docroot/themes/custom/mass_theme/overrides/js/iframe_resizer_focus.js
```diff
@@ -0,0 +1,27 @@
...
+            window.scrollTo({
+              top: scrollTarget + adjustment,
+              behavior: 'smooth'
+            });
+          }
+        }
+      }, '.js-ma-responsive-iframe iframe');
+    }
+  };
+})(jQuery, Drupal);
```
dstorozhuk: @clairesunstudio  

1. I don't see how you use Drupal, jQuery in this code. 
If they are not needed - no need to pass them here. 
2. Drupal.behaviors.initializeIframeResizer -  lets rename this to something like Drupal.behaviors.initializeIframeResizerFocus, to beter reflect what thic code is doing. 
3. Drupal can call Drupal.behaviors.initializeIframeResizer multiple time on the page, so use core/once as a dependency and wrap your code in `once`: 
```
    const elements = once('myfeature-once-id', 'css selector or js DOM element', context);
      // `elements` is always an Array.
      elements.forEach(processingCallback);
processingCallback = function() {
iFrameResize({
         log: false,
         checkOrigin: false,
         messageCallback: function (messageData) {
           const { message, iframe } = messageData;

           if (message.type === 'scrollToFocus') {
             // iframe position in the parent page
             const iframeTop = iframe.getBoundingClientRect().top + window.scrollY;

             // message.offset is the offset inside the iframe
             const scrollTarget = iframeTop + message.offset;
             const adjustment = -250;

             window.scrollTo({
               top: scrollTarget + adjustment,
               behavior: 'smooth'
             });
           }
         }
       }, '.js-ma-responsive-iframe iframe');
}
```
I haven't test that code, it is just an example.

dstorozhuk (changes requested):
I also don't see how you add `mass_theme/iframe_resizer_focus` library to the page. Does your code work after the update?

dstorozhuk (approved):
Looks good to me.

dstorozhuk (changes requested):
Please fix the js linter errors, the rest looks good to me.

dstorozhuk (approved):

PR #3005, opened 2025-06-03
DP-39253: Service page layout paragraphs

arthurbaghdas:
**Deployment steps:**
- [ ] drush mcsplp  (Published nodes)
- [ ] drush mcsplp --unpublished (Unpublished nodes)
- [ ] Entity hierarchy rebuild needed after migration


composer.json
```diff
@@ -461,8 +460,7 @@
                 "Nested paragraphs title gets cut off (https://www.drupal.org/project/paragraphs/issues/3246140)": "https://www.drupal.org/files/issues/2024-03-11/paragraphs-3246140-19-fix_paragraph_title_cutoff.patch"
             },
             "drupal/layout_paragraphs": {
-                "Replace dragula with SortableJS (Scroll to Top After Adding a New Component in Layout Paragraphs Builder when Field Groups enabled https://www.drupal.org/project/layout_paragraphs/issues/3392717)": "patches/layout_paragraphs/Replace-dragula-with-SortableJS-3392717-11.patch",
```
dstorozhuk, line 464: Amazing! They merged our work!

conf/drupal/config/core.extension.yml
```diff
@@ -184,7 +183,6 @@ module:
   options: 0
   papaparse: 0
   paragraphs: 0
-  paragraphs_library: 0
```
dstorozhuk, line 187: @arthurbaghdas  this module is just not used anywhere else?

conf/drupal/config/core.extension.yml
```diff
@@ -100,7 +100,6 @@ module:
   language: 0
   layout_discovery: 0
   layout_paragraphs: 0
-  layout_paragraphs_library: 0
```
dstorozhuk, line 103: Looks like this module was enabled as an experement and not used?

patches/layout_paragraphs/DP-37429_improve_text_paragraph_ux.patch
```diff
@@ -0,0 +1,177 @@
+diff --git a/css/builder.css b/css/builder.css
+index e01d327..14adad2 100644
```
dstorozhuk, line 2: Arthur, I see you added a lot here. Is that something should be contributed back or it is very specific to our project?

docroot/modules/custom/mass_content/mass_content.module
```diff
@@ -871,143 +858,14 @@ function mass_content_preprocess_paragraph__links_downloads_flexible(&$variables
   }
 }
 
-/**
- * Implements hook_preprocess_HOOK().
- */
-function mass_content_preprocess_paragraph__flexible_link_group(&$variables) {
```
dstorozhuk, line 877: I like the fact that code was removed.

docroot/modules/custom/mass_content/src/Drush/Commands/MassContentCommands.php
```diff
@@ -33,198 +37,6 @@ public function __construct(
...
-   *   Number of nodes to process
-   *   Argument provided to the drush command.
-   *
-   * @command mass-content:migrate-dates
-   *
-   * @usage mass-content:migrate-dates foo 5000
-   *   foo is the type of node to update,
-   *   5000 is the number of nodes that will be processed.
-   */
-  public function migrateDateFields(string $type = '', int $limit = 0) {
```
dstorozhuk, line 52: @arthurbaghdas so you simly removed the old comand for previous migration in favor of a new command for a new migration? 
I has that qeustion when we first migrated to LP, but didn't ask it: if this command should be executed once - it looks like it is [hook post update ](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Extension%21module.api.php/function/hook_post_update_NAME/11.x).

This hook already suport bath processing, so not issue with PHP limits other than 3h of process excution time. But I think we dont need that much for that migration. Or we do? 
My conservative here is that we dont track in code and update path, even if it is a content update path. Would be useful to keep historical data in code. Of couse, it is alwasy can be found in GIT, but Drupal suggest to use hook post update, so  someone after 5 years for example will join the project, they will know what done in terms of content. This can be useful simetimes, especially if need to investigate some issue with revisions. So, my point: 
1. Move a new code to hook_post_ update if possible (if is executing up to 2.5 hous).
2. Restore the old code from that file.
3. Make the old migration functions visble in code but not possible to execute with comments for what reason they were created.

docroot/modules/custom/mass_fields/mass_fields.module
```diff
@@ -761,6 +761,44 @@ function mass_fields_theme() {
   ];
 }
 
+/**
+ * Implements hook_field_group_form_process_build_alter().
+ */
+function mass_fields_field_group_form_process_build_alter(array &$element, FormStateInterface $form_state, &$complete_form) {
+  // Only apply to node edit forms.
+  if (isset($form_state->getBuildInfo()['form_id']) && str_contains($form_state->getBuildInfo()['form_id'], 'node_')) {
+    // Your target group machine name from Field Group UI.
```
dstorozhuk: Looks like  chatGPT/claude code left some lefovers 😀.

docroot/modules/custom/mass_fields/mass_fields.module
```diff
@@ -761,6 +761,44 @@ function mass_fields_theme() {
   ];
 }
 
+/**
+ * Implements hook_field_group_form_process_build_alter().
+ */
+function mass_fields_field_group_form_process_build_alter(array &$element, FormStateInterface $form_state, &$complete_form) {
+  // Only apply to node edit forms.
+  if (isset($form_state->getBuildInfo()['form_id']) && str_contains($form_state->getBuildInfo()['form_id'], 'node_')) {
```
dstorozhuk: I would reaither get a Form Class and chek it instead of checking this string 'node_'. it is move Drupal 10 way to do the check.

docroot/modules/custom/mass_fields/src/Plugin/Block/NodeMetaFormBlock.php
```diff
@@ -0,0 +1,129 @@
...
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition): static {
+    return new static(
+      $configuration,
+      $plugin_id,
+      $plugin_definition,
+      $container->get('current_user'),
+      $container->get('date.formatter'),
+    );
+  }
+
```
dstorozhuk, line 84: Comments are missing, defently not a critical, we can skip it, but JetBrain AI can quickly genearte them.

docroot/modules/custom/mass_fields/src/Plugin/Block/NodeMetaFormBlock.php
```diff
@@ -0,0 +1,129 @@
...
+  public function build(): array {
+    /** @var \Drupal\node\NodeInterface $node */
+    $node = $this->getContextValue('node');
+
+    if (!$node instanceof NodeInterface || !$this->currentUser->hasPermission('administer nodes')) {
```
dstorozhuk, line 96: I like that you use Guard pattern here, hovewer, can this be defined in block configuration level with roles? Or may there some module we alredy have on the site allow blok access by permissions? 

Also, have you tested that block with cache anabled? Should not be an issue, because it tied to specific URL, just want to make it is ok in terms of caching.

docroot/modules/custom/mass_fields/src/Plugin/Block/NodeMetaFormBlock.php
```diff
@@ -0,0 +1,129 @@
...
+      // Will be used to pre-fill subject in contact form.
+      'query' => [
+        'edit[subject]' => $title_nid,
+      ],
+      // Set 'attributes' option for URL.
+      'attributes' => [
+        'title' => t('Contact the author of this content.'),
+      ],
+    ]);
+    if ($contact_url->access(\Drupal::currentUser())) {
```
dstorozhuk: 1. I beleve we need some cache context here. 
2. Also, instead of `if` statement plase use '#access' property on render array
3. Idealy to use DI for Drupal::currentUser. Can easily generated with JetBrain. 

`'#access' => $contact_url->access(\Drupal::currentUser())`

docroot/modules/custom/mass_fields/src/Plugin/Block/NodeMetaFormBlock.php
```diff
@@ -0,0 +1,129 @@
...
+    $title_prefix = 'Outreach about Mass.gov page "';
+    $nid = $node->id();
+    $title_suffix = '" (' . $nid . ')';
+    $prefix_suffix_combo_lenth = strlen($title_prefix . $title_suffix);
+    // Fits title into 100 characters or less.
+    $title_max_length = 100 - $prefix_suffix_combo_lenth;
+    $title = (string) $node->getTitle();
+    // Truncates title as needed when forming combined `$title_nid` string.
+    $title_nid = $title_prefix .
+      ((strlen($title) > $title_max_length) ? substr($title, 0, $title_max_length - 1) . '…' : $title) .
```
dstorozhuk: Drupal has a good set of utilities to work with multibite string including truncate. it would be better to use them instead. 
[Unicode::truncate](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Utility%21Unicode.php/function/Unicode%3A%3Atruncate/11.x)

docroot/modules/custom/mass_fields/src/Plugin/Block/NodeMetaFormBlock.php
```diff
@@ -0,0 +1,129 @@
...
+    $form['meta'] = [
+      '#type' => 'container',
+      '#open' => TRUE,
+      '#title' => $this->t('Published'),
+      '#attributes' => ['class' => ['entity-meta__header']],
+    ];
+
+    $form['meta']['published'] = [
+      '#type' => 'item',
+      '#markup' => \Drupal::service('content_moderation.moderation_information')->getWorkflowForEntity($node)->getTypePlugin()->getState($node->moderation_state->value)->label(),
```
dstorozhuk: if not hard - lest covert to DI.

docroot/modules/custom/mass_validation/tests/src/ExistingSite/CollectionSearchValidationTest.php
```diff
@@ -121,7 +121,7 @@ public function testPromoPageCollectionSearchValidation() {
    * Assert that the Custom Search paragraph validation works properly on service page nodes.
    */
   public function testServicePageCollectionSearchValidation() {
-
+    $this->markTestSkipped('Skipping this test temporarily.');
```
dstorozhuk: We need to fix that test before go live.

docroot/themes/custom/mass_admin_theme/mass_admin_theme.libraries.yml
```diff
@@ -86,6 +86,11 @@ layout_paragraphs:
   css:
     layout:
       css/overrides/layout-paragraphs.css: {}
+  js:
```
dstorozhuk, line 89: We need to always update version if we made a change after Drupa 10.2 or something, since may have and issue with css/js caching.

docroot/themes/custom/mass_theme/mass_theme.theme
```diff
@@ -7703,3 +7342,124 @@ function mass_theme_preprocess_pager__collection_all(&$variables) {
...
+    }
+  }
+  return FALSE;
+}
+
+/**
+ * Preprocess paragraph sections from services and organizations.
+ */
+function _mass_theme_group_render_items(array $paragraphs) {
+  // Define fields and paragraphs that should be wrapped, fields have priority over paragraphs.
```
dstorozhuk: Since we are doing new code here. 
I strongly suggest to not use `_helper_functions` approach here and in future which was ok for  Drupal 5/67, but use `ClassResolver` approach from D8+ instead if code shold NOT be injected anywhere, but we want it to be encapsulated. This allow us to structure code better, do DI when needed and have all other benefits of OOP.

docroot/themes/custom/mass_theme/mass_theme.theme
```diff
@@ -7703,3 +7342,124 @@ function mass_theme_preprocess_pager__collection_all(&$variables) {
...
+    $components[] = $component;
+  }
+  return $components;
+}
+
+/**
+ * Implements hook_preprocess_paragraph__service_section().
+ */
+function mass_theme_preprocess_paragraph__service_section(array &$variables) {
+  $paragraph = $variables['paragraph'];
```
dstorozhuk, line 7366: Same comment about Using ClassResolver here.

docroot/themes/custom/mass_theme/mass_theme.theme
```diff
@@ -7703,3 +7342,124 @@ function mass_theme_preprocess_pager__collection_all(&$variables) {
...
+    if ($paragraph->hasField($field_name) && !$paragraph->get($field_name)->isEmpty()) {
+      return TRUE;
+    }
+  }
+  return FALSE;
+}
+
+/**
+ * Preprocess paragraph sections from services and organizations.
+ */
```
dstorozhuk: We should be leave an AI prompt in future in class/function comments 😀. Joking, but proabably can be useful.

dstorozhuk (changes requested):
over all looks good for such a big change, some mandatory and optional chnge requests.

conf/drupal/config/core.extension.yml
arthurbaghdas: yes, this is not used anywhere

conf/drupal/config/core.extension.yml
arthurbaghdas: exactly, we dont use it.

patches/layout_paragraphs/DP-37429_improve_text_paragraph_ux.patch
arthurbaghdas: Part of the code is already in the new major version of the module. but it doesn't work as expected, for example in the new major release the click to open  is not consistent and works with some bugs, I think they will fix it eventually and we will upgrade when we go to the Drupal 11.
Given this, I don't think we need to waste our time contributing to the old version, since there is an ongoing work for the latest version.

docroot/modules/custom/mass_content/mass_content.module
arthurbaghdas: Yeap tried to clean up as much as I can.

docroot/modules/custom/mass_content/src/Drush/Commands/MassContentCommands.php
arthurbaghdas: We had problem several years ago, the deploy hooks and post update hooks doesnt work and hang up eventually, so we decided to move forward with drush commands for huge data processing, so we can be capable of rerunning the command and will have full control over the code. In most cases those commands live under MassContentCommand.php file, and m=ost cases they duplicate each other and were run only once (simulating deploy hook as i described), thats why sometimes we prefer to do a clean up. In this case this is an old code that was executed several years ago, so I don't think there will be a need to revisit it in the future even for the referece.

docroot/modules/custom/mass_fields/mass_fields.module
arthurbaghdas: Thanks, for the suggestion. Done.

docroot/modules/custom/mass_fields/mass_fields.module
arthurbaghdas: OMG 😄

docroot/modules/custom/mass_fields/src/Plugin/Block/NodeMetaFormBlock.php
arthurbaghdas: Most of the code including this permissions check and cache handling comes from the core  file. 

> docroot/core/modules/node/src/NodeForm.php  

Ive actually tested the caching, just in case.

docroot/modules/custom/mass_validation/tests/src/ExistingSite/CollectionSearchValidationTest.php
arthurbaghdas: Oh, totally forgot about this one. thanks for the reminder.

docroot/themes/custom/mass_admin_theme/mass_admin_theme.libraries.yml
arthurbaghdas: Makes sense. Updated.

docroot/modules/custom/mass_fields/src/Plugin/Block/NodeMetaFormBlock.php
arthurbaghdas: Added comments and also changed moderation state service, to be part of Container injection.

docroot/modules/custom/mass_validation/tests/src/ExistingSite/CollectionSearchValidationTest.php
arthurbaghdas: New test added https://github.com/massgov/openmass/blob/1f16883e4371b6478c79e39acb1dd6f439fe43f6/docroot/modules/custom/mass_validation/tests/src/ExistingSiteJavascript/CollectionSearchValidationLayoutParagraphsTest.php

docroot/themes/custom/mass_theme/mass_theme.theme
arthurbaghdas: Changed to be a separate class and called new class resolver.

docroot/themes/custom/mass_theme/mass_theme.theme
arthurbaghdas: Class resolver called here. Thanks for the suggestion.

docroot/modules/custom/mass_fields/src/Plugin/Block/NodeMetaFormBlock.php
dstorozhuk: @arthurbaghdas  this one has missed reply or action.

docroot/themes/custom/mass_admin_theme/js/modal-scroll-guard.js
```diff
@@ -0,0 +1,224 @@
...
+            });
+          }
+        });
+      };
+      patchAjaxInsertOnce._done = true;
+    })();
+
+    if (window.jQuery) {
+      const $ = window.jQuery;
+      $(document)
```
dstorozhuk, line 174: @arthurbaghdas I am unsure why jQuery was added in that way, and as suing with library depedency and the adding it to `)(Drupal, once, jQuery);`

docroot/themes/custom/mass_admin_theme/mass_admin_theme.libraries.yml
```diff
@@ -82,13 +82,33 @@ doc-form:
...
+    - core/once
 
 entity_browser_custom:
   version: VERSION
   css:
     theme:
       css/overrides/entity-browser-custom.css: {}
+
+# mass_admin_theme.libraries.yml (or your_module.libraries.yml)
+modal_scroll_guard:
```
dstorozhuk, line 101: @arthurbaghdas  please use explisit version here.

docroot/themes/custom/mass_theme/mass_theme.theme
dstorozhuk: I would entire code code CL, but for now it is also ok.

dstorozhuk (changes requested):

docroot/themes/custom/mass_admin_theme/mass_admin_theme.libraries.yml
arthurbaghdas: Done

docroot/themes/custom/mass_admin_theme/js/modal-scroll-guard.js
arthurbaghdas: Done.

docroot/modules/custom/mass_fields/src/Plugin/Block/NodeMetaFormBlock.php
arthurbaghdas: Done.

dstorozhuk (approved):
LGTM

PR #3016, opened 2025-06-11
Merge master to develop after Hotfix 0.440.1

dstorozhuk:


dstorozhuk:
@arthurbaghdas  please respove the conflicts. I dont have permission to push to develop.
`'user/reset/*', or 'user/reset/.*',` will work equaly for that path, so use one or another value.

PR #3025, opened 2025-06-17
DP-39642/fix-focus-scroll-followup

clairesunstudio:
UAT URL : https://pr3025-h8kzja1gibqeklhmafjht5chyxl44lf3.tugboatqa.com/forms/pesticide-use-report-gravity-form-test

This PR fixes the config loading issue after the iframe is initiated, which caused the inconsistent behavior across environments. This change fixed the production inconsistency as well as the iOS issue - please test nested form focus scroll in view feature on your mobile device or in browserstack


arthurbaghdas:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 05d436d66d5e042388b81982350e509333c56d7c
</p>

docroot/themes/custom/mass_theme/overrides/js/iframe_resizer_focus.js
```diff
@@ -1,42 +1,92 @@
...
-              const adjustment = -250;
-
-              window.scrollTo({
-                top: scrollTarget + adjustment,
-                behavior: 'smooth'
-              });
-            }
+  // Function to handle focus messages from iframes
+  function handleFocusMessage(messageData) {
+    console.log('Received focus message:', messageData);
```
dstorozhuk: @clairesunstudio  if you want to keep debug code - please make it's displayed conditionally and quiet by default. Or just remove debug code.

dstorozhuk: Or we need this on production?

docroot/themes/custom/mass_theme/overrides/js/iframe_resizer_focus.js
```diff
@@ -1,42 +1,92 @@
...
+      console.log('Message is not a scrollToFocus type:', message);
+    }
+  }
 
-            // Call the original callback if it exists
-            if (typeof originalCallback === 'function') {
-              originalCallback(messageData);
-            }
+  // Override the initIframeResizer behavior to inject our onMessage handler
+  const originalInitIframeResizer = Drupal.behaviors.initIframeResizer;
```
dstorozhuk, line 54: @clairesunstudio have you tested that code with cache enabled and JS aggregation enabled? It is disabled by default on local. 
I suspect it still may not work because dependencies aren’t resolved or resolved incorrectly. 

If you want to use originalInitIframeResizer = Drupal.behaviors.initIframeResizer; - then you need to make sure your code is properly aggregated. There are no guarantees that  `Drupal.behaviors.initIframeResizer;`  will be available in production when you code expect it, if the dependency is not resolved correctly. 
This code is located in `iframe_resizer/init` library, so this library must be added as a dependency to the current library. That is what I explained on our past call about Drupal JS dependencies resolution.

dstorozhuk (changes requested):

docroot/themes/custom/mass_theme/overrides/js/iframe_resizer_focus.js
clairesunstudio: Added the dependency, and fixed the order of loading. Tested with aggregated JS and works well in prod build

docroot/themes/custom/mass_theme/overrides/js/iframe_resizer_focus.js
clairesunstudio: good catch, removed

PR #3066, opened 2025-07-29
Feature/dp 40669 bulk replace docs

arthurbaghdas:
**Description:**

DP-****

**To Test:**

---


docroot/modules/custom/mass_bulk_file_replace/src/Form/ReplaceUploadForm.php
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    $mid_counts = [];
+    foreach ($files as $item) {
+      $filename = (string) ($item['filename'] ?? '');
+      if ($filename === '') {
+        continue;
+      }
+      // Normalize filenames that may have an auto-suffix before the ID token.
+      $normalized = preg_replace('/_\d+(?=_DO_NOT_CHANGE_THIS_MEDIA_ID_\d+)/i', '', $filename);
```
tfleming-ma, line 142: @arthurbaghdas Not a merge blocker, but I noticed that this does the same preg_replace/preg_match in three places, which I think is a bit fragile. I think this code could be put into something like a static method or in a method that lives in a base form that both the confirm and upload form inherit.

tfleming-ma (approved):

docroot/modules/custom/mass_bulk_file_replace/src/Form/ReplaceUploadForm.php
arthurbaghdas: This will be restructured soon. Will take into consideration. Thanks

PR #3107, opened 2025-09-04
DP-41563-Disallow-any-emoji-or-visual-characters-in-text-fields-in-Drupal

jarielrrz:
Added validation to prevent emoji and visual characters in text fields across all content types.


arthurbaghdas (dismissed):
This implementation works, but it feels heavier than it needs to be. Instead of mixing hook_form_alter, hook_entity_presave, and a custom exception subscriber, we should use Drupal’s built-in Validation API.

The clean approach would be to implement a Constraint + ConstraintValidator  and attach it to text fields. That way:
	•	Validation runs consistently through the standard pipeline (both on form submit and programmatic saves).
	•	Error messages are shown inline on the offending fields, no need for redirect/exception handling.
	•	Logic is centralized in one service, so we’re not duplicating regex checks in multiple places.
	•	Constraints can be enabled sitewide or per field via config/Field UI settings, making it flexible and maintainable.

I’d suggest refactoring this PR toward a Constraint-based solution, which will simplify the module and keep it aligned with Drupal best practices.

On the regex side, please note possible false positives/negatives. The current ranges may block characters that are not emoji but are legitimate content:
	•	✈️ (U+2708 AIRPLANE) — often used in travel contexts.
	•	© (U+00A9 COPYRIGHT SIGN) and ® (U+00AE REGISTERED SIGN).
	•	Currency symbols like ₽ (U+20BD RUBLE SIGN) or ₹ (U+20B9 RUPEE SIGN).
	•	Arrows like ➔ (U+2794) or symbols like ★ (U+2605).
	•	Dingbats / pictographs that may be used intentionally in UI text.

At the same time, newer emoji sequences (e.g. skin tone modifiers, flags built via ZWJ sequences) may slip past the current regex.

docroot/modules/custom/emoji_validation/emoji_validation.module
```diff
@@ -0,0 +1,69 @@
...
+}
+
+/**
+ * Implements hook_entity_presave().
+ *
+ * Enforce emoji validation before saving any entity.
+ * Skips taxonomy terms to allow emojis in labels/tags.
+ */
+function emoji_validation_entity_presave(EntityInterface $entity) {
+  // Skip validation for taxonomy terms (allow emojis in labels).
```
jarielrrz: @arthurbaghdas 

I've got emoji blocking working for node titles and custom fields, but struggling with the Labels autocomplete field (entity_reference with auto_create: true).

Tried form validation but it's hacky and causes side effects. Tried presave hooks but auto-created terms throw exceptions white error page.

Currently just excluding taxonomy_term from validation so labels can have emojis. Is there a cleaner way to validate autocomplete before term creation?

jarielrrz: @arthurbaghdas 

I've got emoji blocking working for node titles and custom fields, but struggling with the Labels autocomplete field (entity_reference with auto_create: true).

Tried form validation but it's hacky and causes side effects. Tried presave hooks but auto-created terms throw exceptions white error page.

Currently just excluding taxonomy_term from validation so labels can have emojis. Is there a cleaner way to validate autocomplete before term creation?

docroot/modules/custom/emoji_validation/emoji_validation.module
```diff
@@ -0,0 +1,125 @@
...
+    if (in_array($field_definition->getType(), _emoji_validation_text_field_types(), TRUE)) {
+      $constraints = $field_definition->getConstraints();
+      $constraints['NoEmojiConstraint'] = [];
+      $field_definition->setConstraints($constraints);
+    }
+  }
+}
+
+/**
+ * Implements hook_f_field_widget_single_element_form_alter().
```
tfleming-ma:
```suggestion
 * Implements hook_field_widget_single_element_form_alter().
```

docroot/modules/custom/emoji_validation/emoji_validation.module
```diff
@@ -0,0 +1,125 @@
...
+    $raw = (string) $raw['target_id'];
+  }
+  else {
+    $raw = (string) $raw;
+  }
+
+  // Split tags into labels and strip trailing " (123)" IDs.
+  $labels = [];
+  if (is_string($raw)) {
+    foreach (preg_split('/\s*,\s*/', $raw, -1, PREG_SPLIT_NO_EMPTY) as $part) {
```
tfleming-ma: What do these strings look like? If I'm understanding the regex correctly, it seems like this would be better accomplished with `explode()`

docroot/modules/custom/emoji_validation/emoji_validation.module
```diff
@@ -0,0 +1,125 @@
...
+      $labels[] = preg_replace('/\s*\([^)]+\)\s*$/', '', $part);
```
tfleming-ma: Similar to my previous comment, this looks like something that would be better accomplished with `trim()` if I'm understanding the regex correctly. What is this regex trying to match?

docroot/modules/custom/emoji_validation/emoji_validation.module
arthurbaghdas: I switched to explode(',', $raw) + trim() and drop empties. Much clearer than the split regex.

docroot/modules/custom/emoji_validation/emoji_validation.module
arthurbaghdas: The only regex left is /\s*\(\d+\)\s*$/ — it strips the trailing numeric ID Drupal appends (e.g., turns Flu (123) → Flu). trim() can’t do that.

tfleming-ma (approved):

PR #3135, opened 2025-09-29
DP-42341: Return 404 status code when something is unpublished instead of 403

arthurbaghdas:
**Description:**

DP-****

**To Test:**

---


dstorozhuk:
I added that PR which it related to 404 but based on contrib module https://github.com/massgov/openmass/pull/3140. If there are no strong reason to use custom code - I suggest to use contrib module. 
But, if there are a some reasons to use custom code - like apply 404 it only to Media, Node - then close my PR and merge your, I approved it.

dstorozhuk (approved):
Approve but only if we need to apply 404 fix for Nodes and Media. Hovewer, i don't see a reason to apply it only to Nodes and Media on www.mass.gov, since www.mass.gov/admin will never exists in fact fo public, so it is 404.

arthurbaghdas:
> Approve but only if we need to apply 404 fix for Nodes and Media. Hovewer, i don't see a reason to apply it only to Nodes and Media on [www.mass.gov](http://www.mass.gov), since [www.mass.gov/admin](http://www.mass.gov/admin) will never exists in fact fo public, so it is 404.

Makes total sense

PR #3153, opened 2025-10-08
DP-36234: Verify that Trash clear drush command is working correctly

arthurbaghdas:
**Description:**

DP-****

**To Test:**

---


docroot/modules/custom/trashbin/src/TrashbinPurgeCandidateQuery.php
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    $query = $this->connection->select($base_table, 'b');
+
+    $query->innerJoin(
+      'content_moderation_state_field_data',
+      'md',
+      'md.content_entity_type_id = :etype AND md.content_entity_id = b.' . $id_key . ' AND md.content_entity_revision_id = b.' . $rev_key,
+      [':etype' => $entity_type_id]
+    );
```
dstorozhuk, line 134: **Blocker: this still selects (and deletes) content whose authoritative moderation state is not `trash`.**

The join now correctly filters `content_entity_type_id` — that fixes the original node/media revision-id collision :+1: — but it does not filter `workflow` and does not resolve which `content_moderation_state` row is authoritative. The DB contains nodes with **two CMS rows pointing at the same node revision**: a stale row left over from a workflow migration (`editorial`/`trash`) plus the current row of the bundle workflow. Verified on a fresh local copy of the prod DB:

```
nid=16641, vid=2895161
  cms_id=2124201  workflow=editorial   state=trash   <- stale
  cms_id=5918526  workflow=topic_page  state=draft   <- current
```

`$node->get('moderation_state')->value` returns `draft`, yet this query returns the node as a purge candidate — and `isEntityEligible()` rebuilds the same SQL, so the pre-delete re-check passes too. `drush trashbin:purge node --days-ago=0` would permanently delete **20 draft nodes** on today's data (19 topic_page, 1 campaign_landing):

```
16641 17556 17561 53106 53111 53556 53786 53796 53811 58116
58701 72161 72176 72191 253526 332716 431961 437526 446716 491636
```

Suggested fix, both layers:
- In `TrashbinCommands::purge()`, after `$storage->load($id)`, trust the entity API rather than only SQL: `if ($entity->get('moderation_state')->value !== 'trash') { continue; }` — this one line blocks all 20.
- Here in the query, also constrain the join to the workflow assigned to the bundle (or pick the newest CMS row, e.g. via `MAX(md.id)`), otherwise these rows permanently occupy slots in every `--max` window.

docroot/modules/custom/trashbin/src/TrashbinPurgeCandidateQuery.php
```diff
@@ -0,0 +1,125 @@
...
+    if (!$id_key || !$rev_key) {
+      return NULL;
+    }
+
+    $query = $this->connection->select($base_table, 'b');
+
+    $query->innerJoin(
+      'content_moderation_state_field_data',
+      'md',
+      'md.content_entity_type_id = :etype AND md.content_entity_id = b.' . $id_key . ' AND md.content_entity_revision_id = b.' . $rev_key,
```
dstorozhuk: Separate issue on the same join: no langcode correlation and no `DISTINCT`. Both `node_field_data` and `content_moderation_state_field_data` have one row per langcode, so a translated trashed node fans out — a 7-language node produces 7x7 = 49 joined rows (verified against the 12 multilingual nodes in the current DB; none are trashed today, so this is latent). Once one is trashed: `--max=1000` burns 49 slots on a single node and `Found {count}` over-reports. Sharper edge: a node whose *non-default* translation is in trash still matches (no langcode correlation), and `$entity->delete()` removes the whole node including the live default translation.

Suggest adding `md.langcode = b.langcode` (or `md.default_langcode = 1 AND b.default_langcode = 1`) to the join condition.

docroot/modules/custom/trashbin/src/TrashbinPurgeCandidateQuery.php
```diff
@@ -0,0 +1,125 @@
...
+      [':etype' => $entity_type_id]
+    );
+
+    $revision_table = $definition->getRevisionTable();
+    $rt_timestamp = 'rt.revision_timestamp';
+    if ($entity_type_id === 'media') {
+      $rt_timestamp = 'rt.revision_created';
+    }
```
dstorozhuk, line 214: Hardcoding the media special case breaks other entity types. `entity_type` is a free-form CLI argument and the upfront validation in the command passes for e.g. `taxonomy_term` — which then dies with `SQLSTATE[42S22]: Unknown column 'rt.revision_timestamp'` (its column is also `revision_created`). `$definition->getRevisionMetadataKey('revision_created')` returns the right column for every type (node -> `revision_timestamp`; media/taxonomy_term/menu_link_content -> `revision_created`). Types where that key is NULL (paragraph, path_alias) and types without a `changed` field (`paragraphs_item_field_data` has none) should be rejected cleanly, the same way the base-table/key checks already do, instead of surfacing a raw PDO exception.

docroot/modules/custom/trashbin/src/TrashbinPurgeCandidateQuery.php
```diff
@@ -0,0 +1,125 @@
...
+
+    if ($revision_table) {
+      $query->innerJoin(
+        $revision_table,
+        'rt',
+        'rt.' . $rev_key . ' = b.' . $rev_key . ' AND rt.' . $id_key . ' = b.' . $id_key
+      );
+    }
```
dstorozhuk: `$rt_timestamp` references the `rt` alias unconditionally (lines 106-108, and line 120 below), but the join is only added when `$revision_table` is truthy. If this guard ever fires, the query is guaranteed invalid (`Unknown table 'rt'`) — a baffling SQL error instead of a clean refusal. Either treat a missing revision table as a structural error (`return NULL;`, consistent with the guards above), or drop the conditional.

docroot/modules/custom/trashbin/src/TrashbinPurgeCandidateQuery.php
```diff
@@ -0,0 +1,125 @@
...
+
+    $query->condition('md.moderation_state', 'trash');
+    $query->where('GREATEST(b.changed, ' . $rt_timestamp . ') < :cutoff', [':cutoff' => $cutoff]);
```
dstorozhuk: `node_revision.revision_timestamp` and `media_revision.revision_created` are both nullable. `GREATEST(x, NULL)` is `NULL`, and `NULL < :cutoff` is never true — so such rows are silently never purged. `COALESCE(rt.<ts>, 0)` (or `COALESCE(..., b.changed)`) makes it deterministic.

docroot/modules/custom/trashbin/src/Drush/Commands/TrashbinCommands.php
```diff
@@ -13,39 +14,86 @@ final class TrashbinCommands extends DrushCommands {
...
   #[CLI\Option(name: 'max', description: 'Maximum number of entities to delete.')]
   #[CLI\Option(name: 'days-ago', description: 'Number of days that the item must be unchanged in the trashbin.')]
   #[CLI\Usage(name: 'drush --simulate trashbin:purge node', description: 'Get a report of what would be purged.')]
   public function purge($entity_type, $options = ['max' => 1000, 'days-ago' => 180]) {
-    $maximum = strtotime($options['days-ago'] . ' days ago', $this->time->getCurrentTime());
+    // Capture command start time to avoid racing with edits during execution.
+    $startedAt = $this->time->getCurrentTime();
+    $maximum = strtotime($options['days-ago'] . ' days ago', $startedAt);
+
+    $cutoff = ((int) $options['days-ago'] > 0) ? $maximum : $startedAt;
```
dstorozhuk: **A malformed `--days-ago` silently escalates to "delete ALL trash".** `(int) 'abc'` -> 0, `(int) '-5'` -> -5, `(int) ''` -> 0 — all fall into the else branch, so `--days-ago=abc`, `--days-ago=-5` or an empty value silently behave like `--days-ago=0`, i.e. purge everything, with a log indistinguishable from a normal run. This is a regression vs the old code, where garbage input made `strtotime()` return `FALSE` and the query matched nothing.

The inverse trap also exists: `--days-ago=180abc` -> `(int)` is 180 > 0, but `strtotime('180abc days ago')` is `FALSE`, which coerces to cutoff 0 -> the command logs `Found 0 entities` and exits 0 forever — a broken cron job that looks healthy while trash accumulates.

Suggest validating before use:

```php
if (!ctype_digit((string) $options['days-ago'])) {
  throw new \InvalidArgumentException(sprintf('--days-ago must be a non-negative integer, got "%s".', $options['days-ago']));
}
```

That is loud *and* produces a non-zero exit code. Same treatment for `--max` (`(int) 'abc'` -> 0 -> silent no-op). Given the history of this command, also consider requiring an explicit flag (e.g. `--all`) for the delete-everything mode instead of overloading `0` as the fallback.

docroot/modules/custom/trashbin/src/Drush/Commands/TrashbinCommands.php
```diff
@@ -13,39 +14,86 @@ final class TrashbinCommands extends DrushCommands {
...
+      return;
+    }
+
+    $id_key = $definition->getKey('id');
+    $rev_key = $definition->getKey('revision');
+
+    if (!$id_key || !$rev_key) {
+      $this->logger()->error('Entity type {type} must be revisionable to use trashbin purge (missing id/revision keys).', ['type' => $entity_type]);
+      return;
+    }
```
dstorozhuk: These checks duplicate `buildCandidateQuery()`'s internals — `$base_table`/`$id_key`/`$rev_key` are unused after the guards (and `$definition` entirely), and the service returns a silent `[]`/`FALSE` for the same conditions. Two copies of the same logic will drift. Suggest making the service the single owner: throw a typed exception from `buildCandidateQuery()` for structural problems and drop this block. As written, a structural failure inside the service is indistinguishable from "entity was restored" (see the skip message below).

docroot/modules/custom/trashbin/src/Drush/Commands/TrashbinCommands.php
```diff
@@ -13,39 +14,86 @@ final class TrashbinCommands extends DrushCommands {
...
+
+    if (!$ids) {
+      return;
+    }
+
+    foreach ($ids as $id) {
+      $entity = $storage->load($id);
+      if (!$entity) {
+        continue;
+      }
```
dstorozhuk: This is the only branch in the loop with no log line — an entity that vanished between candidate selection and load disappears from the audit trail, so the upfront `Found {count}` can never be reconciled with the logged outcomes. One notice (`Skipping ID={id}; entity no longer exists.`) fixes it.

docroot/modules/custom/trashbin/src/Drush/Commands/TrashbinCommands.php
```diff
@@ -13,39 +14,86 @@ final class TrashbinCommands extends DrushCommands {
...
+
+      if (!$this->purgeCandidateQuery->isEntityEligible($entity_type, $id, $cutoff)) {
+        $this->logger()->notice('Skipping "{title}". ID={id}; no longer in trash or too recent.', [
+          'title' => $entity->label(),
+          'id' => $entity->id(),
+        ]);
+        continue;
+      }
```
dstorozhuk: Two observability notes: (a) `isEntityEligible()` also returns `FALSE` when the service hits a structural problem (its `buildCandidateQuery()` returns NULL), and then "no longer in trash or too recent" is misleading — an operator would chase a phantom restore; (b) there is no end-of-run summary — `Found {count}` is a pre-check upper bound logged before re-checks and skips, so distinguishing "healthy run, 400 deleted" from "crashed after 3" requires counting `Start delete` lines. A final `deleted X, skipped Y, failed Z` notice would make the cron log auditable.

docroot/modules/custom/trashbin/src/Drush/Commands/TrashbinCommands.php
```diff
@@ -13,39 +14,86 @@ final class TrashbinCommands extends DrushCommands {
...
+
       if (Drush::simulate()) {
-        $this->logger()->notice('Simulated delete of "{title}". ID={id}, {url}', ['title' => $entity->label(), 'id' => $entity->id(), 'url' => $entity->toUrl('canonical', ['absolute' => TRUE])->toString()]);
+        $this->logger()->notice('Simulated delete of "{title}". ID={id}, {url}', [
+          'title' => $entity->label(),
+          'id' => $entity->id(),
+          'url' => $entity->toUrl('canonical', ['absolute' => TRUE])->toString(),
```
dstorozhuk: Simulate-only crash risk: `toUrl('canonical')` throws `UndefinedLinkTemplateException` for entity types without a canonical link template, and URL generation can throw in general. node/media are fine, but the entity-type argument is free-form — and the failure mode is perverse: the cautious dry run crashes while the destructive real run works, which trains people to skip simulation. Guard with `hasLinkTemplate('canonical')` or a narrow try/catch (the URL is decorative in this log line).

docroot/modules/custom/trashbin/src/Drush/Commands/TrashbinCommands.php
```diff
@@ -13,39 +14,86 @@ final class TrashbinCommands extends DrushCommands {
...
+        ]);
       }
       else {
-        $this->logger()->notice('Start delete of "{title}". ID={id}', ['title' => $entity->label(), 'id' => $entity->id()]);
+        $this->logger()->notice('Start delete of "{title}". ID={id}', [
+          'title' => $entity->label(),
+          'id' => $entity->id(),
+        ]);
         $entity->delete();
```
dstorozhuk: One entity whose `delete()` throws — entity_usage cascade hooks, file-move failures (cf. DP-47987), search_api tracking, any `hook_entity_predelete` — aborts the entire remaining batch. And because candidates are ordered deterministically (oldest-first with a stable ID tiebreak), a poison entity lands at the same position every run: nothing behind it is ever purged again, and the nightly purge is permanently wedged with only a stack trace in an unread cron log as the symptom.

Suggest a per-entity try/catch: `logger()->error()` with entity type/ID and the exception message, `continue`, count failures, and exit non-zero if any delete failed. A post-delete confirmation line would also help — `Start delete` is logged *before* `delete()`, so after a crash the last log line names an entity in unknown state.

docroot/modules/custom/trashbin/tests/src/ExistingSite/TrashbinPurgeCandidateQueryTest.php
```diff
@@ -0,0 +1,178 @@
...
+use MassGov\Dtt\MassExistingSiteBase;
+
+/**
+ * Existing-site coverage for trash purge candidate selection.
+ *
+ * @coversDefaultClass \Drupal\trashbin\TrashbinPurgeCandidateQuery
+ *
+ * @group trashbin
+ */
+class TrashbinPurgeCandidateQueryTest extends MassExistingSiteBase {
```
dstorozhuk, line 17: Coverage gaps worth closing in this PR, in priority order:

1. **No regression test for the incident this ticket exists for** — the old hook matched a trashed CMS row of a *different* entity type whose `content_entity_revision_id` collided with the node's vid; the `:etype` join condition is the fix, and nothing pins it. Insert a synthetic `content_moderation_state_field_data` row (`content_entity_type_id = 'media'`, `content_entity_revision_id` = the node's vid, state `trash`) for a published node and assert `isEntityEligible('node', ...)` stays FALSE. That single test pays for the whole PR.
2. **Media is completely untested**, yet it is the dominant purge population (~98k trashed media rows vs ~28k nodes) *and* the one type with the hardcoded `revision_created` branch in the query service. Mirror `createTrashedOrgPageWithActivity()` for a media entity.
3. **The command itself is untested** (`--simulate`, `--max`, `--days-ago=0` vs `N`, actual deletion, the skip-on-restore path). Command-level tests are idiomatic in this repo — see `mass_redirect_normalizer/tests/src/ExistingSite/DrushCommandTest.php`. One blocker: the `Drush::simulate()` static throws when the Robo container isn't initialized (the DTT case); Drush's own guidance is to use `$this->getConfig()->simulate()` in commands, which also makes it injectable for tests.

docroot/modules/custom/trashbin/tests/src/ExistingSite/TrashbinPurgeCandidateQueryTest.php
```diff
@@ -0,0 +1,178 @@
...
+
+  /**
+   * Fetch enough rows to include our fixtures; the site DB has other trash.
+   *
+   * @see self::testCandidateIdsAreOrderedOldestActivityFirst()
+   */
+  private const PURGE_TEST_FETCH_MAX = 500000;
```
dstorozhuk: Minor: the comment explains the wrong mechanism. With cutoff 50000 the WHERE clause already excludes all real site trash (every real timestamp is modern — verified: the candidate pool below that cutoff is empty on a prod copy), so the isolation comes from the *cutoff*, not the big LIMIT. As a side effect two tests filesort the entire ~28k-row trash join. Worth rewording so the next author doesn't copy the wrong model.

docroot/modules/custom/trashbin/tests/src/ExistingSite/TrashbinPurgeCandidateQueryTest.php
```diff
@@ -0,0 +1,178 @@
...
+      ->fields(['changed' => $ts])
+      ->condition('nid', $nid)
+      ->execute();
+    $this->getConnection()->update('node_revision')
+      ->fields(['revision_timestamp' => $ts])
+      ->condition('vid', $vid)
+      ->execute();
+
+    $query = $this->createPurgeCandidateQuery();
+    $ids = $query->getCandidateIds('node', 50, time() + 60);
```
dstorozhuk: This test cannot fail under its own threat model. The assertion window is the **oldest 50** candidates, and the fixture's forced last-activity is only ~400 days old. Measured on a prod copy: the 50th-oldest candidate's `trash_last_activity` is from 2018, and 86,991 nodes of *any* moderation state are older than 400 days — so even if the trash filter were removed entirely (the exact bug class this test guards), the top-50 would still be filled with decade-old rows and `assertNotContains` would still pass.

Suggest asserting via `isEntityEligible('node', $nid, time() + 60)` instead — no LIMIT, no ordering, and since the ancient timestamps satisfy the cutoff clause, eligibility then hinges purely on the moderation filter, which is exactly what this test should pin.

conf/drupal/config/views.view.mass_dashboard_trash.yml
```diff
@@ -613,6 +613,69 @@ display:
           multi_type: separator
           separator: ', '
           field_api_classes: false
+        moderation_state:
+          id: moderation_state
+          table: node_field_data
+          field: moderation_state
+          relationship: vid
+          group_type: group
```
dstorozhuk: Two notes on this addition:

- The view's base table is `node_field_revision`, and the existing moderation_state *filter* on this display reads it directly with `relationship: none`. The new field instead goes through `node_field_data` via `relationship: vid`, which shows the **default revision's** state rather than the state of the revision in the row — and since the `vid` relationship is `required: false` (LEFT JOIN), the cell renders empty when they diverge. `table: node_field_revision` + `relationship: none` would be consistent with the filter and correct per-row.
- `style.options.columns` / `style.options.info` were not updated for the new field, so the next person who saves this view in the UI will produce an unrelated diff on this file.

Also worth confirming this dashboard change is intentional for DP-36234 — it isn't mentioned in the changelog.

changelogs/DP-36234.yml
```diff
@@ -0,0 +1,41 @@
...
+# Changed:
+#  - description: Automating the release branch.
+#    issue: DP-10166
+#  - description: Second change item that needs a description.
+#    issue: DP-19875
+#  - description: Third change item that needs a description along with an issue.
+#    issue: DP-19843
+#
+Fixed:
+  - description: Verify that Trash clear drush command is working correctly.
```
dstorozhuk: The description restates the ticket title (the task), not the change — compare e.g. DP-47987 ("Added check for whether file is a directory before attempting file move."). Suggest something like: "Trash purge now matches trashed items per entity type, deletes oldest first, and re-checks trash state before each delete."

dstorozhuk (commented):
Reviewed the purge rewrite against a fresh local copy of the prod DB.

**What's solid:** the original incident bug is genuinely fixed — the CMS join now constrains `content_entity_type_id`, and candidate coverage matches an independent latest-revision-is-trash count exactly (28,267 = 28,267). Oldest-first ordering with a stable tiebreak, the stable start-time cutoff with strict `<`, and the pre-delete re-check all match what was asked for in the ticket. The constructor `protected` -> `public` + `parent::__construct()` fix matches every other Drush command class in the repo and explains the "no commands defined in the trashbin namespace" error. phpcs is clean, and these are the first tests this module has ever had.

**Two merge-blockers though:**
1. Stale dual-workflow moderation rows make the command hard-delete **20 currently-draft nodes** (concrete nids inline) — same failure class as the original incident, different root cause.
2. A malformed `--days-ago` value silently escalates to the delete-all-trash mode — a typo away from the incident scenario, and a regression vs the old code where garbage input was a harmless no-op.

Also inline: a poison-entity failure that permanently wedges the cron purge, a published-nodes test that provably cannot fail, missing regression/media/command-level coverage, and several smaller items — each anchored to the file/line it concerns.

PR #3206, opened 2025-11-25
Release 0.459.0

arthurbaghdas:
### Changed
  - DP-43031: Allow Stacked Layout content type to be used as a parent for Info Details pages in the primary parent field.
  - DP-43523: BACKPORT undo changes that make parent page not required.


joeg8612:
no release this week

PR #3210, opened 2025-12-02
DP-42935: Pass data on referring pages to form pages

arthurbaghdas:
**Description:**

DP-****

**To Test:**

---


docroot/modules/custom/mass_session_context/js/iframe.js
```diff
@@ -0,0 +1,379 @@
...
+    if (!isFormsUrl(base)) {
+      return;
+    }
+
+    if (iframe.getAttribute(DATA_APPLIED_ATTR) === '1') {
+      return;
+    }
+
+    let url;
+    try {
```
tfleming-ma: @arthurbaghdas Having this try/catch both here and in `isFormsUrl` seems redundant, especially since `isFormsUrl` isn't used elsewhere. I would instead either move the contents out of that function altogether, or just inject the `url.hostname` value into the function and have it only check that the function argument is equal to `forms.mass.gov`

docroot/modules/custom/mass_session_context/mass_session_context.module
```diff
@@ -0,0 +1,55 @@
+<?php
+
+declare(strict_types=1);
+
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
+use Drupal\node\NodeInterface;
+use Drupal\mass_content\Entity\Bundle\node\FormPageBundle;
```
tfleming-ma:
```suggestion
```
It doesn't look like FormPageBundle is used

docroot/modules/custom/mass_session_context/js/iframe.js
```diff
@@ -0,0 +1,379 @@
...
+  const CONTEXT_PARAM_MAP = {
+    linking_page: ['current_page'],
+    previous_page: ['prior_page', 'previous_page'],
+    previous_page_2: ['prior_page_2', 'previous_page_2'],
+    linking_page_org: ['current_page_org'],
+    linking_page_parent_org: ['current_page_parent_org'],
+    previous_page_org: ['prior_page_org', 'previous_page_org'],
+    previous_page_2_org: ['prior_page_2_org', 'previous_page_2_org'],
+    previous_page_parent_org: ['prior_page_parent_org', 'previous_page_parent_org'],
+    previous_page_2_parent_org: ['prior_page_2_parent_org', 'previous_page_2_parent_org']
```
tfleming-ma: @arthurbaghdas What is the purpose of having alternative names for some of the parameters listed here?

docroot/modules/custom/mass_session_context/js/iframe.js
arthurbaghdas: Did some clean up here, yeah we dont need that and I went with just a simple inline check.

docroot/modules/custom/mass_session_context/mass_session_context.module
arthurbaghdas: Yeap

docroot/modules/custom/mass_session_context/js/iframe.js
arthurbaghdas: That was some old mapping, forgot to do the cleanup on this. all set now.

PR #3219, opened 2025-12-12
DP-43938: author report for editors

joeg8612:
**Description:**

DP-43938

**To Test:**

---


dstorozhuk (changes requested):
1. Since the `User: User status (= Yes)` filter was added, and it is not exposed, and set to Yes - there is no reason to display the  Status column. 
2. I added a permission `View author peers`. I suggest using it in views instead of direct roles selection (added a separate commit). 
I can make those changes, if you like - just assign that task to me.

dstorozhuk (approved):

dstorozhuk (approved):

dstorozhuk (approved):

PR #3243, opened 2026-01-26
DP-41803: A11Y datatable uses duplicate table tag

arthurbaghdas:
**Description:**

DP-****

**To Test:**

---


patches/csv-field-wrapper-class-ma__csvtable.patch
```diff
@@ -1,31 +0,0 @@
-diff --git a/js/csv-field.js b/js/csv-field.js
```
clairesunstudio, line 1: I don't think this patch is needed anymore right? Do you want to do a module release to bring that change in?

PR #3250, opened 2026-02-02
DP-44741: Drupal update

arthurbaghdas:
**Description:**

DP-****

**To Test:**

---


dstorozhuk (changes requested):
So far I see the one moderate issue.  iframe resizer added twice in a new composer.lock. 
In a new composer.lock additonal version installed 4.4.5, while the old version also remain in composr.lock. And they are trying to install togeather iframe-resizer. Hovewer 4.3.6 is the one that always installed in my tests. I think we need to remove one or another. But if we decide to keep 4.4.5 - we also need to update the library in Forms project and keep release it togeather. Or, we can remove the 4.4.5 and create follow up ticket.

dstorozhuk:
@arthurbaghdas  the conflicts need to be resolved.

patches/DP-11721-jsonapi-serializer.patch
```diff
@@ -1,13 +1,13 @@
-diff --git a/modules/jsonapi/src/Serializer/Serializer.php b/modules/jsonapi/src/Serializer/Serializer.php
```
dstorozhuk, line 1: Better to rename the DP-11721-jsonapi-serializer.patch, otherwise it can be ignored if patch already applied.

dstorozhuk (changes requested):

dstorozhuk:
@arthurbaghdas  if we remove `mayflower_preprocess_office_hours_status()`It will trigger some error or what? Do we really need that empty function?

dstorozhuk:
changelogs/DP-44741.yml
>  - description: Upgrade to Drupal 10.6.x and PHP to 8.4.

I am little confused here. Are we preparing the code for PHP to 8.4, or we actually planning to switch to 8.4 after the release? 
If we wand to switch - need to change .ddev, tugboat, circleci configs. If we just want to prepare the code - it is better to be more explicit here, like: "Upgrade to Drupal 10.6.x and prepare the code base to be ready for PHP 8.4".

dstorozhuk:
@arthurbaghdas  there are also some code styles issue, should be easy to fix.

docroot/modules/custom/mass_map/templates/map-row.html.twig
```diff
@@ -6,36 +6,36 @@
...
-    'sub': '',
-    'color': '',
-    'id': '',
-    'centered': ''
+  compHeading: {
+    title: 'Locations',
+    sub: '',
+    color: '',
+    id: '',
+    centered: '',
```
tfleming-ma, line 14: @arthurbaghdas Just to confirm, are commas now expected at the end of a list like this in Twig?

docroot/modules/custom/mayflower/mayflower.module
```diff
@@ -822,3 +822,12 @@ function mayflower_preprocess_paragraph__collection_search(&$variables) {
...
   }
 }
+
+/**
+ * Implements hook_preprocess_HOOK().
+ *
+ * Leave this function empty we don't use this feature.
+ */
+function mayflower_preprocess_office_hours_status(&$variables): void {
+
```
tfleming-ma: @arthurbaghdas Any reason to have this included instead of removing the feature? Are we using other features of the office_hours contrib module?

docroot/modules/custom/mass_entity_usage/src/MassEntityUsage.php
```diff
@@ -46,7 +46,7 @@ public function listSourcesPage(EntityInterface $target_entity, $offset, $nest_r
     ]);
 
     // Set a range and restrict usage records to unique sources.
-    $items_per_page = $this->config->get('usage_controller_items_per_page') ?: LocalTaskUsageController::ITEMS_PER_PAGE_DEFAULT;
+    $items_per_page = Controller\ListUsageController::ITEMS_PER_PAGE_DEFAULT;
```
tfleming-ma: @arthurbaghdas Does this change mean that the entity usage module no longer has this configuration or that we are just wanting to always use the default value?

docroot/modules/custom/mass_bulk_file_replace/src/Form/ReplaceUploadForm.php
```diff
@@ -279,7 +279,7 @@ public static function processFileBatch($item, $uid, $username, $accessibility_s
       $destination_uri = rtrim($destination, '/') . '/' . $clean_name;
 
       // Copy uploaded file into final destination, overwriting any existing file.
-      $copied = $file_system->copy($path, $destination_uri, FileSystemInterface::EXISTS_REPLACE);
+      $copied = $file_system->copy($path, $destination_uri);
```
tfleming-ma:
```suggestion
      $copied = $file_system->copy($path, $destination_uri, FileExists::Replace);
```
Looking at FileSystemInterface, it looks like with `FileSystemInterface::EXISTS_REPLACE` removed, the copy method will fall back to `FileExists::Rename`, since the copy method sets `FileExists::Rename` as the default value

.circleci/config.yml
```diff
@@ -308,20 +311,19 @@ jobs:
...
             and:
               - equal: [ phpunit, << parameters.command >> ]
           steps:
-#            run: 'su -s /bin/bash root -c "phpunit docroot/modules/custom --testdox --log-junit /tmp/test-results/dtt/junit.xml"'
             - run:
                 # https://circleci.com/docs/use-the-circleci-cli-to-split-tests/#configure-a-job-running-phpunit-testshttps://circleci.com/docs/use-the-circleci-cli-to-split-tests/#configure-a-job-running-phpunit-tests
                 name: Run functional tests
                 command: |
-                  TESTS_TO_RUN=$(phpunit-finder --config-file /var/www/code/phpunit.xml.dist --bootstrap-file /var/www/code/vendor/weitzman/drupal-test-traits/src/bootstrap.php)
+                  TESTS_TO_RUN=$(phpunit -c /var/www/code/phpunit.xml.dist --list-test-files --colors=never | sed -n 's/^ - //p')
```
tfleming-ma, line 326: @arthurbaghdas See second bulleted comment in Jira

docroot/modules/custom/mass_content/src/Drush/Commands/MassContentCommands.php
```diff
@@ -8,6 +8,7 @@
 use Drupal\Core\Datetime\DrupalDateTime;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Field\FieldItemListInterface;
+use Drupal\Core\Entity\RevisionableStorageInterface;
```
tfleming-ma:
```suggestion
```
This appears to be an unused interface

docroot/themes/custom/mass_admin_theme/js/entity-embed-dialog-alter.js
```diff
@@ -5,10 +5,119 @@
  * This also supports collapsible navigable is the 'is-collapsible' class is
  * added to the main element, and a target element is included.
  */
-(function ($, Drupal) {
+(function ($, Drupal, drupalSettings) {
 
   'use strict';
 
+  function getAjaxInstance(button) {
```
tfleming-ma, line 19: @arthurbaghdas I'm wondering what the reason is for the added JS here. Is this because of another library being removed from core?

arthurbaghdas:
> So far I see the one moderate issue. iframe resizer added twice in a new composer.lock. In a new composer.lock additonal version installed 4.4.5, while the old version also remain in composr.lock. And they are trying to install togeather iframe-resizer. Hovewer 4.3.6 is the one that always installed in my tests. I think we need to remove one or another. But if we decide to keep 4.4.5 - we also need to update the library in Forms project and keep release it togeather. Or, we can remove the 4.4.5 and create follow up ticket.

Fixed

arthurbaghdas:
> @arthurbaghdas the conflicts need to be resolved.

Fixed

patches/DP-11721-jsonapi-serializer.patch
arthurbaghdas: Fixed

arthurbaghdas:
> @arthurbaghdas if we remove `mayflower_preprocess_office_hours_status()`It will trigger some error or what? Do we really need that empty function?

no need, removed it.

arthurbaghdas:
> changelogs/[DP-44741](https://massgov.atlassian.net/browse/DP-44741).yml
> 
> > * description: Upgrade to Drupal 10.6.x and PHP to 8.4.
> 
> I am little confused here. Are we preparing the code for PHP to 8.4, or we actually planning to switch to 8.4 after the release? If we wand to switch - need to change .ddev, tugboat, circleci configs. If we just want to prepare the code - it is better to be more explicit here, like: "Upgrade to Drupal 10.6.x and prepare the code base to be ready for PHP 8.4".

No plans for php 8.4, there is still one module that is not ready for that switch, we will do the upgrade later.
Changelog updated accordingly.

[DP-44741]: https://massgov.atlassian.net/browse/DP-44741?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

arthurbaghdas:
> @arthurbaghdas there are also some code styles issue, should be easy to fix.

Fixed

docroot/modules/custom/mass_map/templates/map-row.html.twig
arthurbaghdas: Yes — trailing commas are now enforced by Twig CS rules for multi-line arrays.

docroot/modules/custom/mayflower/mayflower.module
arthurbaghdas: Removed the code, the config covers this.

docroot/modules/custom/mass_entity_usage/src/MassEntityUsage.php
arthurbaghdas: I updated the subquery paging to use usage_controller_items_per_page with default fallback

docroot/modules/custom/mass_bulk_file_replace/src/Form/ReplaceUploadForm.php
arthurbaghdas: You are totally right.

docroot/modules/custom/mass_content/src/Drush/Commands/MassContentCommands.php
arthurbaghdas: There were more. removed those.

docroot/themes/custom/mass_admin_theme/js/entity-embed-dialog-alter.js
arthurbaghdas: It’s not due to a core library removal. This JS is a Drupal 11 workaround for entity-embed dialog actions sometimes losing Ajax bindings after dialog rebuilds, causing full-page submit/redirect instead of modal flow. We force Next/Embed/Back through Drupal Ajax consistently. It’s covered by our EntityEmbedDialogAjaxBridgeTest.

.circleci/config.yml
arthurbaghdas: Fixed

dstorozhuk (approved):

PR #3255, opened 2026-02-05
DP-44915: Editoria11y 3, node and media a11y views showing content and documents by top-level org

tfleming-ma:
**Description:**
- Updates Editoria11y to 3.x and enables submodules
- Updates a11y author report to be compatible with Editoria11y 3.x, removes custom code and patches
- Adds two MySQL views: one showing top-level orgs for nodes, another for media entities
- Adds two Drupal views:  one showing top-level orgs for nodes, another for documents. These views show data, such as # of pages and data related to Editoria11y alerts and dismissals
- Manually removed legacy Editoria11y 2.x results view

DP-44915

**To Test:**

---


docroot/modules/custom/mass_views/src/Plugin/views/join/Ed11yPidCountJoin.php
```diff
@@ -0,0 +1,35 @@
+<?php
```
tfleming-ma, line 1: Implementation note: this code is based on the Subquery join plugin in Drupal core. I was not able to use the Subquery plugin in hook_views_data due to the issues described here: https://www.drupal.org/project/drupal/issues/3125146

docroot/modules/custom/mass_views/mass_views.install
```diff
@@ -0,0 +1,119 @@
...
+ * @file
+ * Install/update hooks for the mass_views module.
+ */
+
+use Drupal\Core\Database\Database;
+
+/**
+ * Create/replace the MySQL 8 recursive CTE views used for top-org relationships.
+ */
+function mass_views_update_10001(): void {
```
dstorozhuk, line 29: hook_install() is missing. Not critical for the project, but I think it worth to be added, as well as hook_uninstall. This is common for all DB manipulations in a modules. 

It also important for developers, because some developers (like me) can look to the DB schema which can be defined by hook_install, hook_schema, rather then update hooks to understand the DB state.

docroot/modules/custom/mass_views/src/Plugin/views/join/Ed11yPidCountJoin.php
```diff
@@ -0,0 +1,41 @@
...
+class Ed11yPidCountJoin extends JoinPluginBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildJoin($select_query, $table, $view_query) {
+    $pseudoTableAlias = $this->table . '_' . $this->leftTable;
+
+    /** @var \Drupal\mysql\Driver\Database\mysql\Select $subQuery */
+    $subQuery = \Drupal::database()->select($this->table, $pseudoTableAlias);
```
dstorozhuk: Minor:  @tfleming-ma  can you use DI here?

docroot/modules/custom/mass_views/src/Plugin/views/field/OrgTopParentField.php
```diff
@@ -0,0 +1,150 @@
...
+ *
+ * For org_page rows: the org is the node itself.
+ * For other node rows: org(s) come from field_organizations.
+ * The "top-level" org is found by walking org_page.field_parent upward.
+ *
+ * @ingroup views_field_handlers
+ *
+ * @ViewsField("mass_views_org_top_parent_field")
+ */
+class OrgTopParentField extends FieldPluginBase {
```
dstorozhuk: @tfleming-ma  how this plugin is used ? I don't see mention of mass_views_org_top_parent_field . Same for OrgTopParentMediaField. Do we need that code?

dstorozhuk (changes requested):
view https://mass.local/admin/structure/views/view/editoria11y_results  become  broken  after deleting `editoria11y_results`

tfleming-ma:
> view https://mass.local/admin/structure/views/view/editoria11y_results become broken after deleting `editoria11y_results`

Deleted. This is a view from Editoria11y 2.x that has been replaced. It's supposed to be deleted, but because we had our own custom version of the config saved, it persisted.

docroot/modules/custom/mass_views/mass_views.install
tfleming-ma: Added a hook_install and hook_uninstall. Locally, you can safely uninstall and re-enable mass_views to test. When uninstalling, it will also uninstall mass_admin_toolbar.

docroot/modules/custom/mass_views/src/Plugin/views/join/Ed11yPidCountJoin.php
tfleming-ma: Done

docroot/modules/custom/mass_views/src/Plugin/views/field/OrgTopParentField.php
tfleming-ma: This is a leftover artifact from iterating on these views. We are no longer using these fields. So, removed.

docroot/modules/custom/mass_views/mass_views.install
```diff
@@ -0,0 +1,154 @@
...
+ */
+function mass_views_install(): void {
+  create_or_replace_top_org_views();
+}
+
+/**
+ * Implements hook_uninstall().
+ */
+function mass_views_uninstall(): void {
+  delete_top_org_views();
```
dstorozhuk: @tfleming-ma  please rename to  created/delete... functions to `_module_name_FUNCTION_NAME`. it possible if other AI agent will also not follow the Drupal coding standards and generate the same name, so we can have Cannot redeclare function.. error.

dstorozhuk (changes requested):

docroot/modules/custom/mass_views/mass_views.install
tfleming-ma: Ah good catch. That was a miss on my part. Fixed

PR #3295, opened 2026-03-25
DP-45698: Media download controller changes

arthurbaghdas:
**Description:**

DP-****

**To Test:**

---


docroot/modules/custom/mass_media/src/Controller/MassMediaDownloadController.php
```diff
@@ -101,36 +134,100 @@ public function download(MediaInterface $media) {
...
+        $originHost = (string) parse_url($origin, PHP_URL_HOST);
+        $requestHost = $this->requestStack->getCurrentRequest()->getHost();
+
+        // Only fetch when we are not on mass.gov production host.
+        if (!empty($originHost) && strcasecmp($requestHost, $originHost) !== 0) {
+          $originDir = trim((string) ($stageConfig->get('origin_dir') ?? 'files'));
+          $relativePath = str_replace('public://', '', $file_uri);
+          $options = ['verify' => (bool) $stageConfig->get('verify')];
+
+          $this->stageFileProxyDownloadManager->fetch(
```
dstorozhuk: **Critical — wrap `fetch()`: uncaught non-Guzzle errors become a 500 instead of a clean 404.**

`DownloadManager::fetch()` only catches `GuzzleException`. Other failures originating here — the unbounded lock-acquire spin loop, and filesystem writes (disk full / permissions / ENOSPC) — raise non-Guzzle `\Throwable` that propagate out of the controller as an uncaught **500** on a download URL whose correct failure mode is a 404. It is also synchronous in the request path, so a slow/contended origin hangs the user's download. The `bool` return is discarded as well, so a failed fetch leaves no signal on the `mass_media` log channel.

```php
try {
  $fetched = $this->stageFileProxyDownloadManager->fetch($origin, $originDir, $relativePath, $options);
  if (!$fetched) {
    $this->getLogger('mass_media')->warning('Stage File Proxy could not fetch @path from @origin.', ['@path' => $relativePath, '@origin' => $origin]);
  }
}
catch (\Throwable $e) {
  $this->getLogger('mass_media')->error('Stage File Proxy fetch error for @path: @msg', ['@path' => $relativePath, '@msg' => $e->getMessage()]);
}
```

docroot/modules/custom/mass_media/src/Controller/MassMediaDownloadController.php
```diff
@@ -101,36 +134,100 @@ public function download(MediaInterface $media) {
...
 
-    // Returns a 301 Moved Permanently redirect response.
-    $response = new TrustedRedirectResponse($uri, 301);
-    // Adds cache metadata.
-    $response->getCacheableMetadata()->addCacheContexts(['url.site']);
-    $response->addCacheableDependency($media);
-    $response->addCacheableDependency($file);
-    if (!$context->isEmpty()) {
-      $response->addCacheableDependency($context->pop());
+    $response = new BinaryFileResponse($file_uri, Response::HTTP_OK, $headers, $scheme !== 'private');
```
dstorozhuk: **Important — public responses lose `max-age`, so the CDN will not cache them; every download streams through PHP-FPM.**

`$public = ($scheme !== 'private')` makes Symfony call `setPublic()`, overriding the `Cache-Control: private` from `File::getDownloadHeaders()`. Because an explicit `Cache-Control` is now set, `FinishResponseSubscriber` will not add a default `max-age`, so the response is `Cache-Control: public` with **no freshness lifetime**. Confirmed by comparing the same document before/after: the old 301 carried `max-age=604800, public`; this response has `public` and no `max-age`, which Akamai (default config) will not cache.

Combined with moving from a 301 (served by the edge/web server) to streaming every byte through PHP-FPM with no `X-Sendfile`/`X-Accel-Redirect`, this is a real load shift on one of the highest-traffic paths on the site. Per the performance gate (docs/performance.md) it needs an explicit caching strategy + load assessment before merge.

Suggest setting an explicit `max-age`/`s-maxage` (+ `Last-Modified`) for public files and enabling X-Sendfile so large PDFs are not streamed through FPM memory.

dstorozhuk: **Important — no cache metadata tied to the media/file → stale bytes can be served after a file swap.**

The old `TrustedRedirectResponse` carried `addCacheableDependency($media)` / `addCacheableDependency($file)`, so cache tags invalidated the response when content changed. This `BinaryFileResponse` carries none. At the same URL (`/media/{id}/download`) with `public` + ETag, an intermediary/browser that caches it can serve old bytes after the file is replaced. `testMediaDownloadServesUpdatedFileAfterReplacement` only proves Drupal serves fresh bytes per request, not the CDN. (Today this is partly masked by the missing `max-age` above; fixing that makes this live.)

Suggest deriving `Last-Modified`/`ETag` from the file and/or emitting cache-tag surrogate keys Akamai honors, plus a test asserting the validators change after replacement.

docroot/modules/custom/mass_media/src/Controller/MassMediaDownloadController.php
```diff
@@ -101,36 +134,100 @@ public function download(MediaInterface $media) {
...
     }
+
     return $response;
   }
 
+  /**
+   * Determines whether a file exists at the given stream-wrapper URI.
+   */
+  private function fileExists(string $file_uri): bool {
+    $realpath = $this->fileSystem->realpath($file_uri);
```
dstorozhuk: **Important — `fileExists()` diverges from core's `is_file()` and can mis-detect existence or swallow permission errors.**

Core uses `is_file($uri)` on the stream URI. Here, for wrappers where `realpath()` returns `FALSE` (many remote/custom wrappers, and `private://` in some configs), the fallback `file_exists($file_uri)` can report a present-but-unresolvable file as missing (spurious 404) or, on a stale directory entry, a missing file as present. `file_exists()`/`realpath()` also return `FALSE` for 'exists but unreadable', collapsing a permissions problem into a silent 404. Suggest matching core's `is_file()` semantics and pairing the final 404 (line 167) with a debug log.

dstorozhuk (changes requested):

docroot/modules/custom/mass_media/src/Controller/MassMediaDownloadController.php
arthurbaghdas: X-Sendfile requires Acquia/web server configuration (mod_xsendfile or X-Accel-Redirect) — nothing in the codebase enables it today. Worth a follow-up with infra if load testing shows PHP-FPM pressure on document paths.

docroot/modules/custom/mass_media/src/Controller/MassMediaDownloadController.php
arthurbaghdas: Done

docroot/modules/custom/mass_media/src/Controller/MassMediaDownloadController.php
arthurbaghdas: Done

docroot/modules/custom/mass_media/src/Controller/MassMediaDownloadController.php
arthurbaghdas: Done

dstorozhuk (approved):

dstorozhuk (approved):

PR #3305, opened 2026-04-01
DP-45831: Update intenal links on mass.gov that are redirects

arthurbaghdas:
**Description:**

DP-****

**To Test:**

---


changelogs/DP-45831.yml
```diff
@@ -0,0 +1,41 @@
...
+# Changed:
+#  - description: Automating the release branch.
+#    issue: DP-10166
+#  - description: Second change item that needs a description.
+#    issue: DP-19875
+#  - description: Third change item that needs a description along with an issue.
+#    issue: DP-19843
+#
+Changed:
+  - description: Update intenal links on mass.gov that are redirects.
```
tfleming-ma:
```suggestion
  - description: Update internal links on mass.gov that are redirects.
```

docroot/modules/custom/mass_redirect_normalizer/tests/src/ExistingSite/RedirectLinkNormalizationTest.php
```diff
@@ -0,0 +1,2252 @@
+<?php
```
tfleming-ma: This was difficult to review. I think it would have made sense to split into multiple test files, especially between redirect actions, the command, and views.

docroot/modules/custom/mass_redirect_normalizer/mass_redirect_normalizer.module
```diff
@@ -0,0 +1,23 @@
+<?php
+
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\node\NodeInterface;
+use Drupal\paragraphs\ParagraphInterface;
+
+/**
+ * Implements hook_entity_presave().
+ */
+function mass_redirect_normalizer_entity_presave(EntityInterface $entity) {
```
tfleming-ma: Since we're on Drupal 11.3, it would be good to use the new Hooks system. I don't think changing this implementation to Hooks needs to block merging though.

docroot/modules/custom/mass_redirect_normalizer/tests/src/ExistingSite/RedirectLinkNormalizationTest.php
arthurbaghdas: Makes sense, I've split the tests.

docroot/modules/custom/mass_redirect_normalizer/mass_redirect_normalizer.module
arthurbaghdas: Changed.

docroot/modules/custom/mass_redirect_normalizer/src/RedirectLinkResolver.php
```diff
@@ -0,0 +1,466 @@
...
+
+      $next = $this->extractLocalPath($redirect->getRedirectUrl()->toString());
+      if (!$next) {
+        break;
+      }
+      $current = ltrim($next, '/');
+    }
+
+    $finalPath = '/' . ltrim($current, '/');
+    $targetPath = $finalPath . $query . $fragment;
```
dstorozhuk: **Critical — the redirect target's own query string is dropped.**

`resolveRedirectTarget()` captures query/fragment only from the *source* URL (L178-179). Each hop goes through `extractLocalPath()`, which keeps only `PHP_URL_PATH`, so a query string stored in the redirect's destination is discarded:

```
redirect: /old-page → internal:/search?category=health
link to /old-page in content → rewritten to /search   (?category=health is lost)
```

Prod currently has ~185 redirects with a `?` in the target URI — every internal link pointing at one of those sources gets rewritten to the wrong (unfiltered) target.

Suggested fix: keep each hop destination's own query/fragment; append the source query/fragment only when the destination has none (matching how a real 301 chain behaves).

docroot/modules/custom/mass_redirect_normalizer/src/RedirectLinkResolver.php
```diff
@@ -0,0 +1,466 @@
...
+    }
+
+    $query = (string) parse_url((string) $resolved['target_path'], PHP_URL_QUERY);
+    $fragment = (string) parse_url((string) $resolved['target_path'], PHP_URL_FRAGMENT);
+    if (!empty($resolved['entity']) && $query === '' && $fragment === '') {
+      return [
+        'changed' => TRUE,
+        'uri' => 'entity:' . $resolved['entity']->getEntityTypeId() . '/' . $resolved['entity']->id(),
+      ];
+    }
```
dstorozhuk, line 96: **Critical — link fields: `/media/N/download` becomes `entity:media/N`, `/download` is silently dropped.**

When a redirect chain ends at `/media/123/download`, `resolvePathToEntity()` (the regex at L317 allows the `/download` suffix) resolves media 123, and since there's no query/fragment this returns `entity:media/123`. That renders to the media **canonical page**, while `/download` is a separate route (`MassMediaDownloadController` → 301 to the file). So a "download document" link silently becomes a "media page" link. Note that `MassUrlReplacementService` in mass_fields deliberately preserves `/download`. The rich-text path is unaffected (href keeps `target_path` including `/download`) — only link fields hit this.

Suggested fix: emit `entity:` only when `target_path` equals the entity's canonical path; when the final path ends in `/download`, emit `internal:/media/N/download`.

docroot/modules/custom/mass_redirect_normalizer/src/RedirectLinkNormalizationManager.php
```diff
@@ -0,0 +1,325 @@
...
+   * After the paragraph is saved, parent entities must reference the new
+   * paragraph revision. Saving the host node with an in-memory copy of the
+   * pre-normalization paragraph would create another paragraph revision that
+   * reverts the link fix.
+   */
+  private function saveNormalizedParagraphAndAncestors(Paragraph $paragraph): void {
+    $paragraphStorage = $this->entityTypeManager->getStorage('paragraph');
+
+    $this->prepareRevision($paragraph, self::REVISION_MESSAGE);
+    $paragraph->save();
```
dstorozhuk, line 153: **Critical — the paragraph → ancestors → host node save chain is not transactional.**

Sequential independent `save()` calls with no surrounding transaction. An exception mid-chain (DB deadlock, validation, another module's presave) leaves the child paragraph saved with the rewritten link while the host node still references the old revision — and the change log only gets a FAILED row from the worker's catch, telling the operator nothing was applied when half of it actually was.

Suggested fix: wrap the chain in `$this->database->startTransaction()`, or build all mutations in memory and persist them through a single host-node save.

dstorozhuk (commented):
Reviewed the full module with focus on content-integrity risks. Critical findings and test-coverage gaps are posted as inline comments on the relevant lines.

docroot/modules/custom/mass_redirect_normalizer/src/RedirectLinkNormalizationManager.php
```diff
@@ -0,0 +1,325 @@
...
+        return;
+      }
+      $this->replaceParagraphReference($parentParagraph, $freshParagraph);
+      $this->prepareRevision($parentParagraph, self::REVISION_MESSAGE);
+      $parentParagraph->save();
+
+      $freshParagraph = $paragraphStorage->loadRevision((int) $parentParagraph->getRevisionId());
+      if (!$freshParagraph instanceof Paragraph) {
+        return;
+      }
```
dstorozhuk, line 188: **Critical — silent `return;`s here leave the fix invisible on the live page while the report logs success.**

The paragraph is saved first (new revision already persisted), then this method can bail out silently when `loadRevision()`/`load()` of a parent fails. The return value of `replaceParagraphReference()` is also ignored (L149, L163): if the walk doesn't find the reference, the host entity still gets a new revision pointing at the *old* `target_revision_id`. In all of these cases `normalizeEntity()` returns `changed=TRUE` and the worker writes a **succeeded** change-log row (`RedirectLinkNormalizationQueueWorker.php:118-119`).

Result: the normalized paragraph revision is orphaned, the live page keeps the old link, and the report says it was fixed — undetectable via the report.

Suggested fix: throw on these paths (so the worker records a failed row) or propagate a partial status into the change log. A child save whose host update didn't complete must never be logged as `succeeded`.

docroot/modules/custom/mass_redirect_normalizer/src/RedirectLinkChangeLog.php
```diff
@@ -0,0 +1,100 @@
...
+    string $entityType,
+    int $entityId,
+    string $bundle,
+    string $source,
+    string $message,
+  ): void {
+    if (!$this->database->schema()->tableExists(self::TABLE)) {
+      return;
+    }
+    $this->database->insert(self::TABLE)
```
dstorozhuk: **Critical — a failing change-log INSERT aborts the whole 100-entity batch.**

`logFailure()` is called *inside* the worker's catch block (`RedirectLinkNormalizationQueueWorker.php:83-89`). If this INSERT throws (deadlock, lost connection, oversized value — `before_value`/`after_value` are `text` columns capped at 64KB and store the full field HTML), the exception escapes `processItem()` and kills the remaining entities of the batch; already-saved changes stay unlogged. Same applies to `logChanges()` on the success path — the entity save has already committed by the time it runs.

Suggested fix: wrap both inserts in their own try/catch + watchdog. Logging must never be able to crash the worker or hide committed changes.

docroot/modules/custom/mass_redirect_normalizer/src/RedirectLinkNormalizationManager.php
```diff
@@ -0,0 +1,325 @@
...
+    if ($entity instanceof RevisionableInterface) {
+      $entity->setNewRevision();
+    }
+    if ($entity instanceof RevisionLogInterface) {
+      // These automated fixes should show as admin.
+      $entity->setRevisionUserId(1);
+      $entity->setRevisionLogMessage($message);
+      $entity->setRevisionCreationTime($this->time->getRequestTime());
+    }
+  }
```
dstorozhuk, line 383: **`prepareRevision()` doesn't preserve `changed` and doesn't set `setSyncing(TRUE)`.**

Applies to every save path: L85 (node), L135 (paragraph), L150 (ancestor paragraphs), L164 (host node). Every normalized node/paragraph gets `changed = now`. At sweep scale that's a sitewide bump of "last updated" — and `mass_content`'s `EntitySorter` falls back to `changed` for content without an explicit date field, so listing order can shift across the site as the sweep runs. The existing `MassUrlReplacementService` runs inside normal editor-initiated saves so it never had this problem; an automated mass re-save should preserve the original `changed` timestamp (and consider `setSyncing(TRUE)`).

docroot/modules/custom/mass_redirect_normalizer/src/RedirectLinkResolver.php
```diff
@@ -0,0 +1,466 @@
...
+
+    $storage = $this->entityTypeManager->getStorage('redirect');
+    foreach ($candidates as $candidate) {
+      $query = $storage->getQuery()
+        ->accessCheck(FALSE)
+        ->range(0, 1);
+      $group = $query->orConditionGroup()
+        ->condition('redirect_source.path', $candidate)
+        ->condition('redirect_source__path', $candidate);
+      $ids = $query->condition($group)->execute();
```
dstorozhuk: **Non-deterministic redirect pick when one source path has multiple redirect rows.**

This uses `->range(0, 1)` on a path-only match: no filter on `redirect_source__query`, no handling of multiplicity — it rewrites using whatever row the DB returns first. Prod currently has ~136 source paths with more than one redirect row, so for those the chosen target is effectively arbitrary (and a query-specific redirect can be applied to a query-less link). The entity-reference path already treats >1 match as ambiguous and bails (`loadRedirectsBySourcePath()` → `ambiguous_redirects`) — this path should apply the same strictness.

docroot/modules/custom/mass_redirect_normalizer/tests/src/ExistingSite/QueueWorkerTest.php
```diff
@@ -0,0 +1,293 @@
...
+
+/**
+ * Tests queue worker processing, presave hook integration, and change log writes.
+ *
+ * Covers: presave enqueue, worker normalization, env re-entry guard,
+ * change log table writes for successes and failures.
+ *
+ * @group existing-site
+ */
+class QueueWorkerTest extends MassExistingSiteBase {
```
dstorozhuk, line 18: **Test gap — the bulk `['entities' => [...]]` queue payload, the only shape the production sweep creates (batches of 100 via `enqueueIdBulk()` / `flushEnqueueBuffers()`), is never processed by any test.**

Every queue test here feeds the single-entity `['entity_type','entity_id']` shape, so `expandQueuePayload()`'s bulk branch and `loadEntitiesForPairs()` have zero coverage. A regression there means the production sweep silently normalizes nothing while tests stay green.

Suggested: enqueue ≥2 entities via the bulk path, claim the single queue item, run `processItem()`, assert both entities are normalized and both change-log rows are written.

docroot/modules/custom/mass_redirect_normalizer/tests/src/ExistingSite/QueueWorkerTest.php
```diff
@@ -0,0 +1,293 @@
...
+      ->range(0, 1)
+      ->execute()
+      ->fetchField();
+    $this->assertSame('succeeded', $status);
+  }
+
+  /**
+   * Tests change log service logs succeeded and failed rows with status field.
+   */
+  public function testChangeLogStatusColumnForSuccessAndFailure(): void {
```
dstorozhuk, line 528: **Test gap — mid-batch failure isolation is unasserted.**

The per-entity try/catch in `processItem()` is the design guarantee that one bad entity doesn't take out the other 99 in its batch, but this test calls `logFailure()` directly, bypassing the worker's catch. No test drives an entity that throws during processing and asserts that (a) a `failed` row is written and (b) the remaining entities in the same batched item still get processed.

docroot/modules/custom/mass_redirect_normalizer/src/RedirectLinkNormalizationEligibility.php
```diff
@@ -0,0 +1,92 @@
...
+      return !$this->hasNewerUnpublishedDraft($parentNode, $newerDraftCache);
+    }
+
+    return FALSE;
+  }
+
+  /**
+   * Returns TRUE when latest node revision is unpublished and newer.
+   */
+  private function hasNewerUnpublishedDraft(NodeInterface $node, array &$cache): bool {
```
dstorozhuk, line 67: **Test gap — the forward-draft skip has zero tests.**

This is the invariant protecting editors' unpublished drafts from being clobbered by the sweep; nothing in `tests/` references `Eligibility` / `isEligible` at all.

Suggested: published node + a newer unpublished draft containing a redirect link → assert `isEligible()` is FALSE and that draining the queue leaves both revisions untouched.

docroot/modules/custom/mass_redirect_normalizer/tests/src/ExistingSite/DrushCommandTest.php
```diff
@@ -0,0 +1,153 @@
...
+ */
+class DrushCommandTest extends MassExistingSiteBase {
+
+  use RedirectNormalizerTestTrait;
+
+  /**
+   * Tests command bundle filter constrains output.
+   */
+  public function testCommandBundleFiltering(): void {
+    $this->markTestSkipped('Removed by enqueue-only command refactor.');
```
dstorozhuk: **Test gap — most of this file is stubs.**

12 tests here are `markTestSkipped('Removed by enqueue-only command refactor.')`, so actual command coverage is much thinner than the file size suggests. Either delete the stubs or replace them with tests for the current enqueue-only behavior (e.g. concurrent-run rejection while the lock is held).

docroot/modules/custom/mass_redirect_normalizer/tests/src/ExistingSite/ResolverTest.php
```diff
@@ -0,0 +1,894 @@
...
+
+/**
+ * Tests the redirect link resolver and normalization manager behaviour.
+ *
+ * Covers: RedirectLinkResolver, RedirectLinkNormalizationManager,
+ * presave hook integration, paragraph normalization, and entity references.
+ *
+ * @group existing-site
+ */
+class ResolverTest extends MassExistingSiteBase {
```
dstorozhuk, line 20: **Test gap — no DOM round-trip fixture with non-ASCII content / `&nbsp;` / `<drupal-media>`.**

All rich-text fixtures are pure-ASCII `<p><a>`. Since one rewritten link causes the entire field value to be re-serialized through `Html::load()` / `Html::serialize()`, an encoding/serialization regression would corrupt content at scale with no failing test.

Suggested: a body with one redirecting link plus accented/Cyrillic text, `&nbsp;`, and an embedded `<drupal-media>`; assert the link is rewritten and everything else survives byte-for-byte.

docroot/modules/custom/mass_redirect_normalizer/src/RedirectLinkResolver.php
arthurbaghdas: Fixed.

resolveRedirectTarget() now preserves query/fragment from redirect destination URIs (e.g. /old-page → /search?category=health), and only falls back to the source link’s query/fragment when the destination has none. Alias resolution is unchanged.

Added tests for destination query preservation, destination-over-source precedence, and source fallback. All module tests pass.

docroot/modules/custom/mass_redirect_normalizer/src/RedirectLinkResolver.php
arthurbaghdas: Fixed.

normalizeRedirectLinkUri() was converting any resolved media path to entity:media/N, which dropped /download on routes like /media/N/download. We now only use entity: when the resolved path matches the canonical entity route; /download paths stay as internal:/media/N/download. Added a test.

docroot/modules/custom/mass_redirect_normalizer/src/RedirectLinkNormalizationManager.php
arthurbaghdas: Done. Tansactional improvement applied.

docroot/modules/custom/mass_redirect_normalizer/src/RedirectLinkNormalizationManager.php
arthurbaghdas: Fixed.

docroot/modules/custom/mass_redirect_normalizer/src/RedirectLinkChangeLog.php
arthurbaghdas: Fixed

docroot/modules/custom/mass_redirect_normalizer/src/RedirectLinkNormalizationManager.php
arthurbaghdas: Fixed.

docroot/modules/custom/mass_redirect_normalizer/src/RedirectLinkResolver.php
arthurbaghdas: Good catch — agreed. loadRedirectBySourcePath() no longer uses range(0, 1) and instead bails when more than one redirect matches the source path, consistent with loadRedirectsBySourcePath() / ambiguous_redirects on the entity-reference path. Test added for the duplicate-path case.

docroot/modules/custom/mass_redirect_normalizer/tests/src/ExistingSite/QueueWorkerTest.php
arthurbaghdas: Added

docroot/modules/custom/mass_redirect_normalizer/tests/src/ExistingSite/QueueWorkerTest.php
arthurbaghdas: Added

docroot/modules/custom/mass_redirect_normalizer/src/RedirectLinkNormalizationEligibility.php
arthurbaghdas: Added

docroot/modules/custom/mass_redirect_normalizer/tests/src/ExistingSite/DrushCommandTest.php
arthurbaghdas: Done

docroot/modules/custom/mass_redirect_normalizer/tests/src/ExistingSite/ResolverTest.php
arthurbaghdas: Added testRichTextNormalizationPreservesNonAsciiContentAndMediaEmbed() — rich-text fixture with accented/Cyrillic text, &nbsp;, an embedded <drupal-media> tag, and one redirecting link. Asserts the link is rewritten while the prefix and suffix (all non-link markup) survive byte-for-byte through the DOM round-trip.

dstorozhuk (approved):

PR #3347, opened 2026-04-28
DP-46097: Review Issues suggested by codex

arthurbaghdas:
**Description:**

DP-****

**To Test:**

---


docroot/sites/default/settings.php
```diff
@@ -31,6 +31,39 @@
...
+  '^.+\.mass\.gov$',
+  '^.+\.digital\.mass\.gov$',
+  '^.+\.acquia-sites\.com$',
+  '^.+\.prod\.hosting\.acquia\.com$',
+  '^.+\.stg\.hosting\.acquia\.com$',
+  '^.+\.dev\.hosting\.acquia\.com$',
+  '^.+\.tugboatqa\.com$',
+];
+
+$trusted_host_patterns = getenv('DRUPAL_TRUSTED_HOST_PATTERNS');
```
tfleming-ma: @arthurbaghdas We include other settings files after this point. I'm inclined to say we don't need this extra code for including more trusted hosts. This can instead be done with `$settings['trusted_host_patterns'][] = NEW-HOST-PATTERN`

arthurbaghdas: Makes sense, changed.

conf/drupal/config/http_response_headers.response_header.strict_transport_security.yml
```diff
@@ -0,0 +1,10 @@
+uuid: a33d9804-1975-4b55-a7b1-ff3d80411593
+langcode: en
+status: false
+dependencies: {  }
+id: strict_transport_security
+label: Strict-Transport-Security
+description: 'Enforce HTTPS in supported browsers. Disabled until HSTS is coordinated with Akamai and the http to www redirects, see DP-7408.'
```
tfleming-ma:
```suggestion
description: 'Enforce HTTPS in supported browsers. Disabled until DP-7408 is complete.'
```

docroot/sites/default/settings.vm.php
```diff
@@ -24,6 +24,28 @@
...
+ *
+ * settings.php trusts the public mass.gov domains. Local and CI environments
+ * serve the site under hostnames that vary per developer, since the DDEV
+ * project name decides both the domain (mass.local, mass2.local, ...) and the
+ * web container hostname, so trust the whole family of local hostnames here
+ * rather than a single hard coded project name. This file never runs on Acquia.
+ */
+// Hostnames without a dot: the web container (web, mass-web), localhost, and
+// the other service containers used by PHPUnit, Selenium and Backstop.
+$settings['trusted_host_patterns'][] = '^[a-z0-9-]+$';
```
tfleming-ma, line 38: @dstorozhuk These settings will also be applied to Tugboat, which is public facing. Codex suggested adding a check around this set of trusted_host_patterns to prevent setting these patterns in Tugboat.

dstorozhuk: I was unclear about that comment, since Arthur added the trusted host pattern to settings.tugboat.php as addition to this one. But if the comment was to enforce just  tugboat host pattern - I did that instead of what Arthur did.

tfleming-ma: @dstorozhuk The feedback was that the trusted_host_patterns in settings.vm.php will also be included in settings.tugboat.php, since the settings.tugboat.php file is included is below this code. However, the trusted_host_patterns in settings.vm.php appear to be intended only for ddev and test runner environments. So, the suggestion would be to wrap this set of trusted_host_patterns in settings.vm.php with something like

```
if (!getenv('TUGBOAT_ROOT')) {
  $settings['trusted_host_patterns'][] = '^[a-z0-9-]+$';
  ...
}
```
We would then keep the trusted_host_patterns set in settings.tugboat.php

dstorozhuk: yesterday I added enforcement in settings.tugboat.php, so, everything else defined in `$settings['trusted_host_patterns']` earlier is ignored. 

see:  https://github.com/massgov/openmass/pull/3347/changes/383ce3f48b73d6e525609bb3c2d4968c8209ac43

tfleming-ma: Ah, I understand. I missed that this commit was replacing the array. Thanks!

PR #3354, opened 2026-04-29
DP-45871: Glossary Terms Double-Defining Content and Not Recognizing …

arthurbaghdas:
…Full-Term Matches

**Description:**

DP-****

**To Test:**

---


docroot/themes/custom/mass_theme/overrides/js/glossaries.js
```diff
@@ -52,14 +55,15 @@
       // Replace regex special characters with escaped versions.
       replaced = replaced.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
 
-      // Replace any non-escaped, non alphanumeric characters with wildcards
-      replaced = replaced.replace(/[^\w\s]/g, '.?');
+      // Replace punctuation/symbol characters with optional wildcard chars.
+      // Keep backslashes intact so escaped regex tokens are not broken.
+      replaced = replaced.replace(/[^\w\s\\]/g, '.?');
```
jarielrrz: glossary terms with parentheses in the key (e.g. Bond Anticipation Note (BAN), Community Preservation Act (CPA)) look like they won’t match the same text in the page with the new regex building. I reproduced that on a small set of examples. 

/glossary/municipal-finance-glossary
Bond Anticipation Note (BAN)
Community Preservation Act (CPA)
Tax Anticipation Notes (TAN)
Grant Anticipation Notes (GAN)
Federal Aid Anticipation Note (FAAN)
Minimum Residential Factor (MRF)
Municipal Revenue Growth Fact

jarielrrz (changes requested):

docroot/themes/custom/mass_theme/overrides/js/glossaries.js
arthurbaghdas: Good catch, can you test again?

docroot/themes/custom/mass_theme/overrides/js/glossaries.js
```diff
@@ -265,22 +268,51 @@
         }
       });
 
-      // Sort match positions by their start index (descending)
-      matchPositions.sort((a, b) => b.start - a.start);
+      // Prefer complete terms over partial overlaps.
+      // Example: keep "state house notes", reject overlapping "note".
+      matchCandidates.sort((a, b) => {
```
jarielrrz, line 275: The changes look good, and I ran the pr through Codex and it flagged one remaining issue in the overlap resolution logic, which I was able to reproduce with two Municipal Finance Glossary terms: 

General Fund and Fund Balance. 

Using the text:

General Fund Balance is reported annually.


the current code finds both matches:

General Fund at [0, 12)

Fund Balance at [8, 20)

After sorting by start position and accepting the first overlapping match, the code keeps General Fund and drops Fund Balance. So the parenthesis/self-match issue appears to be fixed, but the overlap logic still prefers the earliest match even when a later overlapping term is also valid. The new automated test covers the suffix-overlap case (State House Notes vs. Note), but it does not cover this overlap pattern. I don't believe this is the intended behavior, passing it back  in case the intended behavior here should be different.

jarielrrz (changes requested):

docroot/themes/custom/mass_theme/overrides/js/glossaries.js
arthurbaghdas: On the middle-overlap case (General Fund vs Fund Balance), we only allow one highlight in an overlapping span. I updated the resolver to prefer the longer match, then the earlier one on ties, and added a test for that scenario. For equal-length middle overlaps, the earlier term still wins.

jarielrrz (approved):
LGTM

PR #3357, opened 2026-05-04
Added logger for detecting oversized headers and tracking their conte…

tfleming-ma:
**Description:**
Added logger for detecting oversized headers and tracking their content types

DP-46681

**To Test:**
- [ ] Test locally
- [ ] In .ddev/.env, set `MASS_PURGE_HEADER_SIZE_LOGGER_ENABLED=true`
- [ ] In .ddev/.enc, set `MASS_PURGE_HEADER_SIZE_THRESHOLD=100`
- [ ] Clear caches, load page

---


arthurbaghdas:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against b6455125cd3c68d041bb3191a5250b23d7fad2ea
</p>

docroot/modules/custom/mass_caching/src/EventSubscriber/PurgeHeaderSizeLoggerSubscriber.php
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * The purge tags header name.
+   */
+  private const HEADER_NAME = 'X-Acquia-Purge-Tags';
+
+  /**
+   * Logs when the header line reaches or exceeds this many bytes.
+   */
+  private const HEADER_SIZE_THRESHOLD = 8192;
```
tfleming-ma, line 25: There are a couple ways we can approach the threshold. We can either keep this at 8192, which is Acquia's hard limit, and see if this lines up with the apache errors we currently see, or we lower the threshold and capture what content is close to hitting this limit. I'm currently choosing the former option, but welcome feedback.

docroot/modules/custom/mass_caching/src/EventSubscriber/PurgeHeaderSizeLoggerSubscriber.php
```diff
@@ -0,0 +1,100 @@
...
+    $route_name = (string) $request->attributes->get('_route', '');
+    $path = $request->getPathInfo();
+
+    $content_type = '';
+    $entity_type = '';
+    $entity_id = '';
+
+    $node = $request->attributes->get('node');
+    if (is_numeric($node)) {
+      $node = \Drupal::entityTypeManager()->getStorage('node')->load((int) $node);
```
dstorozhuk: Is this is temporary solution and we will remove it - ok to have no DI. If we are going to keep it - then I would suggest to use DI. 
And in general - it would be nice to allow to enable/disable logger feature.

docroot/modules/custom/mass_caching/src/EventSubscriber/PurgeHeaderSizeLoggerSubscriber.php
```diff
@@ -0,0 +1,100 @@
...
+        'entity_id' => $entity_id ?: 'n/a',
+      ]
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function getSubscribedEvents(): array {
+    return [
```
dstorozhuk, line 107: @tfleming-ma  I suggest to add a switcher here otherwise at the beginning of `onKernelResponse()` to check the environment variable which we can define in Acquia admin UI and control execution of logger.

dstorozhuk (changes requested):
LGTM, just need to a switcher to enable/disable the logger on prod.

docroot/modules/custom/mass_caching/src/EventSubscriber/PurgeHeaderSizeLoggerSubscriber.php
tfleming-ma: Based on similar examples I saw in the codebase, decided to put this in `onKernelResponse` instead. Also looks like I wouldn't be able to call $this->isEnabled() in getSubscribedEvents(). Updated testing instructions.

docroot/modules/custom/mass_caching/src/EventSubscriber/PurgeHeaderSizeLoggerSubscriber.php
tfleming-ma: It should be temporary. But, I think we should use DI anyway. Updated.  Thanks.

docroot/modules/custom/mass_caching/src/EventSubscriber/PurgeHeaderSizeLoggerSubscriber.php
dstorozhuk: @tfleming-ma  simplest variant to make this variable from ENV far, but set default value. If you my suggestion make sense - then just add a merge. Otherwise - just merge.

dstorozhuk (approved):

PR #3361, opened 2026-05-07
DP-43727: Tests for Bulk document import enhancements to validation/verification

arthurbaghdas:
**Description:**

DP-****

**To Test:**

---


docroot/modules/custom/mass_bulk_file_replace/tests/src/ExistingSiteJavascript/BulkFileReplaceUiTest.php
```diff
@@ -0,0 +1,345 @@
...
+  if (!row) {
+    return false;
+  }
+  var cb = row.querySelector('input[type="checkbox"]:not([disabled])');
+  return !!(cb && cb.checked);
+})()
+JS
+    );
+    $this->assertTrue((bool) $checked, 'Matched mismatch row is selected by default.');
+  }
```
dstorozhuk, line 277: **Critical gap: `ReplaceMismatchForm::processBatch()` happy path is never exercised.**

This test seeds tempstore and inspects rendering, but never presses **"Replace Approved Files"** — the entire reason the mismatch screen exists. Verified locally that pressing the button correctly creates a new media revision, strips the `_DO_NOT_CHANGE_THIS_MEDIA_ID_…` token from the persisted filename, and replaces the file content. A regression in `ReplaceMismatchForm.php:195-291` (`processBatch`) would ship undetected.

Suggest adding a test that:
1. Seeds the mismatch tempstore (as this test already does).
2. Presses "Replace Approved Files".
3. Asserts on the **persisted entity** after the batch:
   - `$media->getRevisionId()` advanced;
   - `$file->getFilename()` equals the stripped form;
   - `$file->getFileUri()` points into the media field's `file_directory`;
   - file content matches the uploaded bytes;
   - tempstore entry was cleared (`ReplaceMismatchForm.php:277-281`).

docroot/modules/custom/mass_bulk_file_replace/tests/src/ExistingSiteJavascript/BulkFileReplaceUiTest.php
```diff
@@ -0,0 +1,345 @@
...
+      'status' => 0,
+    ]);
+    $file_orphan->save();
+    $this->markEntityForCleanup($file_orphan);
+
+    $this->seedMismatchTempstore($uid, [(int) $file_token->id(), (int) $file_orphan->id()]);
+
+    $this->drupalGet('admin/ma-dash/download-documents/mismatch');
+
+    $this->assertSession()->pageTextNotContains('DO_NOT_CHANGE_THIS_MEDIA_ID_');
```
dstorozhuk, line 258: **Critical gap: token stripping is asserted only at the rendering layer.**

`pageTextNotContains('DO_NOT_CHANGE_THIS_MEDIA_ID_')` confirms the display name on the review page, but `ReplaceMismatchForm.php:211-227` (the actual filename normalization on save) is never asserted against. After the eventual `processBatch` runs, `$file->getFilename()` should equal `brand-new.pdf` (token + auto-suffix removed). Today nothing fails if that normalization regresses.

Once the test for finding #1 (above) is added, this becomes a single extra `assertSame('brand-new.pdf', $file->getFilename())` after reloading the media.

docroot/modules/custom/mass_bulk_file_replace/tests/src/ExistingSiteJavascript/BulkFileReplaceUiTest.php
```diff
@@ -0,0 +1,345 @@
...
+
+    $this->getSession()->getPage()->pressButton('Start replacement');
+    $wait_js = "document.body && (document.body.innerText.indexOf(" .
+      "'All uploaded files were successfully replaced.') !== -1 || " .
+      "document.body.innerText.indexOf('Some uploaded files need') !== -1 || " .
+      "document.body.innerText.indexOf('No files were uploaded') !== -1)";
+    $this->getSession()->wait(120000, $wait_js);
+
+    $page_text = $this->getSession()->getPage()->getText();
+    $this->assertStringNotContainsString('No files were uploaded', $page_text);
```
dstorozhuk, line 219: **Critical: this wait + assertion can silently pass on a real regression.**

```php
$wait_js = "... 'All uploaded files were successfully replaced.' || 'Some uploaded files need' || 'No files were uploaded' ...";
$this->getSession()->wait(120000, $wait_js);
$page_text = $this->getSession()->getPage()->getText();
$this->assertStringNotContainsString('No files were uploaded', $page_text);
```

The wait resolves as soon as **any** of the three messages appears, and only `'No files were uploaded'` is asserted absent. If the auto-replace contract breaks and the user is redirected to the mismatch screen (showing `'Some uploaded files need'`), the wait still resolves, the assertion still passes, and the test only catches the regression *if* the media reload at lines 228-237 happens to show a stale entity.

Suggest:
```php
$success = "document.body.innerText.indexOf('All uploaded files were successfully replaced.') !== -1";
$this->getSession()->wait(120000, $success);
$page_text = $this->getSession()->getPage()->getText();
$this->assertStringContainsString('All uploaded files were successfully replaced.', $page_text);
$this->assertStringNotContainsString('Some uploaded files need', $page_text);
```

docroot/modules/custom/mass_bulk_file_replace/tests/src/ExistingSite/BulkFileReplaceTest.php
```diff
@@ -0,0 +1,285 @@
...
+
+    $context = [];
+    ReplaceUploadForm::processFileBatch([
+      'filename' => $upload_name,
+      'path' => $path,
+    ], $uid, 'tester', 'verified_accessible', $context);
+
+    $store = \Drupal::service('tempstore.private')->get('mass_bulk_file_replace');
+    $mismatch = $store->get('mismatch_files_' . $uid) ?? [];
+    $this->assertSame([], $mismatch);
```
dstorozhuk: **`?? []` masks a NULL tempstore, hiding the case where the code path never ran.**

```php
$mismatch = $store->get('mismatch_files_' . $uid) ?? [];
$this->assertSame([], $mismatch);
```

If `processFileBatch` short-circuits before touching tempstore (e.g. an early return that wasn't supposed to happen), `get()` returns NULL → `?? []` becomes `[]` → assertion passes. The same pattern repeats at line 191.

For the safe-match auto-replace path, `processFileBatch` never writes to tempstore at all, so the correct assertion is `assertNull($store->get('mismatch_files_' . $uid))`. That distinguishes "code ran and stored nothing" from "code never ran."

docroot/modules/custom/mass_bulk_file_replace/tests/src/ExistingSiteJavascript/BulkFileReplaceUiTest.php
arthurbaghdas: Done

docroot/modules/custom/mass_bulk_file_replace/tests/src/ExistingSiteJavascript/BulkFileReplaceUiTest.php
arthurbaghdas: Done

docroot/modules/custom/mass_bulk_file_replace/tests/src/ExistingSiteJavascript/BulkFileReplaceUiTest.php
arthurbaghdas: Done

docroot/modules/custom/mass_bulk_file_replace/tests/src/ExistingSite/BulkFileReplaceTest.php
arthurbaghdas: Done

docroot/modules/custom/mass_bulk_file_replace/tests/src/ExistingSiteJavascript/BulkFileReplaceUiTest.php
```diff
@@ -0,0 +1,303 @@
...
+  }
+
+  /**
+   * Sets the Drupal current user for tempstore operations.
+   */
+  private function useBulkReplaceTestUser(int $uid): void {
+    $account = User::load($uid);
+    if ($account) {
+      \Drupal::currentUser()->setAccount($account);
+    }
```
dstorozhuk: **Silent no-op on missing user — regression of the previously-correct pattern.**

If `User::load($uid)` returns NULL (cleanup race, stale `$bulkReplaceUserId`, DB hiccup), `setAccount` doesn't run and every downstream tempstore op silently switches to the anonymous user's namespace (`mismatch_files_0`). `assertMismatchTempstoreUnset` at line 300 then passes triumphantly on an always-empty key — green for the wrong reason.

In the earlier version, `testCancelClearsMismatchTempstore` had an explicit `assertNotNull($account)` right before the account switch. The refactor routed all four call sites (`tearDown` line 48, `seedMismatchTempstoreForUser` line 71, `testUploadAutoReplacesOnExactFilename` line 189, `testCancelClearsMismatchTempstore` line 299) through this helper — losing that guard everywhere.

One line inside the helper fixes all four sites:

```suggestion
    $account = User::load($uid);
    $this->assertNotNull($account, "Bulk replace test user $uid must exist before switching account.");
    \Drupal::currentUser()->setAccount($account);
```

docroot/modules/custom/mass_bulk_file_replace/tests/src/ExistingSiteJavascript/BulkFileReplaceUiTest.php
```diff
@@ -0,0 +1,303 @@
...
+    $this->assertNotNull($uid);
+    $this->useBulkReplaceTestUser($uid);
+    $this->clearMismatchTempstore($uid);
+
+    $media = $this->createDocumentMediaWithFile('housing-proposal.pdf', 'before');
+    $mid = (int) $media->id();
+    // Basename must match token + id; Dropzone posts this name to the server.
+    $upload_basename = 'housing-proposal_DO_NOT_CHANGE_THIS_MEDIA_ID_' . $mid . '.pdf';
+    $dir = sys_get_temp_dir() . '/bulk-replace-ui-' . $this->randomMachineName();
+    mkdir($dir, 0777, TRUE);
```
dstorozhuk: **`mkdir()` return value ignored.**

If `mkdir` fails (permissions / disk full / race), `writeDiskFile` then fails at `file_put_contents` and the error points at the file, not the real cause. The other `mkdir` site in `writeDiskFile` at line 161 at least guards with `is_dir()` — this one has no guard at all.

```suggestion
    $this->assertTrue(mkdir($dir, 0777, TRUE) || is_dir($dir), "Failed to create $dir");
```

dstorozhuk (changes requested):

dstorozhuk (approved):

PR #3366, opened 2026-05-11
DP-46855: Akamai purge changes

tfleming-ma:
**Description:**
Purge URLs in Akamai when pages are saved in more cases.

DP-46855

**To Test:**
- [ ] Make a change to a page that doesn't affect the URL alias (not title)
- [ ] Run `drush pqb` and check for URL purges in the queue. You should see the node URL and its aliases in the queue

NOTE: If you try these steps on an environment without this fix, you won't see the URL updated.

---


docroot/modules/custom/mass_caching/mass_caching.module
```diff
@@ -183,16 +188,21 @@ function mass_caching_entity_update(EntityInterface $entity) {
...
+ * to run after pathauto's hook_entity_update().
  */
 function mass_caching_entity_insert(EntityInterface $entity) {
   switch ($entity->getEntityTypeId()) {
     case 'media':
       mass_caching_purge_media($entity);
       break;
 
+    case 'node':
+      \Drupal::service('mass_caching.node_purge_handler')->purge($entity);
```
tfleming-ma: This was initially implemented as a helper function like the media and redirect purges. I think it make more sense that all the purges be called as services instead of helper functions, but my preference is that we wait to handle converting the other two helper functions to services in a separate pull request.

docroot/modules/custom/mass_caching/src/NodePurgeHandler.php
```diff
@@ -0,0 +1,134 @@
...
+  public function purge(NodeInterface $node): void {
+    if (!$this->shouldPurge($node)) {
+      return;
+    }
+
+    $paths = ['/node/' . $node->id()];
+    $paths[] = $this->getCurrentAlias($node);
+    $paths[] = $this->getPathFieldAlias($node);
+
+    if (isset($node->original) && $node->original instanceof NodeInterface) {
```
dstorozhuk: @tfleming-ma : 
`@trigger_error("Getting the original property is deprecated in drupal:11.2.0 and is removed from drupal:12.0.0. Use \Drupal\Core\Entity\EntityInterface::getOriginal() instead. See https://www.drupal.org/node/3295826", E_USER_DEPRECATED);` Need to fix across the code.

docroot/modules/custom/mass_caching/src/Hook/NodePurgeHandler.php
```diff
@@ -0,0 +1,134 @@
...
+   * @param \Drupal\node\NodeInterface $node
+   *   A node entity.
+   */
+  public function purge(NodeInterface $node): void {
+    if (!$this->shouldPurge($node)) {
+      return;
+    }
+
+    $paths = ['/node/' . $node->id()];
+    $paths[] = $this->getCurrentAlias($node);
```
dstorozhuk, line 80: @tfleming-ma  worth to test that scenario: 
Script: the page is published, it has a pathauto-generated alias. The user changes the title, pathauto regenerates the alias.

So the ideas is to test purge will still work after title/alias change. If tested and works fine - then skip this comment.

docroot/modules/custom/mass_caching/mass_caching.module
dstorozhuk: @tfleming-ma  I would suggest  add #[hook] to `NodePurgeHandler::purge` and simply place `switch` to `shouldPurge()` of add `if` guard. There is no demand for service for this functionality. 

I don't like to support old hooks definitions approach if we have a loot of legacy code written in that way. We should use new OOP way where possible, this makes code understanding, maintainable, and structured.

dstorozhuk (changes requested):
Left some comments.

docroot/modules/custom/mass_caching/src/Hook/NodePurgeHandler.php
tfleming-ma: Yeah, tested and worked fine, and actually, that was one of the few scenarios before this change where URLs would be purged.

docroot/modules/custom/mass_caching/src/Hook/NodePurgeHandler.php
```diff
@@ -0,0 +1,176 @@
...
+   * Purge URL paths on entity update.
+   *
+   * @param \Drupal\Core\Entity\EntityInterface $entity
+   *
+   * @return void
+   *   This method does not return any value.
+   *
+   * @throws \Drupal\Core\TypedData\Exception\MissingDataException
+   */
+  #[Hook('entity_update', order: new OrderAfter(modules: ['pathauto']))]
```
tfleming-ma: @dstorozhuk See this hook and the other for entityInsert

docroot/modules/custom/mass_caching/mass_caching.services.yml
```diff
@@ -8,6 +8,7 @@ services:
   manual_purger:
     class: Drupal\mass_caching\ManualPurger
     arguments: ['@purge.queuers', '@purge.invalidation.factory', '@purge.queue', '@settings']
+  Drupal\mass_caching\ManualPurger: '@manual_purger'
```
tfleming-ma, line 11: This alias was added due to an error and recommendation from the validate job in Circle

docroot/modules/custom/mass_caching/src/NodePurgeHandler.php
tfleming-ma: Resolved

docroot/modules/custom/mass_caching/mass_caching.module
tfleming-ma: See changes in NodePurgeHandler. Hooks have been added for entityUpdate and entityInsert

docroot/modules/custom/mass_caching/mass_caching.services.yml
dstorozhuk: @tfleming-ma  we don't need definition in service.yml https://www.drupal.org/node/3442349 after  d11.1 

if Circle trigger an error - then we need to fix the the CirlceCI config, but this can be separate task. If not passible to pass validation without service.yml definition - I would propose create a follow up ticket, and merge this tasks as is.

docroot/modules/custom/mass_caching/src/Hook/NodePurgeHandler.php
dstorozhuk: @tfleming-ma  you can simply collapse two methods: 

```
#[Hook('entity_update', order: new OrderAfter(modules: ['pathauto']))]
#[Hook('entity_insert', order: new OrderAfter(modules: ['pathauto']))]
 public function entityUpdate(EntityInterface $entity): void {
    if ($entity instanceof NodeInterface) {
      $this->purgeNode($entity);
    }
  }

```

dstorozhuk (changes requested):
Minor request.

docroot/modules/custom/mass_caching/src/Hook/NodePurgeHandler.php
tfleming-ma: Yeah, so I am aware that you can do this, but chose to split after comparing to other modules' implementations, specifically Content Moderation in core and the AI contrib module, and noticing some editor behavior. In the editor, it seems like if the hook name and method name don't match (in this case it matches if the method is defined in camelCase and the hook is defined in snake_case), the method is marked as unused. 

With all that said, updated to single `entityInsertOrUpdate` method, which I think is ultimately a better approach.

docroot/modules/custom/mass_caching/mass_caching.services.yml
tfleming-ma: Will resolve in a follow-up ticket

dstorozhuk (approved):
@tfleming-ma  Look good to me. The ticket is not on me, if there is nothing from you side - you can merge.

PR #3388, opened 2026-05-21
DP-46854: Enable Akamai cache tag purging

tfleming-ma:
**Description:**
- Enables Akamai cache tag purging
- Sets list in Akamai purger config to block certain cache tags. Standard list, minus `node_list`, is used: https://www.drupal.org/docs/contributed-modules/akamai/cache-tag-purging
- Explicitly removes the `node_list` tag. Other `node_list:` tags are allowed. This is done because items defined in Akamai's default config are all treated as a `strpos` matches
- Uses hashing from the Acquia Purge module to hash Akamai's cache tags

DP-46854

**To Test:**

---


conf/drupal/config/akamai.settings.yml
```diff
@@ -9,16 +9,68 @@ client_secret: akamai_client_secret
 access_token: akamai_access_token
 edgerc_path: ''
 edgerc_section: default
-basepath: 'https://stage.mass.gov'
+basepath: 'https://www.mass.gov'
```
dstorozhuk: @tfleming-ma  is that expected change? if yes - curious why it was https://stage.mass.gov before. Probably of testing and was somehow overridden on Prod?

docroot/modules/custom/mass_caching/src/EventSubscriber/AkamaiHeaderCreationSubscriber.php
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  /**
+   * Hashes cache tags before Akamai writes them to the response header.
+   *
+   * @param \Drupal\akamai\Event\AkamaiHeaderEvents $event
+   *   The Akamai header creation event.
+   */
+  public function hashTags(AkamaiHeaderEvents $event): void {
+    $event->data = Hash::cacheTags($event->data);
```
dstorozhuk: **C1 (Critical) — Header tags are hashed, but the tag-purge path is not. Akamai tag purges will silently match nothing.**

`hashTags()` rewrites every `Edge-Cache-Tag` value with `acquia_purge`'s `Hash::cacheTags()` (4-char base32). Akamai purges by **exact string match**, so the string stored in the header must equal the string sent in a Fast Purge "tag" request.

The tag-purge path never hashes:
- `AkamaiTagPurger::invalidate()` (contrib) builds purge tags via `CacheTagFormatter::format()` only — no hashing.
- No `AkamaiPurgeEvents::PURGE_CREATION` subscriber in custom code hashes (the only one, `mass_caching/src/AkamaiPurger.php`, handles URLs).

So the header stores e.g. `sq46` for `node:123`, but a purge sends `node_123` → matches nothing, the purge still reports `SUCCEEDED`, and content stays stale until TTL. A silent no-op.

**Latent today** because only the URL purger (`akamai`) is enabled in `purge.plugins.yml`, not `akamai_tag`. But whoever enables `akamai_tag` next will hit this. Fix: add a `PURGE_CREATION` subscriber applying the identical `Hash::cacheTags()`, and a round-trip test (see I4).

Proof — round-trip test run in CI, fails as expected: https://app.circleci.com/pipelines/gh/massgov/openmass/42235/workflows/70d6ae47-3e97-4722-96b3-1ed1cd8ae970/jobs/179682/tests

docroot/modules/custom/mass_caching/src/EventSubscriber/AkamaiHeaderCreationSubscriber.php
```diff
@@ -0,0 +1,45 @@
...
+namespace Drupal\mass_caching\EventSubscriber;
+
+use Drupal\acquia_purge\AcquiaCloud\Hash;
+use Drupal\akamai\Event\AkamaiHeaderEvents;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * Alters Akamai header cache tags before the header is created.
+ */
+class AkamaiHeaderCreationSubscriber implements EventSubscriberInterface {
```
dstorozhuk, line 15: **I3 (Important) — No guard for Akamai's 128-tag COUNT limit.**

Hashing solves the 8 KB header **size** limit, but not Akamai's **128-tag count** limit. On heavy pages (e.g. org pages with large menus) the tag set can still exceed 128 after the denylist + `node_list` removal. Nothing here counts the tags or logs when the limit is exceeded, so Akamai will silently truncate/reject the oversized header.

Consider logging a warning (via `logger.channel.mass_caching`) when `count($tags) > 128`, so it is visible instead of failing silently. (Note: `AkamaiTagPurger`'s existing warning is for tag *length* > 128 chars — a different limit, and on a purger that is not enabled.)

conf/drupal/config/akamai.settings.yml
```diff
@@ -9,16 +9,68 @@ client_secret: akamai_client_secret
...
+  - local_
+  - entity_
+  - workflow_
+  - element_
+  - contextual_links_
+  - link_relation_
+  - content_moderation_
+  - webform_submission_list
+  - redirect_list
+  - 'node_list:action'
```
dstorozhuk, line 40: **I1 (Important) — Denylisting every `node_list:<bundle>` defeats edge tag invalidation and contradicts the ticket.**

All 36 `node_list:<bundle>` tags are denylisted here, and the bare `node_list` is stripped in `AkamaiHeaderCreationSubscriber`, so **no `node_list` tag of any kind reaches the `Edge-Cache-Tag` header**. List/landing pages (news, events, etc.) will then never be invalidated at the edge when content changes — which is much of the point of this feature ("ensure news and events on other pages update as soon as possible").

This also contradicts the ticket follow-up (comment of May 28: *"Don't block `node_list:<bundle>`"*, and *"1-2 are covered in the pull request"*). Either this is an intentional ultra-conservative start — then please document it and confirm with the team — or the bundle entries were added by mistake and should be removed from the denylist.

conf/drupal/config/akamai.settings.yml
```diff
@@ -9,16 +9,68 @@ client_secret: akamai_client_secret
...
+edge_cache_tag_header: true
 purge_urls_with_hostname: false
-edge_cache_tag_header_blacklist: {  }
+edge_cache_tag_header_blacklist:
+  - 4xx-response
+  - theme_registry
+  - route_match
+  - routes
+  - 'config:'
+  - user
```
dstorozhuk, line 29: **I2 (Important) — Denylist matches by substring, not prefix.**

The contrib filter is `strpos($tag, $prefix) !== FALSE` (`CacheableResponseSubscriber::onRespond()`) — a **substring** match anywhere in the tag, not a prefix. Bare tokens like `user`, `entity_`, `local_`, `routes` silently drop any tag merely *containing* that substring. It matches the current intent, but it is a footgun: a future short entry (e.g. `node`) would nuke `node:123`, `node_view`, and every `node_list:*`.

Worth a comment here noting entries are matched as substrings, or moving the denylist into the subscriber with explicit prefix semantics.

docroot/modules/custom/mass_caching/tests/src/ExistingSite/AkamaiCacheableResponseSubscriberTest.php
```diff
@@ -0,0 +1,121 @@
...
+        'edge_cache_tag_header_blacklist' => [],
+      ],
+    ]);
+    $this->container->get('config.factory')->clearStaticCache();
+
+    $response = $kernel->handle($request);
+    $this->assertEquals(200, $response->getStatusCode());
+    $this->assertTrue($response->headers->has('Edge-Cache-Tag'));
+    $tags = $this->getEdgeCacheTags($response);
+    $this->assertContains('l2u0', $tags);
```
dstorozhuk, line 77: **I4 (Important) — Tests cover only header generation, not the header↔purge round-trip (the contract that actually matters).**

These assertions prove the header is hashed (`rendered` → `l2u0`), but nothing asserts that the hashed string in the header equals the string a tag purge sends to Akamai — the contract the whole feature depends on (see C1).

Note the contradiction this test already encodes: it asserts the header contains `l2u0` for `rendered`, yet a tag purge would send `rendered` — they can never match.

A round-trip test catches this. I wrote one and ran it in CI; it fails as expected: https://app.circleci.com/pipelines/gh/massgov/openmass/42235/workflows/70d6ae47-3e97-4722-96b3-1ed1cd8ae970/jobs/179682/tests

Suggest adding such a test (header tag == purge tag for the same source tag) before enabling `akamai_tag`.

dstorozhuk (changes requested):

conf/drupal/config/akamai.settings.yml
tfleming-ma: Yeah, that's annoying, and apparently was reported as a bug: https://www.drupal.org/project/akamai/issues/3527333. Will patch

conf/drupal/config/akamai.settings.yml
tfleming-ma: Because we can remove tags from the denylist manually in the UI, this full list was intentional for the initial release and the recent comments in the ticket reflect this.

The bare `node_list` tag is stripped in `AkamaiHeaderCreationSubscriber` so that we can allow some `node_list:` bundle tags while ignoring the bare `node_list` tag and still respect the intention of the prefix-driven denylist provided by the contrib module.

conf/drupal/config/akamai.settings.yml
tfleming-ma: I'm going to roll this change back. I made the change based on the module configuration, which made it seem like it was required to point to the correct Akamai property. In practice, however, the basepath configuration is only used by the default URL purger to construct URLs. We do not use this configuration and instead use a custom configuration that adds both the edit and www versions of URLs to the purge queue (`mass_caching.hosts`). The hosts are set in settings.php

docroot/modules/custom/mass_caching/tests/src/ExistingSite/AkamaiCacheableResponseSubscriberTest.php
tfleming-ma: I decided to take a different approach than the one implemented here, but similar to what is suggested in C1 under "Latent today": https://github.com/massgov/openmass/pull/3418.

I've added a test called `testHeaderTagHashesMatchPurgeTagHashes` inside of `AkamaiHeaderCreationSubscriberTest` and implemented a solution in `AkamaiHeaderCreationSubscriber` that runs `hashTags` on the `PURGE_CREATION` event.

conf/drupal/config/akamai.settings.yml
tfleming-ma: Merge request created on ticket. I downloaded the diff and added it as a patch to PR: https://www.drupal.org/project/akamai/issues/3527333

docroot/modules/custom/mass_caching/src/EventSubscriber/AkamaiHeaderCreationSubscriber.php
tfleming-ma: See https://github.com/massgov/openmass/pull/3388#discussion_r3467651759

docroot/modules/custom/mass_caching/src/EventSubscriber/AkamaiHeaderCreationSubscriber.php
tfleming-ma: I added logging to `onHeaderCreation`. Codex also added a test for this logging. I don't think this is something I would have done myself since it's not a functional change. But, this does make testing the logger easier and we do need to pass in the logger anyway to create subscriber, so I left this in place. See `testOversizedHeaderTagSetLogged()`

dstorozhuk (approved):

PR #3394, opened 2026-05-26
DP-47095-Non-admins-should-not-see-or-edit-approval-fields-in-user-profile

jarielrrz:
Restrict user profile approval fields so only administrators can view or edit them.


dstorozhuk (changes requested):
@jarielrrz  the code is not related to `mass_org_access`. Please move to mass_utility or better mass_fields

PR #3395, opened 2026-05-26
DP-47095-Non-admins-should-not-see-or-edit-approval-fields-in-user-profile-develop

jarielrrz:
Restrict user approval fields to users with administer users permission and normalize editor profile form spacing after the fields are hidden.

### DIFF
--- a/docroot/modules/custom/mass_fields/mass_fields.module
+++ b/docroot/modules/custom/mass_fields/mass_fields.module
@@ -740,6 +740,16 @@ function mass_fields_paragraph_view(array &$build, EntityInterface $entity, Enti
  * Implements hook_entity_field_access().
  */
 function mass_fields_entity_field_access($operation, FieldDefinitionInterface $field_definition, AccountInterface $account, ?FieldItemListInterface $items = NULL) {
+  // User approval fields are for access managers only.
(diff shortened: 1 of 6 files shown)


docroot/modules/custom/mass_fields/mass_fields.module
```diff
@@ -740,6 +740,16 @@ function mass_fields_paragraph_view(array &$build, EntityInterface $entity, Enti
  * Implements hook_entity_field_access().
  */
 function mass_fields_entity_field_access($operation, FieldDefinitionInterface $field_definition, AccountInterface $account, ?FieldItemListInterface $items = NULL) {
+  // User approval fields are for access managers only.
+  if ($field_definition->getTargetEntityTypeId() === 'user') {
```
dstorozhuk: @jarielrrz  please use new OOP style #hook for `entity_field_access`.  You previous implementation was on the right directions, but in wrong place. You can extract the code you added in previous PR and add to mass_fields. See more here about OOP hook in Drupal: https://www.drupal.org/node/3442349

dstorozhuk (changes requested):

dstorozhuk (approved):

PR #3409, opened 2026-06-17
DP-47635: fix-message-box-color-contrast

clairesunstudio:


jarielrrz (changes requested):
Changelog needs a small fix before merge, Danger is failing because Fixed must be an array.

jarielrrz (approved):
LGTM

PR #3413, opened 2026-06-22
DP-47686: Added additional filters to editoria11y dismissal report

Eileleji:
**Description:**
Added additional filters to editoria11y dismissal report.

DP-47686

**To Test:**

---


jarielrrz (changes requested):
Requesting changes on two items:

composer.lock is out of sync with composer.json
CI validate is failing because the lockfile wasn’t refreshed after composer changes.
Please run ddev composer update --lock and commit the updated composer.lock.

Missing count/grouping behavior from prior feedback
This PR still does not implement:
image

“When there are multiple alerts with the same exact data in all fields, it doesn’t make sense to show multiple lines. We can instead show a column with the count. You can turn on grouping and make a field that counts.”
I’m still seeing this as not included in the current view config.

Otherwise, the alert-type patch itself looks good to me.

jarielrrz (approved):
LGTM

PR #3417, opened 2026-06-22
DP-47707: Stop collection_all View from spamming the watchdog log 2

arthurbaghdas:
**Description:**

DP-****

**To Test:**

---

### DIFF
--- a/docroot/modules/custom/mass_views/mass_views.module
+++ b/docroot/modules/custom/mass_views/mass_views.module
@@ -505,7 +505,12 @@ function mass_views_views_pre_render(ViewExecutable $view) {
       // Media rows are unioned into this node-based view in mass_content.module.
       // Unset node-only fields so Views does not try to load node entities/fields
       // for media rows (e.g. description fields used by the keyword search filter).
+      // Match on $field->table here is the Views YAML "table:" value at pre_render,
+      // not the runtime query tableAlias. Any field whose table starts with "node"
+      // is node-specific; excluded fields are only needed for query-time search.
       foreach ($view->field as $field_id => $field) {
+        // rendered_entity_mixed is registered on table "node" (mass_content.module)
+        // but must be kept: it renders both node and media rows via entity_type.
         if ($field_id === 'rendered_entity_mixed') {
           continue;
         }
--- a/docroot/modules/custom/mass_views/mass_views.module
+++ b/docroot/modules/custom/mass_views/mass_views.module
@@ -493,22 +493,26 @@ function mass_views_views_pre_render(ViewExecutable $view) {
   }
 
   if ($view->id() == 'collection_all') {
+    $has_media_rows = FALSE;
     foreach ($view->result as $row) {
-      // Check if the row entity is a media entity
       if ($row->_entity == NULL && $row->entity_type == 'media') {
-        // If the $row is of type media we should unset the fields below, since the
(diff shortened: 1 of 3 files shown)


docroot/modules/custom/mass_views/mass_views.module
```diff
@@ -493,22 +493,26 @@ function mass_views_views_pre_render(ViewExecutable $view) {
   }
 
   if ($view->id() == 'collection_all') {
+    $has_media_rows = FALSE;
```
dstorozhuk, line 496: **[Architecture / follow-up — not a blocker]** This fix extends the procedural hook `mass_views_views_pre_render()`. Per the project convention (CLAUDE.md, Drupal 11.3) new hook code should prefer OOP `#[Hook()]` attribute classes (in `src/Hook/`, namespace `Drupal\mass_views\Hook`) with DI over procedural hooks and `\Drupal::` statics.

This hook only operates on `$view` (no services needed), so it is one of the cheapest hooks to migrate to a `#[Hook('views_pre_render')]` class. The catch: this PR edits an *existing* procedural hook, `mass_views` has no `src/Hook/` yet (9 procedural hooks live in this `.module`), and a single hook cannot be implemented both procedurally and via `#[Hook()]` without firing twice — so a clean migration would also have to move the unrelated `news_curated_list` branch and remove the procedural function.

Recommendation: keep this PR as-is and track the `mass_views` hook migration in a separate follow-up ticket.

docroot/modules/custom/mass_views/mass_views.module
```diff
@@ -493,22 +493,26 @@ function mass_views_views_pre_render(ViewExecutable $view) {
...
+      // Media rows are unioned into this node-based view in mass_content.module.
+      // Unset node-only fields so Views does not try to load node entities/fields
+      // for media rows (e.g. description fields used by the keyword search filter).
+      foreach ($view->field as $field_id => $field) {
+        if ($field_id === 'rendered_entity_mixed') {
+          continue;
+        }
+        $table = $field->table ?? '';
+        $entity_type = $field->options['entity_type'] ?? '';
+        if (str_starts_with($table, 'node') || $entity_type === 'node') {
```
dstorozhuk, line 519: **Nit (nice-to-have).** A one-line comment here would save the next reader: the match is against `$field->table` as it stands at `pre_render` (the YAML `table:` value, not the runtime `tableAlias`), and `rendered_entity_mixed` is explicitly skipped above because it is registered with `table === 'node'` (`mass_content.module:959`) yet is the one field that must survive — it renders both node and media rows.

Also worth noting in the PR description: this broader `str_starts_with($table, 'node')` criterion now additionally unsets `field_collection_description`, which the old hard-coded 4-field list did not. Verified safe (it is `exclude: true`, data reaches the template via the loaded node entity), but the behavior is now generic rather than an explicit allow-list.

dstorozhuk (changes requested):

docroot/modules/custom/mass_views/mass_views.module
arthurbaghdas: I think we can have a separate ticket for migrating all hooks.

docroot/modules/custom/mass_views/mass_views.module
arthurbaghdas: Added a comment

dstorozhuk (approved):

PR #3421, opened 2026-06-23
DP-45698: Strengthen media download access-control tests

dstorozhuk:
**Description:**

Follow-up to #3295 (DP-45698), based on `feature/DP-45698_media_download_link`. Strengthens the access-control test coverage of the rewritten media download controller. No production code changes — tests + changelog only.

- **Owner-only restricted download** — new positive path: the document owner (granted via the existing `author` role) receives the actual file bytes (200), while a different authenticated non-admin user and anonymous users are denied. This exercises the real `mass_media_media_access()` owner-only contract instead of relying on an incidental core `-1`.
- **Exact denial codes** — both denial tests now assert the precise status (`404`) instead of `assertNotEquals(200)`, which previously also passed on a 500 or a redirect and masked the contract. Empirically, Mass.gov surfaces denial of these documents as 404 (hiding their existence), not a bare 403.
- **Fixes a misleading prior test** — the old restricted test created the "author" with the `administrator` role (which bypasses the restriction entirely) and only proved generic private-file denial; it now drives the genuine owner-only path.

Notes for reviewers:
- A `restricted` moderation state is effectively unpublished (`status 0`) with its file kept in `private://`, so the owner needs `view own unpublished media` — supplied by the `author` role (`download media` comes from `authenticated`).
- The owner uses an existing role rather than a freshly created one on purpose: in an ExistingSite test the web server serving the request does not reliably see the permissions of a role created moments earlier in the test process.

**To Test:**
- [ ] `ddev exec phpunit docroot/modules/custom/mass_media/tests/src/ExistingSite/MediaDownloadTest.php` — all 7 tests pass (requires the `selenium-chrome` service).

**Jira:** DP-45698

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/changelogs/DP-45698.yml
+++ b/changelogs/DP-45698.yml
@@ -39,3 +39,5 @@
 Changed:
   - description: Serve media download links as binary instead of redirect.
     issue: DP-45698
+  - description: Strengthen media download access-control tests with an owner-only restricted-document download case and exact denial status codes.
+    issue: DP-45698
--- a/changelogs/DP-45698.yml
+++ b/changelogs/DP-45698.yml
@@ -39,5 +39,3 @@
 Changed:
   - description: Serve media download links as binary instead of redirect.
     issue: DP-45698
-  - description: Strengthen media download access-control tests with an owner-only restricted-document download case and exact denial status codes.
-    issue: DP-45698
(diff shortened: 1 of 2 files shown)


arthurbaghdas:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 5f3437e1ec9ddf6b13e5f9be28f13b19099bb1bd
</p>

PR #3425, opened 2026-06-24
DP-46833-False-warning-Event-component-reported-missing-on-associated-service-page-after-publishing-event

jarielrrz:
Fix false warning for missing Events component on associated service pages

[DP-46833](https://massgov.atlassian.net/browse/DP-46833)

[DP-46833]: https://massgov.atlassian.net/browse/DP-46833?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dstorozhuk:
Looks good, the fix is correct — verified against the field configs: `field_service_sections` allows `org_events` directly (that's why only service pages false-warned, org sections only allow `org_section_long_form`), and the new check covers exactly that case.

One ask before merge: please add a negative test. The current test only covers the TRUE case. Add a service_page with a `service_section` but no `org_events` and assert the check returns FALSE — that guards the other direction: the warning must still fire when the component is genuinely missing.

Something like:

```php
public function testMissingOrgEventsStillReportedOnServicePage() {
  $service_section = Paragraph::create(['type' => 'service_section']);
  $service_section->save();
  $this->markEntityForCleanup($service_section);

  $service_page = $this->createNode([
    'type' => 'service_page',
    'title' => 'Test Service Page Without Events',
    'field_service_lede' => 'Test lede.',
    'field_service_sections' => [['entity' => $service_section]],
    'moderation_state' => MassModeration::PUBLISHED,
  ]);

  $this->assertFalse(_mass_validation_check_paragraph_bundle($service_page, 'org_events'));
}
```

Minor: the code comment says "after migration" — direct placement of `org_events` on `field_service_sections` is just what the field's `target_bundles` allows, not a migration artifact. Worth rewording while you're in there.

dstorozhuk (approved):

PR #3428, opened 2026-06-25
DP-47847: entity usage stale fix

joeg8612:
**Description:**

DP-****

**To Test:**

---


docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
```diff
@@ -164,3 +166,43 @@ function _mass_entity_usage_get_nested_paragraph_ids(array $paragraph_ids) {
...
+  $candidate_ids = _mass_entity_usage_get_nested_paragraph_ids(array_map('intval', $root_ids));
+  $stale_ids = array_diff($candidate_ids, array_keys($current_ids));
+  if (!$stale_ids) {
+    return;
+  }
+
+  $usage = \Drupal::service('entity_usage.usage');
+  foreach ($stale_ids as $paragraph_id) {
+    $usage->deleteBySourceEntity((int) $paragraph_id, 'paragraph');
+  }
```
dstorozhuk: **C1 — expensive tree walk on every node save (blocker).**

`hook_entity_usage_queue_tracking_should_remove_usage()` runs for every node insert/update in the tracker queue. This cleanup is invoked unconditionally at the top and loads the node's whole paragraph tree via `$storage->load()` in a loop — twice.

Benchmarked on a production DB copy (median of 6 runs, caches reset), `develop` vs this PR on the same nodes:

| node | state | `develop` hook | this PR | delta |
|------|-------|---------------:|--------:|------:|
| 910501 | published | 0.6 ms | 662 ms | +661 ms |
| 417906 | published | 0.5 ms | 948 ms | +947 ms |
| 491016 | trash | 784 ms | 1281 ms | +497 ms |
| 464731 | unpublished | 471 ms | 897 ms | +426 ms |

On `develop` the tree is only walked for `unpublished`/`trash` nodes (inside the delete branch); published nodes return in ~0.5 ms. Because the cleanup runs before the moderation check, the walk now also runs on the published save path — a 0.5 ms hook becomes 600–950 ms.

The suggestion below adds a cheap guard: compare the node's current root paragraphs (field values, no entity loads) against the roots still parented in storage, bail out when nothing is detached, and expand only detached subtrees via `parent_id` back-refs instead of loading every paragraph. A paragraph has a single parent, so a detached root means its whole subtree is detached — coverage is preserved. Measured on the same nodes:

| node | state | this PR | guarded | speed-up |
|------|-------|--------:|--------:|---------:|
| 910501 | published | 662 ms | 8.6 ms | ~77x |
| 417906 | published | 948 ms | 13.6 ms | ~70x |
| 491016 | trash | 1281 ms | 26.6 ms | ~48x |
| 464731 | unpublished | 897 ms | 7.2 ms | ~124x |

Stale sets matched the original on 4/5 nodes; on the 5th the guarded version found one extra paragraph the forward-walk missed (a nested child replaced inside a surviving parent) — confirmed orphaned, so coverage is equal-or-better with no false positives.

```suggestion
  // Cheap guard: collect the node's current root paragraph ids straight from
  // the field values (no entity loads), and the root paragraphs still parented
  // to this node in storage (one query). A paragraph has a single parent, so a
  // root no longer referenced by the current revision means its whole subtree
  // is detached. If no root is detached there is nothing stale and we skip the
  // expensive tree walk entirely.
  $current_root_ids = [];
  foreach ($node->getFieldDefinitions() as $name => $definition) {
    if ($definition->getType() !== 'entity_reference_revisions') {
      continue;
    }
    if ($definition->getSetting('target_type') !== 'paragraph') {
      continue;
    }
    foreach ($node->get($name) as $item) {
      if ($item->target_id) {
        $current_root_ids[(int) $item->target_id] = TRUE;
      }
    }
  }

  $db_root_ids = \Drupal::database()
    ->select('paragraphs_item_field_data', 'pifd')
    ->fields('pifd', ['id'])
    ->condition('parent_type', 'node')
    ->condition('parent_id', $node->id())
    ->distinct()
    ->execute()
    ->fetchCol();

  $detached_roots = array_diff(array_map('intval', $db_root_ids), array_keys($current_root_ids));
  if (!$detached_roots) {
    return;
  }

  // Expand the detached subtrees via the storage table (parent_id back-refs)
  // instead of loading every paragraph entity, then delete their usage.
  $database = \Drupal::database();
  $stale_ids = [];
  $frontier = array_values($detached_roots);
  while ($frontier) {
    foreach ($frontier as $id) {
      $stale_ids[$id] = TRUE;
    }
    $children = $database->select('paragraphs_item_field_data', 'pifd')
      ->fields('pifd', ['id'])
      ->condition('parent_type', 'paragraph')
      ->condition('parent_id', $frontier, 'IN')
      ->distinct()
      ->execute()
      ->fetchCol();
    $frontier = array_diff(array_map('intval', $children), array_keys($stale_ids));
  }

  $usage = \Drupal::service('entity_usage.usage');
  foreach (array_keys($stale_ids) as $paragraph_id) {
    $usage->deleteBySourceEntity((int) $paragraph_id, 'paragraph');
  }
```

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
```diff
@@ -62,6 +62,8 @@ function mass_entity_usage_entity_usage_queue_tracking_should_remove_usage($enti
     return FALSE;
   }
 
+  _mass_entity_usage_delete_stale_paragraph_usage($entity);
```
dstorozhuk: **I1 — destructive side-effect inside a boolean predicate.**

`..._should_remove_usage()` is a predicate; the queue worker aggregates implementations via `in_array(TRUE, $results, TRUE)`. Putting an unconditional delete here is fragile — verified behaviourally that on a published node the hook returns `FALSE` yet still deletes the stale rows. If this hook is later made conditional/cached or a second implementation is added, the cleanup silently stops. Move it to a dedicated node-lifecycle hook, or gate it (the C1 guard already does most of this).

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
```diff
@@ -164,3 +166,43 @@ function _mass_entity_usage_get_nested_paragraph_ids(array $paragraph_ids) {
...
+
+  $candidate_ids = _mass_entity_usage_get_nested_paragraph_ids(array_map('intval', $root_ids));
+  $stale_ids = array_diff($candidate_ids, array_keys($current_ids));
+  if (!$stale_ids) {
+    return;
+  }
+
+  $usage = \Drupal::service('entity_usage.usage');
+  foreach ($stale_ids as $paragraph_id) {
+    $usage->deleteBySourceEntity((int) $paragraph_id, 'paragraph');
```
dstorozhuk: **I3 — deletions are silent.**

The contrib worker logs every usage deletion; this loop calls `deleteBySourceEntity()` with no log line. If it ever over-deletes there is no audit trail. Log the deleted paragraph ids (count/ids).

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
```diff
@@ -164,3 +166,43 @@ function _mass_entity_usage_get_nested_paragraph_ids(array $paragraph_ids) {
...
+ * When a node revision replaces or removes a paragraph tree, older paragraph
+ * entities can remain in storage long enough for their entity_usage rows to
+ * persist. Compare all paragraph trees currently parented to the node against
+ * the node's active paragraph tree and delete usage rows for detached
+ * paragraphs.
+ *
+ * @param \Drupal\Core\Entity\EntityInterface $node
+ *   The node currently being processed by the usage tracker queue.
+ */
+function _mass_entity_usage_delete_stale_paragraph_usage($node) {
```
dstorozhuk: **Structure — prefer an OOP `#[Hook]` class.**

Since Drupal 11.1, hooks should be implemented as `#[Hook]` attribute methods on a class in the module's `Hook` namespace (`src/Hook/`), registered as an autowired service, rather than procedural functions plus `\Drupal::` static calls. Worth extracting the tree-walk helpers — notably `_mass_entity_usage_get_nested_paragraph_ids()` — and these hook implementations into a hook class, injecting `entity_usage.usage`, the database and the entity type manager via the constructor; that makes the logic unit-testable and mockable instead of relying on `\Drupal::service()` lookups.

Refs: [Support for object oriented hook implementations](https://www.drupal.org/node/3442349) · [Understanding hooks](https://www.drupal.org/docs/develop/creating-modules/understanding-hooks)

dstorozhuk (changes requested):

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
tfleming-ma: I applied these changes and then expanded on them. See `deleteStaleParagraphUsage`. You can see the changes to these suggestions in this commit prior to moving the helper function over to a Hook class here: https://github.com/massgov/openmass/pull/3428/changes/8dc58e1c65623c993a66c99a6863cd9c739bdafa

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
tfleming-ma: This helper function is now a method called `deleteStaleParagraphUsage`. This method runs on hook_entity_update after a check that the default revision is being updated.

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
tfleming-ma: Logging added. See `deleteStaleParagraphUsage`

docroot/modules/custom/mass_entity_usage/mass_entity_usage.module
tfleming-ma: I moved all hook implementations over to Hook classes. The changes not directly relevant to the fix were moved to `MassEntityUsageHooks`, the .module file was deleted, and the services.yml parameter `mass_entity_usage.skip_procedural_hook_scan: true` was added. I also added a test for hook_queue_info_alter.

dstorozhuk (approved):
@tfleming-ma please   change  core_version_requirement: ^10 || ^11 to  core_version_requirement: ^11.1 .

PR #3444, opened 2026-07-08
DP-47869-Guide-Page-Rich-Text-Field-Link-Autocomplete-Not-Populating

jarielrrz:
Fix Guide Page rich text Linkit autocomplete regression by removing the `mayflower/global` dependency from the Message box CKEditor5 library.


docroot/modules/custom/mayflower/mayflower.module
```diff
@@ -111,6 +111,16 @@ function mayflower_library_info_build() {
       'mayflower/modernizr',
     ],
   ];
+  // CSS-only variant of mayflower/global for admin/editor contexts that need
+  // Mayflower styles without the vendor JS (which replaces Drupal's jQuery).
+  $libraries['global_css'] = [
+    'version' => $mf_installed_version . '-' . $mf_installed_reference,
```
dstorozhuk, line 117: @jarielrrz  please add  `'version' => $mf_installed_version . '-' . $mf_installed_reference,`

dstorozhuk (changes requested):

docroot/modules/custom/mayflower/mayflower.module
jarielrrz: This line is already present. Version is the same way as global / layout_paragraphs.

dstorozhuk (approved):

PR #3452, opened 2026-07-16
DP-45938-Restrict-“Map”-Field-on-Organization-Pages-to-Location-Content-Types-Only

jarielrrz:
Restrict Map fields on Organization and Service pages to Location content.


docroot/modules/custom/mass_validation/mass_validation.module
```diff
@@ -314,6 +314,13 @@ function _mass_validation_field_organizations_validate(array &$form, FormStateIn
   }
 }
 
+/**
+ * Validates Map field references on layout paragraph component forms.
+ */
+function _mass_validation_validate_location_references_layout_paragraphs(array &$form, FormStateInterface $form_state): void {
+  \Drupal::service('mass_validation.location_reference_validator')->validateLayoutParagraphForm($form_state);
+}
```
dstorozhuk: The whole validator layer could shrink to re-enabling the widget's own reference validation — the field's selection handler (the `location_pages` view) already restricts to Location; it's just that core's `EntityReferenceAutocompleteWidget` opts out of validating pasted IDs against it (`'#validate_reference' => FALSE`). Re-enabling it covers the org_page classic subform, the LP component form, and the `quick_node_clone` form in one place, and also rejects nonexistent IDs.

Since new code in this repo uses OOP `#[Hook]` classes, this would be a new `docroot/modules/custom/mass_validation/src/Hook/MassValidationWidgetHooks.php` instead of anything in the `.module` file:

```php
<?php

namespace Drupal\mass_validation\Hook;

use Drupal\Core\Form\FormStateInterface;
use Drupal\Core\Hook\Attribute\Hook;

/**
 * OOP hook implementations for DP-45938 Map field reference validation.
 */
class MassValidationWidgetHooks {

  /**
   * Restores selection-handler validation for the Map field autocomplete.
   *
   * Core's EntityReferenceAutocompleteWidget disables it, relying on the
   * ValidReference constraint — which never runs for paragraphs edited via
   * the classic inline widget.
   */
  #[Hook('field_widget_single_element_entity_reference_autocomplete_form_alter')]
  public function mapFieldValidateReference(array &$element, FormStateInterface $form_state, array $context): void {
    if ($context['items']->getFieldDefinition()->getName() === 'field_org_ref_locations') {
      $element['target_id']['#validate_reference'] = TRUE;
    }
  }

}
```

If we go this route, `LocationReferenceValidator`, `mass_validation.services.yml`, this wrapper function and the other two wiring points can all be dropped. Trade-offs: core's generic error message instead of the custom one, and unpublished/address-less locations are rejected too (the view filters on those — consistent with what the autocomplete actually suggests).

dstorozhuk (changes requested):

docroot/modules/custom/mass_validation/tests/src/ExistingSite/LocationReferenceValidationTest.php
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  /**
+   * Assert the service page Layout Paragraphs path rejects non-locations.
+   *
+   * Service pages edit the Map field through the Layout Paragraphs component
+   * form, whose #validate handler calls validateLayoutParagraphForm(). This
+   * exercises that path directly with both invalid and valid references.
+   */
+  public function testServicePageLayoutParagraphsPathRejectsNonLocationReference(): void {
```
dstorozhuk: This test doesn't actually exercise the service-page path it's named after — it hand-builds a `FormState` and calls the validator service directly, so it keeps passing even if the `case 'org_locations':` wiring in `mass_validation_form_layout_paragraphs_component_form_alter()` is deleted or the LP component form's value structure changes. The only wiring it verifies is the `services.yml` registration.

Options:
- Upgrade it to a real LP component-form test — there's a ready-made template in this module: `tests/src/ExistingSiteJavascript/CollectionSearchValidationLayoutParagraphsTest.php` (drives the LP modal on a service_page, submits an invalid value, asserts the error in the dialog).
- Or rename it (e.g. `testValidatorServiceRejectsNonLocationReference`) so it doesn't read as end-to-end coverage.
- If the `#[Hook]` / `#validate_reference` approach is adopted instead, this test can simply be dropped — the org_page browser test above already proves the widget-level validation fires.

docroot/modules/custom/mass_validation/tests/src/ExistingSite/LocationReferenceValidationTest.php
```diff
@@ -0,0 +1,100 @@
...
+    $this->visit($org_node->toUrl()->toString() . '/edit');
+    $this->assertEquals(200, $this->getSession()->getStatusCode());
+    $page = $this->getSession()->getPage();
+    $page->selectFieldOption('edit-moderation-state-0-state', 'Unpublished');
+    $page->pressButton('edit-field-organization-sections-add-more-add-more-button-org-section-long-form');
+    $page->pressButton('edit-field-organization-sections-0-subform-field-section-long-form-content-add-more-add-more-button-org-locations');
+    $page->fillField('edit-field-organization-sections-0-subform-field-section-long-form-content-0-subform-field-org-ref-locations-0-target-id', $contact->label() . ' (' . $contact->id() . ')');
+    $page->pressButton('edit-submit');
+
+    $validation_text = 'Only location pages can be referenced in this field.';
```
dstorozhuk: The org_page browser test is the valuable one — please keep it whichever way the implementation goes: it exercises the real nested-paragraph form and proves invalid references are blocked (and, via `failOnLoggedErrors()`, that nothing fatals on submit).

Two notes:
- If the `#validate_reference` / `#[Hook]` approach is adopted, this assertion needs to target core's message instead (`The referenced entity (node: X) does not exist.` / `This entity ... cannot be referenced.`).
- Worth adding a second submit with a valid Location reference asserting the message is absent, so the test also guards against false positives (validator suddenly rejecting legitimate locations).

dstorozhuk (changes requested):
Follow-up on the tests (see also the `#[Hook]` suggestion above).

dstorozhuk (approved):

PR #3455, opened 2026-07-16
DP-46465: Regression tests for regenerate decision tree and worker resume

dstorozhuk:
## Objective

Add the missing regression tests for the two places in #3435 where a future change could silently recreate the DP-46437 production incident (an accidental full rebuild wipes all usage data and takes 1.5-2 days to recover). The command decision tree and the batch worker's wipe decision had no coverage; both are pinned now.

**Jira:** [DP-46465](https://massgov.atlassian.net/browse/DP-46465)

## Why these tests are needed

1. **The command decision tree is the ticket.** The confirmation, the `--force`/`-y` bypass, resume-over-fresh, and the recent-completion exit all live in `MassEntityUsageCommands::recreate()`, and none of it was exercised by tests. Any future edit to that if/elseif ladder could silently re-enable the accidental full rebuild. The new unit tests pin every branch, including the order-of-operations guard: a declined/non-interactive `--reset` must abort **before** `beginFreshEnqueueRun()` clears saved progress and the queue.

2. **The 24-hour-timeout bug already shipped once.** Before 34785ec, an expired completion flag was re-stamped on every bare invocation, so a new run was permanently impossible without `--reset`. `testExpiredCompletionFlagStartsFreshRunWithoutRestamping` fails on the pre-fix code (verified) and passes after it — that bug can't come back unnoticed.

3. **The worker's resume-vs-wipe decision is the only code path that deletes usage data.** `queueSourcesBatchWorker()` decides between restoring saved progress and calling `bulkDeleteSources()`. A regression there deletes all usage records while claiming to resume. The new ExistingSite tests prove: resume keeps existing records and continues from the saved position; stale-run progress is discarded and records are wiped. The pair is self-validating — the stale case proves the sentinel row is really deletable, the resume case proves it survives.

The worker tests use the `redirect` entity type and snapshot/restore state and usage rows, so they cannot destroy real node/paragraph usage data even if run against a shared environment.

## Test results

- `MassEntityUsageCommandsDecisionTest`: 8 tests — green (and the timeout-regression test verified to fail on pre-34785ec code).
- `UsageRegenerateWorkerResumeTest`: 2 tests — green.
- Full `mass_entity_usage` suite: 42 tests / 127 assertions — green.
- phpcs clean.

[DP-46465]: https://massgov.atlassian.net/browse/DP-46465?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changelogs/DP-46465.yml
+++ b/changelogs/DP-46465.yml
@@ -39,3 +39,5 @@
 Added:
   - description: Add confirmation to entity usage regenerate so that it will confirm before starting a new run.
     issue: DP-46465
+  - description: Add regression tests for the usage-regenerate decision tree and the batch worker resume logic.
+    issue: DP-46465
--- a/docroot/modules/custom/mass_entity_usage/tests/src/ExistingSite/UsageRegenerateWorkerResumeTest.php
+++ b/docroot/modules/custom/mass_entity_usage/tests/src/ExistingSite/UsageRegenerateWorkerResumeTest.php
@@ -1,0 +1,189 @@
+<?php
+
+namespace Drupal\Tests\mass_entity_usage\ExistingSite;
+
+use Drupal\mass_entity_usage\EntityUsageQueueBatchManager;
+use MassGov\Dtt\MassExistingSiteBase;
+
+/**
+ * Tests the batch worker's resume-vs-wipe decision.
+ *
+ * The worker (queueSourcesBatchWorker) is the only code path deleting usage data
+ * (bulkDeleteSources). On resume — saved progress belonging to the current
+ * run — it must restore the sandbox and keep existing records; only a fresh
+ * start (no progress, or progress from a stale run) may wipe. A regression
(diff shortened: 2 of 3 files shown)


joeg8612:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against c3d2193006fcf9686fea3d7e27619dfc459cdc82
</p>

tfleming-ma:
Looks good to me, but getting phpunit errors related to config. Merging into main ticket and will resolve issue there.

PR #3456, opened 2026-07-17
DP-48112-Update-Editoria11y-Accessibility-Report-to-Display-Current-Issue-Counts

jarielrrz:


conf/drupal/config/views.view.accessibility_report_for_authors.yml
```diff
@@ -555,12 +555,12 @@ display:
           multi_type: separator
           separator: ', '
           field_api_classes: false
-        content_count:
-          id: content_count
-          table: ed11y_result
-          field: content_count
-          relationship: none
-          group_type: sum
+        content_results:
```
dstorozhuk: **"Issue count" no longer respects the exposed "Alert type, linked" filter.**

`ed11y_page.content_results` is a per-page total, independent of which `ed11y_result` rows survive the exposed `result_key` filter. The old `SUM(ed11y_result.content_count)` summed only the selected alert types. Measured on a prod DB copy: nid 973951 has `content_results = 14` while its per-type rows are `IMAGE_DECORATIVE = 3`, `HEADING_SKIPPED_LEVEL = 1`, `LINK_URL = 2`, … — filtering the report to a single alert type used to show **3**, now it will show **14**. The `DESC` sort on this column is wrong in the same way, and `data_export_1` (CSV) inherits both.

Suggested fix: keep the field/sort as `ed11y_result.content_count` / `group_type: sum` and let the new query-alter WHERE clauses handle the staleness fix on their own — I verified that is sufficient: stale rows are already eliminated by the path predicate, and for the single surviving page row `content_results` always equals `SUM(content_count)`.

Minor, related: the new `content_results > 0` filter is currently inert — every page with `content_results = 0` has no `ed11y_result` rows at all (Editoria11y purges page + result rows on a clean rescan), and `ed11y_result` is the base table, so those pages can never appear regardless. Harmless belt-and-braces, just not what fixes the blue-checkmark case.

docroot/modules/custom/mass_views/mass_views.module
```diff
@@ -427,6 +431,36 @@ function mass_views_views_query_alter(ViewExecutable $view, QueryPluginBase $que
...
+      break;
+    }
+  }
+
+  if (!$page_alias) {
+    return;
+  }
+
+  // Match current alias (or /node/{nid}); BINARY avoids collation mismatches.
+  $current_path = "COALESCE((SELECT pa.alias FROM {path_alias} pa WHERE pa.path = CONCAT('/node/', $page_alias.entity_id) AND pa.status = 1 AND (pa.langcode = $page_alias.page_language OR pa.langcode = 'und') ORDER BY CASE WHEN pa.langcode = $page_alias.page_language THEN 0 ELSE 1 END, pa.id DESC LIMIT 1), CONCAT('/node/', $page_alias.entity_id))";
```
dstorozhuk, line 469: The fix mechanism itself checks out — all four nodes from the ticket (663676, 124226, 195981, 333516) have their single `ed11y_page` row stored under an outdated alias, and both new predicates independently exclude them. A few points though, all verified against a local prod DB copy:

1. **Silent no-op fallback.** If `ed11y_page` is ever missing from the table queue (someone edits the view in Views UI and removes the `editoria11y_result_to_page` relationship, a partial config import, etc.), the `!$page_alias` early return silently disables the staleness filtering and the exact DP-48112 bug comes back with zero signal. Worth a `\Drupal::logger('mass_views')->warning()` on that path so the regression shows up in watchdog/New Relic instead of a re-filed ticket.

2. **Scope: ~900 nodes silently drop off the report.** The path-equality predicate removes 896 distinct nodes (8,974 → 8,078 on the prod copy), at least 313 of them published pages with unresolved issues — their URL changed (mostly pathauto event-pattern renames) and they were never rescanned at the new URL. Their issues may still exist; they are now invisible until someone re-views the page, which for an author is indistinguishable from "fixed". Probably an acceptable trade-off, but it deserves an explicit product sign-off, a mention in the PR description, and ideally a fourth bullet in the report header's disclaimer.

3. **Latent multilingual / query-string mismatch.** Editoria11y stores `page_path` from `$entity->toUrl('canonical')->toString()` plus preserved query params — that includes the language prefix (`/es/<alias>`), while `path_alias.alias` never contains a prefix or query string. Today every row in the DB is `en` with no prefix, so nothing breaks — but the first scan of a translated node will silently never match and the page will never appear on the report. Worth a code comment at minimum, or prepending the language prefix when `page_language` is not the default.

4. **The `updated = (SELECT MAX(...))` subquery is provably redundant today.** `ed11y_page` has a unique key on `(page_path, page_language)` and Editoria11y purges stale-path rows on rescan, so restricted to the current path it can only disambiguate rows differing by language — and there are none (all 13,977 rows are `en`). It doubles the correlated-subquery cost per row, which the batch CSV export re-pays on every batch. Consider dropping it, or add a comment explaining the case it defends against.

5. **`BINARY` — right call, but the comment undersells it and the syntax is deprecated.** The cast is genuinely required: `ed11y_page.page_path` and `path_alias.alias` have different collations (verified on both a local copy and feature4), and a plain `=` throws error 1267 "Illegal mix of collations". However `BINARY expr` is deprecated since MySQL 8.0.27 — `CAST(... AS BINARY)` (or an explicit `COLLATE utf8mb4_bin`) is the forward-compatible spelling. Also suggest naming the two collations in the comment; on a fresh local install with uniform collation someone could remove the cast and see everything pass.

6. **No test coverage.** `mass_views` has an ExistingSite suite, and this view has never had any. Two cheap tests would lock this in: a smoke test (`drupalGet()` of the report — the base class fails on any logged SQL error, so it catches contrib schema drift against this raw SQL for free) and a regression test seeding two `ed11y_page` rows (stale alias with count 7, current alias with count 2) asserting the node shows once with count 2. `mass_redirect_normalizer/tests/src/ExistingSite/ReportViewTest.php` is a ready-made template for the seeding + `Views::getView()->execute()` pattern.

changelogs/DP-48112.yml
```diff
@@ -0,0 +1,3 @@
+Fixed:
+  - description: Accessibility report for authors now shows current Editoria11y issue counts and ignores stale scans from old page URLs.
```
dstorozhuk: Nit: the description does not mention the newly added `content_results > 0` filter — pages whose latest scan found zero issues are now hidden from the report, which is a user-visible change. Suggested wording: "Accessibility report for authors now shows current Editoria11y issue counts, ignores stale scans from old page URLs, and hides pages with no currently detected issues."

docroot/modules/custom/mass_views/src/Plugin/views/filter/OrgFilter.php
```diff
@@ -36,22 +36,25 @@ public function valueForm(&$form, FormStateInterface $form_state) {
...
+        $snippet = "$nid_alias.nid NOT IN ($p_nid) AND NOT EXISTS (SELECT 1 FROM {node__field_organizations} nfo WHERE nfo.entity_id = $nid_alias.nid AND nfo.field_organizations_target_id IN ($p_org))";
+      }
+      else {
+        $snippet = "$nid_alias.nid IN ($p_nid) OR EXISTS (SELECT 1 FROM {node__field_organizations} nfo WHERE nfo.entity_id = $nid_alias.nid AND nfo.field_organizations_target_id IN ($p_org))";
+      }
+      $this->query->addWhereExpression($this->options['group'], $snippet, [
+        $p_nid => $value,
+        $p_org => $value,
+      ]);
     }
```
dstorozhuk, line 81: **Blocking — this is the ~2.4x regression on All Content described in the review summary.**

The `INNER JOIN` was doing real work here: it let MySQL start from the small `node__field_organizations` row set via `field_organizations_target_id` and widen from there. A correlated `EXISTS` in the `WHERE` clause cannot drive the plan, so the pager count query goes 268 ms -> ~1200 ms and the page as a whole 590 ms -> ~1450 ms.

The fan-out you are fixing is real, but it comes from joining the *raw* delta table. Joining a **de-duplicated** row set keeps both properties — no fan-out, and the join still drives the plan. Measured with the code below: All Content 596–703 ms (develop: 568–704 ms), raw SQL 327–371 ms, count query 255–280 ms, identical `total = 9387`; the report filtered by two orgs returns the same un-inflated count as your version (62), at 138 ms vs 325 ms.

`'table formula'` has to be a `SelectInterface`, not a string — `Select::addJoin()` runs a plain string through `escapeTable()` and mangles it into `Identifier name 'SELECTDISTINCTentity_id...' is too long`. I hit that first; the version below is what actually runs.

```suggestion
    // ONLY add the conditions if we have a value to filter on.
    if ($value = $this->getValue()) {
      $nid_alias = $this->query->ensureTable('node_field_data', $this->relationship);

      if ($this->operator === '!=') {
        $p_nid = $this->placeholder() . '[]';
        $p_org = $this->placeholder() . '[]';
        $snippet = "$nid_alias.nid NOT IN ($p_nid) AND NOT EXISTS (SELECT 1 FROM {node__field_organizations} nfo WHERE nfo.entity_id = $nid_alias.nid AND nfo.deleted = 0 AND nfo.field_organizations_target_id IN ($p_org))";
        $this->query->addWhereExpression($this->options['group'], $snippet, [
          $p_nid => $value,
          $p_org => $value,
        ]);
        return;
      }

      // Join a de-duplicated set of node ids instead of the raw field table.
      // node__field_organizations has one row per org delta, so joining it
      // directly fans out rows and inflates SUM() fields (e.g. accessibility
      // report issue counts). SELECT DISTINCT removes the fan-out while the
      // INNER JOIN keeps the selectivity that a correlated EXISTS loses.
      $database = \Drupal::database();
      $subquery = $database->select('node__field_organizations', 'nfo');
      $subquery->addField('nfo', 'entity_id');
      $subquery->condition('nfo.deleted', 0);
      $subquery->condition('nfo.field_organizations_target_id', $value, 'IN');
      $subquery->distinct();
      // An org_page matches itself, even with no field_organizations value.
      $self = $database->select('node_field_data', 'nfd');
      $self->addField('nfd', 'nid');
      $self->condition('nfd.nid', $value, 'IN');
      $subquery->union($self);

      $join = \Drupal::service('plugin.manager.views.join')->createInstance('standard', [
        'table' => 'node__field_organizations',
        'table formula' => $subquery,
        'field' => 'entity_id',
        'left_table' => 'node_field_data',
        'left_field' => 'nid',
        'type' => 'INNER',
      ]);
      $this->query->addTable('node__field_organizations', $this->relationship, $join, 'org_set');
    }
```

Two details in there worth calling out: `addTable()` (not `ensureTable()`) with an explicit `org_set` alias, so the filter never collides with a `node__field_organizations` join the view itself adds for an Organization field; and `nfo.deleted = 0`, which Views used to add to the join and the `EXISTS` version dropped (harmless today — there are no `deleted = 1` rows, Drupal moves those to `field_deleted_data_*` — but keeps parity).

dstorozhuk (commented):
Second review round. All eight items from round 1 are addressed — fallback logging, the 4th disclaimer bullet, the multilingual comment, the redundant `MAX(updated)` subquery dropped, `CAST(... AS BINARY)` with the collation names spelled out, tests, the field reverted to `SUM(ed11y_result.content_count)`, and the changelog rewrite. I re-verified the staleness fix on a prod-like local copy: filtered to org 67286 the summed issue count drops 163 -> 111, and both new tests pass locally (2 tests, 8 assertions, 14.5s).

Two things on the org-filter change that is new since round 1 — one of them blocking.

### Blocking: `OrgFilter` costs All Content ~2.4x

Replacing the `INNER JOIN` with a correlated `EXISTS` removes the selectivity the planner was using (start from the small org row set, then widen) and forces a scan. Measured on a local prod copy (139,748 nodes / 187,108 org rows), org 67286 (12,729 nodes), views cache disabled, 5 runs:

| | develop | this PR | suggestion below |
|---|---|---|---|
| All Content (`content:page_1`, Views execute) | 568–704 ms | **1309–1624 ms** | 596–703 ms |
| the same SQL, raw mysql | 371 ms | 661–794 ms | 327–371 ms |
| pager count query | 268 ms | **1077–1297 ms** | 255–280 ms |

The result set is identical in all three (`total = 9387`, same 50 rows) — for All Content the change fixes nothing (fan-out there is absorbed by `DISTINCT`; only the aggregating report was broken by it) and only costs time.

`mass_views_node_org_filter` is configured on **34 views**, including `content` (All Content), `my_content` (My Content) and `advancedsearch`, so per CLAUDE.md this needs the perf gate in `docs/performance.md` before merge. There is a measured alternative inline below.

### `OrgTopParentFilter` is the opposite — keep it

There the old join was a `LEFT JOIN`, which was the slow one. Advanced Search filtered by a top-level org: **4876–7844 ms on develop -> 1987–4589 ms with this PR**. Worth keeping as is.

The report itself goes 82–109 ms -> 200–211 ms; that is the price of the correlated `path_alias` subquery and is fine for a 63-row report.

### Side note, in the PR's favour

On develop, picking **more than one** organization in the filter throws `SQLSTATE[42000]` (`nid = (:p_0, :p_1)` — an `=` against an array placeholder). This PR fixes that as a side effect. Good catch; worth a line in the changelog.

Everything above was measured locally against a sanitized prod copy, and the suggested replacement is `phpcs -n` clean and passes the test class.

docroot/modules/custom/mass_views/src/Plugin/views/filter/OrgTopParentFilter.php
```diff
@@ -58,16 +58,19 @@ public function query() {
...
+    // (e.g. accessibility report issue counts: 252 issues x 2 orgs = 504).
     $or_group = $this->query->setWhereGroup('OR');
     $this->query->addWhere($or_group, "$nid_alias.nid", $org_ids, 'IN');
-    $this->query->addWhere($or_group, "$org_table_alias.field_organizations_target_id", $org_ids, 'IN');
+    $placeholder = $this->placeholder() . '[]';
+    $this->query->addWhereExpression(
+      $or_group,
+      "EXISTS (SELECT 1 FROM {node__field_organizations} nfo WHERE nfo.entity_id = $nid_alias.nid AND nfo.field_organizations_target_id IN ($placeholder))",
+      [$placeholder => $org_ids]
+    );
```
dstorozhuk, line 73: Not a problem — flagging it so it does not get reverted along with the other filter. Here the old join was a `LEFT JOIN`, which was the expensive one, so `EXISTS` is a straight win: Advanced Search filtered by a top-level org goes **4876–7844 ms on develop -> 1987–4589 ms** with this change (same result set), measured on a prod copy with 12,729 nodes under the selected org.

Only nit, for parity with the join Views used to build: add `nfo.deleted = 0` to the subquery.

docroot/modules/custom/mass_views/tests/src/ExistingSite/AccessibilityReportForAuthorsTest.php
```diff
@@ -0,0 +1,189 @@
...
+      $database = \Drupal::database();
+      $database->delete('ed11y_result')
+        ->condition('pid', $this->seededPids, 'IN')
+        ->execute();
+      $database->delete('ed11y_page')
+        ->condition('pid', $this->seededPids, 'IN')
+        ->execute();
+    }
+    parent::tearDown();
+  }
```
dstorozhuk, line 58: The test mutates the global request (line 116) and never puts it back. ExistingSite tests share one PHP process, so `status=1` leaks into every test that runs after this one — anything relying on `input_required_on_request`, or simply on an empty query string, can start behaving differently depending on test order. Cheapest fix is to restore it in `tearDown()`:

```suggestion
  /**
   * Pids inserted by this test, cleaned up in tearDown.
   *
   * @var int[]
   */
  private array $seededPids = [];

  /**
   * Original 'status' query parameter, restored in tearDown.
   */
  private ?string $originalStatusInput = NULL;

  /**
   * Whether the request carried a 'status' query parameter to begin with.
   */
  private bool $hadStatusInput = FALSE;

  /**
   * {@inheritdoc}
   */
  protected function tearDown(): void {
    $query = \Drupal::request()->query;
    if ($this->hadStatusInput) {
      $query->set('status', $this->originalStatusInput);
    }
    else {
      $query->remove('status');
    }
    if ($this->seededPids) {
      $database = \Drupal::database();
      $database->delete('ed11y_result')
        ->condition('pid', $this->seededPids, 'IN')
        ->execute();
      $database->delete('ed11y_page')
        ->condition('pid', $this->seededPids, 'IN')
        ->execute();
    }
    parent::tearDown();
  }
```

docroot/modules/custom/mass_views/tests/src/ExistingSite/AccessibilityReportForAuthorsTest.php
```diff
@@ -0,0 +1,189 @@
...
+      'page_title' => $title,
+      'content_results' => 2,
+      'dev_results' => 0,
+      'updated' => $now,
+    ]);
+    $this->insertEd11yResult($current_pid, 'LINK_URL', 2, $now);
+
+    // input_required_on_request only builds the query when the HTTP request
+    // has query parameters (not just setExposedInput).
+    \Drupal::request()->query->set('status', '1');
```
dstorozhuk: Pairs with the `tearDown()` suggestion above — record what was there before overwriting it:

```suggestion
    // input_required_on_request only builds the query when the HTTP request
    // has query parameters (not just setExposedInput).
    $request_query = \Drupal::request()->query;
    $this->hadStatusInput = $request_query->has('status');
    $this->originalStatusInput = $this->hadStatusInput ? (string) $request_query->get('status') : NULL;
    $request_query->set('status', '1');
```

docroot/modules/custom/mass_views/tests/src/ExistingSite/AccessibilityReportForAuthorsTest.php
```diff
@@ -0,0 +1,189 @@
...
+      if (!$this->rowMatchesEntityId($row, $nid)) {
+        continue;
+      }
+      $matching[] = $this->rowContentCount($row);
+    }
+
+    $this->assertNotEmpty($view->result, 'View should return results when status filter is in the request.');
+    $this->assertCount(1, $matching, 'Node should appear once (stale alias scan ignored). Current alias: ' . $current_alias);
+    $this->assertSame(2, $matching[0], 'Issue count should come from the current-URL scan.');
+  }
```
dstorozhuk, line 154: **Recommended: a test for the double-count bug.** That is the regression QA actually found (node 663396 showing 504 instead of 252) and the one whose fix touches 34 reports, and right now nothing locks it in — the two tests here only cover the staleness fix.

I wrote and ran this one: it **fails on develop with `Failed asserting that 10 is identical to 5`** — the exact `252 x 2 orgs = 504` shape from the ticket — and **passes with this PR's filters** (and with the `OrgFilter` replacement suggested above). Runs in ~2s.

```suggestion
  }

  /**
   * Issue counts are not multiplied when a node has several organizations.
   */
  public function testIssueCountIsNotInflatedByMultipleOrganizations(): void {
    $parent_org = $this->createNode([
      'type' => 'org_page',
      'title' => 'DP-48112 parent org ' . $this->randomMachineName(6),
      'status' => 1,
      'moderation_state' => MassModeration::PUBLISHED,
    ]);
    $child_org = $this->createNode([
      'type' => 'org_page',
      'title' => 'DP-48112 child org ' . $this->randomMachineName(6),
      'status' => 1,
      'moderation_state' => MassModeration::PUBLISHED,
      'field_parent' => [['target_id' => $parent_org->id()]],
    ]);
    // A page tagged with two orgs from the same subtree: the pre-fix JOIN
    // returned one row per org delta, so SUM(content_count) doubled.
    $node = $this->createNode([
      'type' => 'service_page',
      'title' => 'DP-48112 multi-org ' . $this->randomMachineName(8),
      'status' => 1,
      'moderation_state' => MassModeration::PUBLISHED,
      'field_organizations' => [
        ['target_id' => $parent_org->id()],
        ['target_id' => $child_org->id()],
      ],
    ]);
    $nid = (int) $node->id();
    $now = \Drupal::time()->getRequestTime();

    $pid = $this->insertEd11yPage([
      'entity_id' => $nid,
      'entity_type' => 'Service',
      'route_name' => 'entity.node.canonical',
      'page_path' => \Drupal::service('path_alias.manager')->getAliasByPath('/node/' . $nid),
      'page_language' => 'en',
      'page_title' => $node->getTitle(),
      'content_results' => 5,
      'dev_results' => 0,
      'updated' => $now,
    ]);
    $this->insertEd11yResult($pid, 'LINK_URL', 5, $now);

    \Drupal::request()->query->set('status', '1');

    $view = Views::getView('accessibility_report_for_authors');
    $view->setDisplay('default');
    $view->setItemsPerPage(0);
    $view->initHandlers();
    $view->filter['node_parent_org_top_filter']->options['exposed'] = FALSE;
    $view->filter['node_parent_org_top_filter']->value = [$parent_org->id()];
    $view->execute();

    $matching = [];
    foreach ($view->result as $row) {
      if ($this->rowMatchesEntityId($row, $nid)) {
        $matching[] = $this->rowContentCount($row);
      }
    }

    $this->assertCount(1, $matching, 'A node tagged with two orgs in the subtree should appear once.');
    $this->assertSame(5, $matching[0], 'Issue count must not be multiplied by the number of organizations.');
  }
```

docroot/modules/custom/mass_views/src/Plugin/views/filter/OrgFilter.php
```diff
@@ -36,22 +36,48 @@ public function valueForm(&$form, FormStateInterface $form_state) {
...
-      $this->query->addWhereExpression($this->options['group'], $snippet, [$p1 => $value]);
+      $join = \Drupal::service('plugin.manager.views.join')->createInstance('standard', [
+        'table' => 'node__field_organizations',
+        'table formula' => $subquery,
+        'field' => 'entity_id',
+        'left_table' => 'node_field_data',
+        'left_field' => 'nid',
+        'type' => 'INNER',
+      ]);
+      $this->query->addTable('node__field_organizations', $this->relationship, $join, 'org_set');
```
dstorozhuk: **Blocking: the hard-coded `org_set` alias silently drops the second Organization filter in a display.**

`Sql::addTable()` → `queueTable()` (core `Sql.php:526-530`) returns early when the requested alias is already in the table queue — no error, no log; the second filter's subquery, with its own values, is simply discarded. This is live config, not a hypothetical: `parent_reports` `page_1` (`/admin/ma-dash/reports/parent-reports`) and `data_export_1` carry two instances of this plugin — `node_org_filter` ("Organization of Child", relationship none) and `node_org_filter_1` ("Organization of Parent", relationship `node__field_primary_parent`).

Measured on a prod DB copy with both filters set (child org 67286 + parent org 6661): develop returns **57** rows, the previous EXISTS version of this branch returns **57**, current HEAD returns **8911** — the parent org id never even makes it into the bound arguments. The page and its CSV export show child-only results while the UI shows both filters applied. (Single-filter behavior is byte-for-byte identical to develop — this only breaks when both are used together.)

Fix is one line: make the alias unique per handler instance, e.g. `'org_set_' . $this->options['id']` (handler ids are unique within a display), or pass `NULL` and let Views generate one. Please also add a regression assertion for the two-filter case: both filters set → the small intersection (57-row shape), not the child-only count.

Three smaller things in the same branch:
- `$nid_alias` (line 41) looks unused in the `=` path but is load-bearing — it forces `node_field_data` into the query so the hard-coded `'left_table' => 'node_field_data'` resolves. Either use it as the `left_table` or add a one-line comment; a future cleanup deleting it would be a landmine.
- The INNER JOIN is unconditionally AND'ed, while the `!=` branch respects `$this->options['group']`. No org filter sits in an OR filter group in any of the 35 views today, so this is latent — but worth a comment stating the join form assumes an AND group.
- The `getValue()` docblock still says `@return int|null`; it returns `int[]|null`, and multi-value support is now the headline of the changelog entry.

Verified good, for the record: the perf regression from the EXISTS round is gone — All Content is back to develop parity (351/398 ms vs develop's 374/421 ms; the EXISTS version was 735/997 ms) with identical result sets; multi-org selection now works in both operators (develop errors out); the fan-out fix holds under aggregation, and the UNION self-match arm behaves correctly for org_pages with and without `field_organizations`. Placeholder handling of the embedded subquery is safe (core compiles join subqueries against the outer query's placeholder counter).

docroot/modules/custom/mass_views/src/Plugin/views/filter/OrgTopParentFilter.php
```diff
@@ -58,16 +58,19 @@ public function query() {
...
+    // Use EXISTS instead of JOINing node__field_organizations. That table has
+    // one row per org delta; a JOIN fans out rows and inflates SUM() fields
+    // (e.g. accessibility report issue counts: 252 issues x 2 orgs = 504).
     $or_group = $this->query->setWhereGroup('OR');
     $this->query->addWhere($or_group, "$nid_alias.nid", $org_ids, 'IN');
-    $this->query->addWhere($or_group, "$org_table_alias.field_organizations_target_id", $org_ids, 'IN');
+    $placeholder = $this->placeholder() . '[]';
+    $this->query->addWhereExpression(
+      $or_group,
+      "EXISTS (SELECT 1 FROM {node__field_organizations} nfo WHERE nfo.entity_id = $nid_alias.nid AND nfo.deleted = 0 AND nfo.field_organizations_target_id IN ($placeholder))",
```
dstorozhuk, line 71: Non-blocking — filed [DP-48624](https://massgov.atlassian.net/browse/DP-48624) as a follow-up. This filter kept the `nid IN (...) OR EXISTS (...)` shape while `OrgFilter` moved to the de-duplicated INNER JOIN, and it is now the slowest of the three predicates: isolated predicate 129 ms vs 14 ms for the JOIN form (identical result sets); `/admin/advsearch/page` with the filter applied ~2.2 s vs ~0.4 s. Configured on 17 displays.

To be clear: this is still a big improvement over develop (the old LEFT JOIN shape measured 4.9–7.8 s on the same page), and QA's prod-vs-Tugboat numbers (10 s vs 2 s) corroborate that this PR already speeds the page up — hence follow-up, not blocker. The swap is mechanical: the `UNION` arm of the `org_set` subquery is exactly the `nid IN (org_ids)` half of this predicate.

The `deleted = 0` addition — thanks; all three predicates (org_set subquery, the `!=` branch, this EXISTS) are consistent now.

docroot/modules/custom/mass_views/tests/src/ExistingSite/AccessibilityReportForAuthorsTest.php
```diff
@@ -0,0 +1,304 @@
...
+
+  /**
+   * Issue counts are not multiplied when a node has several organizations.
+   */
+  public function testIssueCountIsNotInflatedByMultipleOrganizations(): void {
```
dstorozhuk, line 171: **Recommended: one more scenario in this test — two selected organizations.** As written, the direct `node_org_filter` block is not revert-sensitive: with a single selected org, develop's old JOIN kept only the matching delta row, so the SUM stayed 5 and both assertions pass even on the old buggy code. Selecting **two** orgs is the shape that (a) threw the SQL error on develop — the changelog's multi-org claim, currently untested — and (b) doubles the SUM on any naive raw-join fix. One block covers both:

```php
// Multiple selected organizations (develop threw an SQL error on
// `nid = (:p[])`, and a raw delta JOIN would double the SUM).
$view = Views::getView('accessibility_report_for_authors');
$view->setDisplay('default');
$view->setItemsPerPage(0);
$view->initHandlers();
$view->filter['node_org_filter']->options['exposed'] = FALSE;
$view->filter['node_org_filter']->operator = '=';
$view->filter['node_org_filter']->value = [
  ['target_id' => $parent_org->id()],
  ['target_id' => $child_org->id()],
];
$view->execute();
// ... same matching loop ...
$this->assertCount(1, $matching, 'Selecting two orgs must not error or duplicate the node.');
$this->assertSame(5, $matching[0], 'Two matching org deltas must not double the SUM.');
```

Fails on develop via the logged SQL error (the base class turns those into failures), fails on a raw-join fix via SUM = 10, passes on this PR. The top-parent scenario above is already revert-sensitive as-is (reverting the filter yields SUM = 10) — nice work on that one.

Also worth covering, non-blocking:
- the `!=` operator — user-reachable via `expose.use_operator` in `parent_reports`, and its semantics changed substantially (true exclusion; org-less nodes are now included);
- the org_page self-match `UNION` arm — an org_page with no `field_organizations` value must still match itself; a future "simplify away the odd UNION" refactor would silently drop org pages from every org-filtered listing.

Two minor hazards: `setItemsPerPage(0)` in the stale-alias test runs with no org constraint, so the result loop walks the entire aggregated ed11y dataset of the pulled DB — cheap to constrain by tagging the node with a fresh org_page the way this test does; and the manually created `PathAlias` is not passed to `markEntityForCleanup()`.

The `setStatusRequestQuery()`/`tearDown()` restore is correct and complete — verified there is no handler-state leak either (`Views::getView()` builds a fresh executable each call, bypassing the memoized one).

changelogs/DP-48112.yml
```diff
@@ -0,0 +1,3 @@
+Fixed:
+  - description: Accessibility report for authors now shows current Editoria11y issue counts, ignores stale scans from old page URLs, and no longer double-counts issues when filtering by organization. Organization filters also accept multiple organizations without SQL errors.
```
dstorozhuk: Nit: "Organization filters" (plural) slightly overstates it — only the direct Organization filter changed; the top-parent filter is single-select (`#tags => FALSE`) and never had the multi-value error. The entry also omits the other two real changes in this commit: restoring All Content performance (the EXISTS shape cost ~2.4x there) and the `deleted = 0` hardening.

docroot/modules/custom/mass_views/src/Plugin/views/filter/OrgFilter.php
```diff
@@ -36,30 +36,61 @@ public function valueForm(&$form, FormStateInterface $form_state) {
...
+        'table' => 'node__field_organizations',
+        'table formula' => $subquery,
+        'field' => 'entity_id',
+        'left_table' => $nid_alias,
+        'left_field' => 'nid',
+        'type' => 'INNER',
+      ]);
+      // Handler IDs are unique per display. A hard-coded alias lets
+      // Sql::queueTable() silently discard the second instance's subquery.
+      $this->query->addTable('node__field_organizations', $this->relationship, $join, 'org_set_' . $this->options['id']);
```
dstorozhuk, line 85: **Round 4 — the alias fix is verified, no blockers left in this file.**

Re-ran the round-3 scenario on a prod DB copy: child + parent org filters on `parent_reports` are back to the correct **57-row** intersection (was 8911 on the broken alias), with two independent joins (`org_set_node_org_filter`, `org_set_node_org_filter_1`) and correct ON clauses; single-filter results are byte-identical to develop. Also swept all 49 displays carrying this filter comparing `'left_table' => $nid_alias` against the previous literal: identical in 47, correct where it matters in 2 (the two parent_reports displays), worse in none — the alias is the API-correct input to `buildJoin()`/`getTableInfo()`, so this is the better form regardless. Alias length is fine too (longest is `org_set_node_org_filter_1`, 25 chars vs MySQL's 64 limit). phpcs clean.

Two tiny nits, take or leave:
- The comment above `addTable()` reads inverted now that the code is fixed: "A hard-coded alias **lets** `Sql::queueTable()` silently discard…" describes the current code as if it still hard-codes. "…**would let**…" reads right.
- `addTable()` can return `FALSE` when the relationship path can't be ensured, and the filter would then silently apply no restriction. Not reachable with any current config (checked all 49 displays), but a one-line guard would make a future config change loud instead of silent.

docroot/modules/custom/mass_views/src/Plugin/views/filter/OrgFilterMedia.php
```diff
@@ -98,10 +98,10 @@ public function query() {
   }
 
   /**
-   * Retrieve a single usable int value from the input value.
+   * Retrieve usable organization IDs from the input value.
    *
-   * @return int|null
-   *   The organization ID, or NULL.
+   * @return int[]|null
```
dstorozhuk, line 103: Heads-up while touching this class: `query()` above still has the multi-value bug this PR fixed on the node side — line 95 builds `field_organizations_target_id = :placeholder[]`, so selecting two **different** organizations in the exposed autocomplete (it's `#tags => TRUE`, a normal input) throws `SQLSTATE[21000] Cardinality violation` → 500. Verified over HTTP on `/admin/ma-dash/documents`, and reproduced on clean develop — **pre-existing, not a regression of this PR**. Filed as [DP-48656](https://massgov.atlassian.net/browse/DP-48656) (affects all 9 media views using this plugin).

Since this PR now documents `getValue()` as returning `int[]`, either fix it here in one line (`=` → `IN (...)`, same shape as `OrgFilter`'s `!=` branch) and claim it in the changelog, or leave it to DP-48656 — but then the changelog wording needs narrowing (see the changelog comment).

changelogs/DP-48112.yml
dstorozhuk: Please narrow "Organization filters also accept multiple organizations without SQL errors" — it's true for the node Organization filter (now test-backed), but false for the media one: `OrgFilterMedia` still 500s on two selected organizations (pre-existing, tracked in [DP-48656](https://massgov.atlassian.net/browse/DP-48656)). Suggested: "The content Organization filter also accepts multiple organizations without SQL errors."

Optional: the `!=` operator's behavior also changed meaningfully (it is now the exact complement of `=`; on `parent_reports` it excludes 8,911 rows where develop excluded 17 — a correctness improvement, but user-visible via the exposed operator dropdown) and isn't mentioned.

docroot/modules/custom/mass_views/tests/src/ExistingSite/OrgFilterTest.php
```diff
@@ -0,0 +1,132 @@
...
+      'type' => 'service_page',
+      'title' => $title,
+      'status' => 1,
+      'moderation_state' => MassModeration::PUBLISHED,
+    ];
+    if ($org) {
+      $values['field_organizations'] = [['target_id' => $org->id()]];
+    }
+    if ($parent) {
+      $values['field_primary_parent'] = [['target_id' => $parent->id()]];
```
dstorozhuk, line 108: Good test — verified it's genuinely revert-sensitive: on the broken-alias code the display returns the child-only set, so `assertNotContains($child_only_decoy)` fails on exactly the right line, and the three-way decoy setup also catches the symmetric case (parent filter winning instead).

Small robustness notes, none blocking:
- `resultNids()` scans row properties for the first `nid`/`*_nid` key — on this display the row also carries `node_field_data_node__field_primary_parent_nid`, and it only works because the base `nid` happens to come first in the SELECT list. `$row->nid` is deterministic (the base-field alias on this display) and simpler.
- `parentReportsView(): object` — `ViewExecutable` would be the honest return type.
- While this file is hot, two cheap additions would close the remaining gaps from the earlier rounds: a `!=` scenario (reachable in prod via the exposed operator dropdown on both parent_reports org filters, and its semantics changed substantially in this PR), and an org_page self-match check (the `union($self)` arm in `OrgFilter::query()` — deleting it today fails no test, but would silently drop org pages from every org-filtered listing).

docroot/modules/custom/mass_views/tests/src/ExistingSite/AccessibilityReportForAuthorsTest.php
```diff
@@ -0,0 +1,349 @@
...
+
+    $view = Views::getView('accessibility_report_for_authors');
+    $this->assertNotNull($view);
+    $view->setDisplay('default');
+    $view->setItemsPerPage(0);
+    $view->initHandlers();
+    $view->filter['node_org_filter']->options['exposed'] = FALSE;
+    $view->filter['node_org_filter']->operator = '=';
+    $view->filter['node_org_filter']->value = [['target_id' => $org->id()]];
+    $view->setExposedInput(['status' => '1']);
```
dstorozhuk: Nit: this `setExposedInput(['status' => '1'])` is a measured no-op — with the request mutated by `setStatusRequestQuery()` above, `getExposedInput()` auto-populates from the request, and the sibling test doesn't use it. Worth dropping for consistency so the next reader doesn't conclude it's the load-bearing part.

For the record (checked because develop's DP-47588 merge changed the input-required machinery): `setStatusRequestQuery()` is still the load-bearing piece — `InputRequiredFromRequest::exposedFilterApplied()` reads the raw request query, and with `setExposedInput()` alone the view executes but returns 0 rows. The comment above it is still accurate. The org constraint added to the stale-alias test and the `markEntityForCleanup()` for the PathAlias both look right — thanks.

PR #3457, opened 2026-07-19
DP-47963: Fix Content Performance report export timeout on large result sets

dstorozhuk:
**Description:**
Exporting the Content Performance report failed with an AJAX HTTP 503 whenever the report returned more than 1,000 rows, because the export rendered one batch of 1,000 rows per request and that request did not finish in time.

Changed the `views_data_export` batch size on the `data_export_1` display from 1,000 to 100 rows. Also turned off link rendering on the CSV fields (Title, Authored by, Collections, Reusable label, Organizations, Parent page, Nos per 1000) — the CSV strips tags, so those links never appeared in the output anyway.

The CSV output is unchanged.

**Jira:**
DP-47963

**To Test:**
- [ ] Go to `/admin/content/performance` and filter the report so it returns more than 1,000 rows (e.g. Organization including all suborganizations = Massachusetts Department of Revenue).
- [ ] Export the results — the batch progress bar runs to completion and the CSV downloads, with no 503 error.
- [ ] Open the CSV: it has one header row and every row from the report.

---


jarielrrz (approved):
LGTM

PR #3465, opened 2026-07-29
DP-46813-Review-media-entity-name-logic

jarielrrz:
Keep document media names in sync with field_title so authors no longer see old titles.


tfleming-ma:
Looks good to me!

PR #3481, opened 2026-08-18
DP-47935: Increased max cache tag limit for edge cache tag header logging

tfleming-ma:
**Description:**
Set max cache tag limit to reflect edge cache tag limit.

DP-47935

---

### DIFF
--- a/docroot/modules/custom/mass_caching/src/EventSubscriber/AkamaiHeaderCreationSubscriber.php
+++ b/docroot/modules/custom/mass_caching/src/EventSubscriber/AkamaiHeaderCreationSubscriber.php
@@ -17,7 +17,7 @@ class AkamaiHeaderCreationSubscriber implements EventSubscriberInterface {
   /**
    * Akamai Edge-Cache-Tag response header tag count limit.
    */
-  private const MAX_HEADER_TAGS = 128;
+  private const MAX_HEADER_TAGS = 1024;
 
   /**
    * Akamai cache tag formatter.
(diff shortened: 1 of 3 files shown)


dstorozhuk (approved):

docroot/modules/custom/mass_caching/src/EventSubscriber/AkamaiHeaderCreationSubscriber.php
```diff
@@ -17,7 +17,7 @@ class AkamaiHeaderCreationSubscriber implements EventSubscriberInterface {
   /**
    * Akamai Edge-Cache-Tag response header tag count limit.
    */
-  private const MAX_HEADER_TAGS = 128;
+  private const MAX_HEADER_TAGS = 1024;
```
dstorozhuk, line 20: @tfleming-ma  looks good to and can be merged. 
One small note, not merge blocker. 

Make MAX_HEADER_TAGS public and use in test, so next time when we need to changes the MAX_HEADER_TAGS - it will be done in one place. But is really up to you if you want to fix that now or keep it as is.

tfleming-ma: Good suggestion, but since this is causing noise on production, I'd rather handle as a follow-up ticket. The likelihood of this value changing again soon is low, and I was thinking we should also allow this logging to be enabled or disabled with an environment variable like we do with Acquia tag logging.

PR #3489, opened 2026-08-28
DP-48546: Preserve logs in Admin Audit Trail

tfleming-ma:
**Description:**

DP-48546

---


joeg8612:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 0a20bf677ec4e4558d7887e2da618f4a036f2d8c
</p>

dstorozhuk:
**Not all delete operations are kept forever.**

`excludePermanent()` protects rows with `operation = 'delete'` only, but `admin_audit_trail` 1.0.10 logs several delete operations under other names:

- node: `translation delete`
- taxonomy: `term delete`, `vocabulary delete`
- menu: `link delete`, `link translation delete` (safe, `type = menu` is permanent)
- user_roles: `role_deleted` (safe, `type = user_roles` is permanent)

So term deletes, vocabulary deletes and node translation deletes fall into the "anything else" bucket and are purged after 2 months, while the report says "Delete operations: forever".

Suggested fix: match deletes with `->condition('operation', '%delete%', 'NOT LIKE')` (or an explicit list) in `excludePermanent()`, and update the expectation in `testCronCleanupQueries`. The unit test mocks the query builder, so it cannot catch this on its own.

**How to verify manually (local):**

1. Delete a taxonomy term and a node translation, and delete a regular node. Check `/admin/reports/audit-trail`: you should see `term delete`, `translation delete` and `delete` rows.
2. Temporarily shorten the thresholds in `applyRetentionPolicy()` (e.g. `-2 months` -> `-1 minute`), or backdate the rows instead:
   ```sql
   UPDATE admin_audit_trail SET created = created - 86400*90 WHERE operation LIKE '%delete%';
   ```
3. Run `drush cron`.
4. Current branch: the `term delete` and `translation delete` rows are gone, only `delete` survives. After the fix all three should still be there.

docroot/modules/custom/mass_admin_audit_trail/src/Hook/MassAdminAuditTrailHooks.php
```diff
@@ -0,0 +1,142 @@
...
+      ],
+    ];
+  }
+
+  /**
+   * Excludes permanently retained records from a delete query.
+   */
+  private function excludePermanent(Delete $query): void {
+    $query
+      ->condition('operation', 'delete', '<>')
```
dstorozhuk, line 138: Contrib logs `term delete`, `vocabulary delete` and `translation delete` too, so exact match misses them.
```suggestion
      ->condition('operation', '%delete%', 'NOT LIKE')
```

docroot/modules/custom/mass_admin_audit_trail/tests/src/Unit/Hook/MassAdminAuditTrailHooksTest.php
```diff
@@ -0,0 +1,220 @@
...
+      $insert_update_query,
+      $two_month_query,
+    );
+
+    // Run cron against a fixed time so each cutoff is deterministic.
+    $this->hooks($database, $this->timeAt($now))->cron();
+
+    // Every query must preserve deletes and the protected event types.
+    $permanent_conditions = [
+      ['operation', 'delete', '<>'],
```
dstorozhuk, line 60:
```suggestion
      ['operation', '%delete%', 'NOT LIKE'],
```

dstorozhuk (commented):
Suggested fix for the delete-operation retention gap (see my comment above), plus the matching test expectation.

docroot/modules/custom/mass_admin_audit_trail/src/Hook/MassAdminAuditTrailHooks.php
tfleming-ma: Changed to checking against a list of PERMENANT_OPS.

docroot/modules/custom/mass_admin_audit_trail/tests/src/Unit/Hook/MassAdminAuditTrailHooksTest.php
tfleming-ma: Changed to checking against a list of PERMENANT_OPS.

dstorozhuk (approved):

PR #3493, opened 2026-09-03
DP-48482: Keep per-node Rabbit Hole overrides working after the 2.x upgrade

dstorozhuk:
**Description:**

Targets `feature/DP-48482_php-8-4-upgrade`, not `develop`.

On that branch, deploying the Rabbit Hole 1.x → 2.x upgrade silently drops every per-node override, so Person pages that are published through one start returning 404 — 367 published `person` nodes in my sanitized copy, since the `person` bundle default is `page_not_found`.

`rabbit_hole_update_8104` only creates the new `rabbit_hole__settings` field for bundles whose stored settings still carry `allow_override`, and the branch already ships `rabbit_hole.behavior_settings.*` in the 2.x format where that key is gone. So the field is never created, `rabbit_hole_update_8106` reports "There are no entities to be updated" and marks itself complete, and Rabbit Hole 2.x — which reads only that field, never the legacy `rh_action` — falls every node back to its bundle default. Nothing errors and `drush config:status` stays clean.

This PR adds the two pieces that were missing:

1. The field configuration: `field.storage.{node,taxonomy_term}.rabbit_hole__settings`, the `field.field.*.rabbit_hole__settings` instances for the bundles that allowed an override, and the form/view displays that carry the widget.
2. `mass_content_deploy_rabbit_hole_settings()`, which copies `rh_action` — plus the redirect settings, for `page_redirect` — onto the new field. Exporting the config alone is not enough: the field comes back empty, because `rabbit_hole_update_8106` is already marked as run and would find nothing even if replayed. The hook runs after config import, batches 25 entities at a time, and saves with `MASS_FLAGGING_BYPASS`, `setNewRevision(FALSE)` and `setSyncing(TRUE)` so it does not notify watchers, grow the revision table, or reshuffle All Content by `changed`.

It also restores `rabbit_hole.behavior_settings.*` to the versions already on the target branch — a local export had regressed the `taxonomy_term` settings back to the 1.x `taxonomy_vocabulary_*` ids with an empty `entity_type_id`, which Rabbit Hole 2.x cannot resolve.

DP-48482

**To Test:**

Needs a database still on Rabbit Hole 1.x — `\Drupal::keyValue('system.schema')->get('rabbit_hole')` returns `8101`.

- [ ] On `develop`, run `drush deploy` and note a published `person` alias whose node has `rh_action = display_page` — it returns 200, while any other published `/person/...` returns 404
- [ ] Check out `feature/DP-48482_php-8-4-upgrade`, `composer install`, `drush deploy` — the same alias now returns 404
- [ ] Check out this branch, `composer install`, `drush deploy` — the alias returns 200 again, the others stay 404
- [ ] `drush config:status` reports no differences
- [ ] `SELECT COUNT(*) FROM node__rabbit_hole__settings` matches the legacy count from `SELECT COUNT(*) FROM node_field_data WHERE rh_action NOT IN ('', 'bundle_default')`

---


joeg8612:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against f953804258d91f8eea6cd92a3ecd67fa57e982a7
</p>

PR #3498, opened 2026-09-10
DP-48764-Update-insecure-drupal-modules-that-are-not-in-the-PHP-upgrade-ticket

jarielrrz:


joeg8612:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  Changelog is not valid the following item(s) needs to be fixed (data.Changed[0] should have required property '.issue')
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 160042f8bc1a073703136692eba8259e1533e095
</p>

dstorozhuk:
Reviewed with a focus on blockers only. Two of them, plus one regression that comes from a dropped patch.

### 1. Diff 2.0.1 fatals on plugin discovery (address + dynamic_entity_reference)

Diff 2.x declares `FieldDiffBuilderInterface::build(): mixed` and `CoreFieldBuilder::build(): array`. Two contrib plugins we have enabled still declare `build()` with no return type:

* `docroot/modules/contrib/address/src/Plugin/diff/Field/AddressFieldBuilder.php:24`
* `docroot/modules/contrib/dynamic_entity_reference/src/Plugin/diff/Field/DynamicEntityReferenceFieldBuilder.php:25`

"Declaration must be compatible" is a hard fatal. So any call that builds the diff builder definitions dies, whether or not the entity has an address field.

Where it breaks:

* `/node/{nid}/revisions/view/{left}/{right}/split_fields` (this is exactly the URL embedded in every watch notification email, see `mass_flagging.module:88`)
* `/admin/config/content/diff/fields`

How to catch it by hand: open any node, go to the Revisions tab, select two revisions and press Compare. Expect a WSOD / 500, and a PHP fatal in the log with `AddressFieldBuilder::build()`. `/admin/config/content/diff/fields` should fail the same way.

The `summary()` workaround in this PR only avoids the plugin manager inside mass_flagging, so the page the notification links to is still dead. There is no fixed release to move to (address 2.0.4 is the latest, DER only fixes it in 5.0.0-rc1), so both modules need a local composer patch adding the return types.

### 2. The rewritten `summary()` silently suppresses watch notifications

`mass_flagging/src/Service/MassFlaggingEntityComparison.php:177` says it restores the Diff 8.x-1.10 field map, but it is not equivalent. 1.10 only kept a field when `createInstanceForFieldDefinition()` returned a plugin (NULL means the field type has no builder, or it is set to hidden). The new version only checks `showDiff()`, so that gate is gone.

Running both implementations side by side on a local copy of prod content: the field map for a service_page grows from 105 to 523 entries, and `Revision create time`, `Changed`, `Metatags (Hidden field for JSON support)` and `Revision translation affected` now show up as "changed".

The bad part is not the noise in the message, it is `checkRevisionforImageSectionChanges()` at `:113`. It only checks `count($elements) > 0` and skips every non paragraph key, and `changed` / `revision_timestamp` differ on every single revision. So it now returns TRUE where it used to return FALSE, and `mass_flagging.module:167` does `return;` on TRUE, which means watchers get no email at all. 

How to catch it by hand: watch a node as a second user, then edit only node level fields (title, organization, parent page) and save a new revision. Before this PR the watcher gets an email, after it they get nothing. Also compare the "Changes on: ..." line in an email that does go out.

Patching the two contrib modules from point 1 makes the plugin gate work again, and then `summary()` can be a literal copy of the 1.10 implementation instead of a reimplementation.

### 3. Dropping `diff-3228798-17.patch` is an accessibility regression

The infinite loop patch (3033455) is genuinely fixed upstream, 2.0.1 has the `array_key_exists()` guard, so removing that one is fine.

The accessibility patch is not upstream. In diff 2.0.1:

* `RevisionOverviewForm::buildSelectColumn()` is still `'#title_display' => 'invisible'` with no `#title`, so the revision select radios have no accessible name
* the per row links are back to plain `Revert` and `Delete`, so a screen reader user gets dozens of identically named links on one page

Both were fixed by the patch we are removing. Either re-roll it against 2.0.1 or open a follow up ticket before this ships, this is an editor facing page on a state site.

### Tests

`docroot/modules/custom/mass_flagging` has no `tests/` directory at all, which is why CI is green on both blockers. Given that this PR rewrites the notification logic, it would be good to add at least ExistingSite coverage for "node level only edit still notifies a watcher" and "image only paragraph edit does not". If that is too much for this ticket, a separate ticket for mass_flagging test coverage works, but it should exist before we change this service again.

composer.json
```diff
@@ -183,9 +183,9 @@
         "drupal/conditional_fields": "^4.0@alpha",
         "drupal/config_ignore": "^3.3",
         "drupal/content_language_no_outbound": "1.0.0-rc1",
-        "drupal/core-composer-scaffold": "~11.3.12",
-        "drupal/core-project-message": "~11.3.12",
-        "drupal/core-recommended": "~11.3.12",
+        "drupal/core-composer-scaffold": "11.3.16",
```
dstorozhuk: @jarielrrz  please change back to `~`.

dstorozhuk (changes requested):

composer.json
dstorozhuk: Also see the ticket for additional comments.

dstorozhuk (approved):

PR #3502, opened 2026-09-17
DP-48604-500-error-fix-for-event-listing-page

jarielrrz:


joeg8612:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against a9e953deea445d4bfdb220e41c58c6724c731fe2
</p>

docroot/modules/custom/mass_more_lists/src/Controller/EventsController.php
```diff
@@ -66,108 +91,188 @@ public function upcomingPage(NodeInterface $node) {
...
-    throw new CacheableNotFoundHttpException($metadata);
+
+    $query = $request->query->all();
+    $mayflower_page = (int) $query['_page'];
+    unset($query['_page']);
+    if ($mayflower_page > 1 && !isset($query['page'])) {
+      $query['page'] = $mayflower_page - 1;
+    }
+
+    return $this->redirect($route_name, ['node' => $node->id()], ['query' => $query]);
```
dstorozhuk, line 275: `$this->redirect()` returns a plain `RedirectResponse`, so this is a 302 with `Cache-Control: must-revalidate, no-cache, private` and `X-Drupal-Dynamic-Cache: UNCACHEABLE (no cacheability)` (verified locally). Every hit on a legacy `?_page=...` URL bypasses Varnish/Akamai and runs a full Drupal bootstrap, and `?_page=1` is exactly the URL currently indexed by Google, it is in the ticket description.

Since the `_page` to `page` mapping is permanent, this should be a 301, and it should be edge-cacheable: `LocalRedirectResponse` implements `CacheableResponseInterface`, so `addCacheableDependency()` with a `url.query_args:_page` context would let Akamai serve the redirect. A 302 tells crawlers the old URL is still live, so they keep coming back to origin for it.

docroot/modules/custom/mass_more_lists/src/Controller/EventsController.php
```diff
@@ -66,108 +91,188 @@ public function upcomingPage(NodeInterface $node) {
...
+   *   Either 'past' or 'upcoming'.
+   * @param int $total
+   *   Total matching events.
+   *
+   * @return array
+   *   Events, results heading, and pager render array.
+   */
+  private function getPagedEvents(NodeInterface $node, string $type, int $total): array {
+    $limit = $type === 'past' ? self::PAST_EVENTS_PER_PAGE : self::EVENTS_PER_PAGE;
+    $page = $this->pagerManager->createPager($total, $limit)->getCurrentPage();
```
dstorozhuk, line 216: `Pager::setCurrentPage()` clamps out-of-range values, so any `page` value returns 200 and renders the last page: `?page=9999` gave me "Showing 3201-3276 of 3276 results". These listing pages also emit no `<link rel="canonical">` at all, so every integer is a separate indexable URL rendering 100 event teasers.

Joseph raised exactly this cost in the ticket ("a crawler will crawl all those next links and just add to our pageviews"). Worth either returning a 404 when the requested page is past `$pager->getTotalPages()`, or adding a canonical to the listing pages.

dstorozhuk (commented):
Approach looks right and the fix is real. I ran both versions locally against a prod DB copy, node 6681 (Division of Occupational Licensure, 3,276 past events):

| | develop | this PR |
|---|---|---|
| `/node/6681/events/past` | 8.03s, 6.6 MB HTML | 2.42s cold / 0.05s warm, 314 KB |

Also verified by hand: `?page=1` renders "Showing 101-200 of 3276", `?_page=1` redirects to the bare URL, `?_page=5` redirects to `?page=4`, upcoming with 23 events gives 3 pages of 10, past with 40 events gives one page and no pager. phpcs and twig-cs-fixer are clean. The Mayflower pagination JS does not hijack the server-side pager, it binds to a `.js-pagination` container that `mass_theme`'s `pager.html.twig` never emits.

Three notes inline.

docroot/modules/custom/mass_more_lists/src/Controller/EventsController.php
```diff
@@ -66,108 +91,188 @@ public function upcomingPage(NodeInterface $node) {
...
+      : $this->eventManager->getUpcoming($node, $limit, $offset);
+
+    $start = $offset + 1;
+    $end = min($offset + $limit, $total);
+
+    $pager = [];
+    if ($total > $limit) {
+      $pager = [
+        '#type' => 'pager',
+        '#tags' => [
```
dstorozhuk, line 229: `Previous` and `Next` are untranslated literals. Copied verbatim from `MassMoreLists::buildList()`, so not a regression introduced here, but we just added Portuguese in DP-48740, so `$this->t()` would be worth it.

dstorozhuk (approved):

